PR #54, opened 2024-11-08
London | Elhadj Abdoul Diallo | Module-Data-Groups | WEEK3 - alarmclock

eediallo:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

I created a branch (feature/alarmclock) in which I solved the alarm clock project. To complete the project I implemented the following features: 

- Given the alarm is set with a valid time When one second passes Then the “Time Remaining” title should decrement by 1 second

- Given the alarm is set with a time of 00:00 When the timer reaches 00:00 Then the alarm sound should play - continuously

- Given the alarm sound is currently playing When the user clicks the “Stop Alarm” button Then the alarm sound should stop playing

- Given the alarm is set with a time of 00:10 When the timer reaches 00:00 Then the background color should change And the alarm sound should play

- Given the user has not set an alarm When the page first loads Then the “Time Remaining” title should show 00:00 And no alarm sound should play

### DIFF
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -46,6 +46,8 @@ function playAudioWhenTimerReachesZero(totalTimeInSeconds, setIntervalId) {
     playAlarm();
   }
 }
+
+// event to pause alarm
 stopBtn.addEventListener("click", pauseAlarm);
 
 //===============changeBackgroundColor==============================
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -1,6 +1,9 @@
 function setAlarm() {
   const heading = document.querySelector("#timeRemaining");
   const time = document.querySelector("#alarmSet").value;
+  const totalTimeInSeconds = time;
+  const totalMinutes = Math.floor(totalTimeInSeconds / 60);
+  const remainingSeconds = totalTimeInSeconds % 60;
 
   heading.innerText = "Time Remaining: 00:19";
 }
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -38,7 +38,7 @@ function countDownEverySeconds(totalTimeInSeconds, heading) {
 
 //=================playAudioWhenTimerReachesZero=========================
 function playAudioWhenTimerReachesZero(totalTimeInSeconds, setIntervalId) {
-  if (totalTimeInSeconds < 0) {
+  if (totalTimeInSeconds <= 0) {
     clearInterval(setIntervalId);
     playAlarm();
   }
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -25,11 +25,15 @@ function countDownEverySeconds(totalTimeInSeconds, heading) {
       totalMinutes
     )}:${formatTimeUnit(remainingSeconds)}`;
 
-    if (totalTimeInSeconds <= 0) {
-      clearInterval(setIntervalId);
-      playAlarm();
-    }
+    playAudioWhenTimerReachesZero(totalTimeInSeconds, setIntervalId);
   }, 1000);
+}
+
+function playAudioWhenTimerReachesZero(totalTimeInSeconds, setIntervalId) {
+  if (totalTimeInSeconds < 0) {
+    clearInterval(setIntervalId);
+    playAlarm();
+  }
 }
 
 // DO NOT EDIT BELOW HERE
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -36,7 +36,6 @@ function countDownEverySeconds(totalTimeInSeconds, heading) {
   }, 1000);
 }
 
-
 //=================playAudioWhenTimerReachesZero=========================
 function playAudioWhenTimerReachesZero(totalTimeInSeconds, setIntervalId) {
   if (totalTimeInSeconds < 0) {
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -48,6 +48,7 @@ function playAudioWhenTimerReachesZero(totalTimeInSeconds, setIntervalId) {
 }
 stopBtn.addEventListener("click", pauseAlarm);
 
+//===============changeBackgroundColor==============================
 function changeBackgroundColor() {
   const body = document.querySelector("body");
   body.style.backgroundColor = "#444";
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -45,6 +45,7 @@ function playAudioWhenTimerReachesZero(totalTimeInSeconds, setIntervalId) {
   }
 }
 
+stopBtn.addEventListener("click", pauseAlarm);
 // DO NOT EDIT BELOW HERE
 
 var audio = new Audio("alarmsound.mp3");
@@ -67,5 +68,4 @@ function pauseAlarm() {
   audio.pause();
 }
 
-stopBtn.addEventListener("click", pauseAlarm);
 window.onload = setup;
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -4,6 +4,7 @@ function formatTimeUnit(unit) {
 
 const heading = document.querySelector("#timeRemaining");
 const stopBtn = document.querySelector("#stop");
+heading.innerText = "Time Remaining: 00:00";
 //===========setAlarm======================
 function setAlarm() {
   const totalTimeInSeconds = parseInt(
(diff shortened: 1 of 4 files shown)


Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,5 +1,61 @@
...
+  setInitialMinutesAndSeconds(heading, totalTimeInSeconds);
+  countDownEverySeconds(totalTimeInSeconds, heading);
+}
+
+//=============setInitialMinutesAndSeconds=======================
+function setInitialMinutesAndSeconds(heading, totalTimeInSeconds) {
+  const initialMinutes = Math.floor(totalTimeInSeconds / 60);
+  const InitialSeconds = totalTimeInSeconds % 60;
+
+  heading.innerText = `Time Remaining: ${formatTimeUnit(
```
illicitonion, line 22: There are three different places in your code where you basically write ``heading.innerText = `Time Remaining: [some expression to convert to a timestamp]`;``, and there are two places you have the code which takes a number of seconds and turn it into a string that looks like `MM:SS`.

Imagine if you wanted to change part of this. Let's imagine you wanted to change the text "Time Remaining" to "Remaining Time" - you'd need to change three different parts of code. It'd also be easy for you to only change two of them, and miss the third one, and then your UI would change from "Remaining Time" to "Time Remaining" at some point.

Also, imagine if there was a bug in your code which converts seconds to `MM:SS`, if you wanted to show `HH:MM:SS` if the timer was for more than 60 minutes - you would currently need to fix 2-3 different places.

Can you think of a way to only write the code that takes a number of seconds and sets the `heading.innerText` to `MM:SS` one time?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,5 +1,61 @@
...
+}
+
+//================countDownEverySeconds======================
+function countDownEverySeconds(totalTimeInSeconds, heading) {
+  const setIntervalId = setInterval(() => {
+    totalTimeInSeconds--;
+    const totalMinutes = Math.floor(totalTimeInSeconds / 60);
+    const remainingSeconds = totalTimeInSeconds % 60;
+    heading.innerText = `Time Remaining: ${formatTimeUnit(
+      totalMinutes
```
illicitonion, line 34: I find it a little hard to read this code which is split onto lines like this - I think it would be easier to read either:

All on one line:
```js
heading.innerText = `Time Remaining: ${formatTimeUnit(totalMinutes)}:${formatTimeUnit(remainingSeconds)}`;
```

or pulling out intermediate variables:
```js
const formattedMinutes = formatTimeUnit(totalMinutes);
const formattedSeconds = formatTimeUnit(remainingSeconds);
heading.innerText = `${formattedMinutes}:${formattedSeconds}`;
```

What do you think? Which of the three do you prefer? Which is the most clear?

illicitonion (commented):
This generally looks really good, nicely put together!

Sprint-3/alarmclock/alarmclock.js
eediallo: This is an excellent observation. I can use a function to handle this.

Sprint-3/alarmclock/alarmclock.js
eediallo: I think I prefer the second approach because with prettier the 1st one would be formated as it it at the moment

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,5 +1,69 @@
...
+
+//=========formatAndDisplayTime=========================
+function formatAndDisplayTime(text, minutes, seconds ){
+  const formattedMinutes = formatTimeUnit(minutes)
+  const formattedSeconds = formatTimeUnit(seconds)
+  heading.innerText = `${text} ${formattedMinutes}:${formattedSeconds}`
+}
+
+//===============calculateMinutesAndSeconds================
+function calculateMinutesAndSeconds(totalTimeInSeconds){
```
illicitonion, line 27: This works (and reduces duplication of logic), but you're still ending up with both of your callers doing exactly the same thing - calling this function and then passing it to `formatAndDisplayTime`.

Is there a reason `formatAndDisplayTime` couldn't accept `totalTimeInSecond`, and do the calculation of minutes+seconds itself?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,5 +1,69 @@
...
+  const totalMinutes = Math.floor(totalTimeInSeconds / 60);
+  const remainingSeconds = totalTimeInSeconds % 60;
+  return [totalMinutes, remainingSeconds]
+}
 
+//=============setInitialMinutesAndSeconds=======================
+function setInitialMinutesAndSeconds(totalTimeInSeconds) {
+  const [initialMinutes, InitialSeconds] = calculateMinutesAndSeconds(totalTimeInSeconds)
```
illicitonion, line 35: Nice use of destructuring!

Sprint-3/alarmclock/alarmclock.js
eediallo: formatAndDisplayTime can do it. trying to fix it now

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,5 +1,63 @@
...
+function formatAndDisplayTime(totalTimeInSeconds){
+  const totalMinutes = Math.floor(totalTimeInSeconds / 60);
+  const remainingSeconds = totalTimeInSeconds % 60;
+  const formattedMinutes = formatTimeUnit(totalMinutes)
+  const formattedSeconds = formatTimeUnit(remainingSeconds)
+  heading.innerText = `${headerText} ${formattedMinutes}:${formattedSeconds}`
+}
+
+//=============setInitialMinutesAndSeconds=======================
+function setInitialMinutesAndSeconds(totalTimeInSeconds) {
```
illicitonion, line 30: I'm not sure how much value this function is bringing - what do you think is better about having this function, rather than just calling `formatAndDisplayTime(totalTimeInSeconds)` where it's currently called?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,5 +1,63 @@
-function setAlarm() {}
+const heading = document.querySelector("#timeRemaining");
+const stopBtn = document.querySelector("#stop");
+heading.innerText = "Time Remaining: 00:00";
+const headerText = 'Remaining Time:'
```
illicitonion, line 4: This code now exhibits that bug I suggested could be possible - the header text changes from "Time Remaining" to "Remaining Time" when you start the timer. This feels like a bug?

How could you structure this code to make that bug less likely to happen?

illicitonion (commented):
Looking really good!

Sprint-3/alarmclock/alarmclock.js
eediallo: I don't think it has much value in this context. We can just get rid of it and call ```formatAndDisplayTime(totalTimeInSeconds)``` where it is currently called

Sprint-3/alarmclock/alarmclock.js
eediallo: I think I can play with the ```heading.innerText = "Time Remaining: 00:00"```. to extract the ```Time Remaining:``` and store it into a variable. I am not 100% sure but that's the idea I have at the moment

illicitonion: You have already written a function that sets the heading properly given a number of seconds - could you think how you could re-use that function, rather than writing `heading.innerText = ...` again here?

eediallo: Thinking about it now

eediallo: I can call ```formatAndDisplayTime``` with ```totalTimeInSeconds``` set to 0 as the default value.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,5 +1,59 @@
-function setAlarm() {}
+const heading = document.querySelector("#timeRemaining");
+const stopBtn = document.querySelector("#stop");
 
+const headerText = 'Remaining Time:'
+formatAndDisplayTime(totalTimeInSeconds = 0) // displays initial heading with timer values (00:00)
```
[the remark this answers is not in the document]
illicitonion: Aha! This is something interesting to dig into!

In this code:
```js
function formatAndDisplayTime(totalTimeInSeconds){
```

The function `formatAndDisplayTime` is saying "You should pass me one value as an argument. I will give it a name, `totalTimeInSeconds`, for as long as I'm running, and that name will disappear when I finish running". It's a name that only matters to the function itself inside of it.

So if you call `formatAndDisplayTime(0)`, when we jump inside `formatAndDisplayTime` it will assign `0` to `totalTimeInSeconds`, and when `formatAndDisplayTime` finishes, there won't be something called `totalTimeInSeconds` any more.

Everywhere that you have a variable called `totalTimeInSecond`, you could use a completely different name if you wanted to - the name of what you have when you _call_ a function is unrelated to the name you use when you're inside the function.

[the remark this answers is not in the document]
eediallo: Thank you for clarifying this. It is really helpful

illicitonion (commented):
Amazing, this looks great! 🎉 

Do go and take the things you've learnt here, and apply them to your other projects, and remember them when you're writing code in the future!

eediallo:
> Amazing, this looks great! 🎉
> 
> Do go and take the things you've learnt here, and apply them to your other projects, and remember them when you're writing code in the future!

I will. Thank you so much.

PR #58, opened 2024-11-14
NW7 | Sue Mohamadi | Module-Data-Groups | Sprint - 1

SueMsOl:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -3,9 +3,15 @@
 // If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
 
 function calculateMedian(list) {
+  list.sort((a,b)=> a-b);
   const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  if (list.length%2 == 1) {
+  const medianOdd = list[middleIndex];
+  return medianOdd;
 }
-
+else {
+  const medianEven = (list[middleIndex] + list[middleIndex -1 ])/2;
+  return medianEven;
+}
+}
 module.exports = calculateMedian;
--- a/Sprint-1/implement/max.js
+++ b/Sprint-1/implement/max.js
@@ -1,4 +1,7 @@
+//const ITEMS = ["a","b","c"];
 function findMax(elements) {
+    let numericItems = elements.filter(value => typeof value === 'number');
+    return numericItems.length > 0 ? Math.max(...numericItems) : -Infinity;
 }
-
+//console.log(findMax(ITEMS));
 module.exports = findMax;
--- a/Sprint-1/implement/max.js
+++ b/Sprint-1/implement/max.js
@@ -1,7 +1,7 @@
-const ITEMS = [10,50,60,"a"];
+//const ITEMS = [10,50,60,"a"];
 function findMax(elements) {
     let numericItems = elements.filter(value => typeof value === 'number');
     return numericItems.length > 0 ? Math.max(...numericItems) : -Infinity;
 }
-console.log(findMax(ITEMS));
+//console.log(findMax(ITEMS));
 module.exports = findMax;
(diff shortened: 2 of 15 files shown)


eliza-an (changes requested):
Its not finished yet but i haave to run so saving the changes. I will try to submit the rest when I can

Sprint-1/fix/median.js
```diff
@@ -3,9 +3,15 @@
 // If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
 
 function calculateMedian(list) {
+  list.sort((a,b)=> a-b);
```
cjyuan, line 6: Array is passed to a function ***by reference***  (instead of ***by value***). This implies when an array is passed to a function via parameter, the array is **shared** by the caller and the ***callee*** (the function being called).

If you are unfamiliar with this concept, you can try asking ChatGPT "Explain how array is passed by reference in JS".

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,7 @@
+//const ITEMS = [10,50,60,"a"];
 function findMax(elements) {
+    let numericItems = elements.filter(value => typeof value === 'number');
+    return numericItems.length > 0 ? Math.max(...numericItems) : -Infinity;
 }
-
+//console.log(findMax(ITEMS));
```
cjyuan, line 6: What do you expect from the following function calls (on extreme cases)?

```
findMax([-Infinity])
findMax([-Infinity, 0])
findMax([NaN])
findMax([0, NaN, 1])
```

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,8 @@
 function sum(elements) {
+    let numericalValues = elements.filter(item => typeof item === "number");
+    let sumOfValues = numericalValues.reduce((a,b) => a+b, 0);
+    return numericalValues.length > 0 ? sumOfValues : 0;
 }
-
+//const ITEMS = ["a","b"];
```
cjyuan, line 6: What do you expect from the following function calls (on extreme cases)?

```
sum([Infinity, 1]);
sum([Infinity, Infinity]);
sum([NaN, 1]);
sum([Infinity, -Infinity]);
```

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,37 @@ const sum = require("./sum.js");
 // Given an empty array
 // When passed to the sum function
 // Then it should return 0
-test.todo("given an empty array, returns 0")
-
+//test.todo("given an empty array, returns 0")
+test (" ", ()=>{
```
cjyuan, line 17: Could include a concise description in the form "Expect 0 when given an empty array" or even "Empty array" would be good.

This way, when a test failed, we can have more info about why the test failed.

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,37 @@ const sum = require("./sum.js");
...
-
+test (" ", ()=>{
+    expect(sum([-10,-80])).toEqual(-90); 
+ })
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
-
+test (" ", ()=>{
+    expect(sum([-10.8,-80.2])).toEqual(-91);
```
cjyuan, line 36: Decimal numbers in most programming languages (including JS) are internally represented in "floating point number" format. Floating point arithmetic is not exact. For example, the result of `46.5678 - 46 === 0.5678` is false because `46.5678 - 46` only yield a value that is very close to `0.5678`.

Sprint-1/stretch/aoc-2018-day1/solution.js
```diff
@@ -0,0 +1,20 @@
+const fs = require("fs");
+const input = fs.readFileSync("./input.txt", "utf8");
+function calculateFrequency(arg1){
+    let inputToArray = arg1.split('\n').map(Number);
```
cjyuan, line 4: Can you explain how `.map(Number)` works? This is something relatively advanced but not difficult to understand.

cjyuan (commented):
Code is good. I just had some comments and suggestions.

PR #60, opened 2024-11-15
Glasgow | Prati Amalden | Module-Data-Groups | Sprint 1

PratiAmalden:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-1/implement/sum.test.js
+++ b/Sprint-1/implement/sum.test.js
@@ -34,7 +34,7 @@ describe("sum", () => {
   // When passed to the sum function
   // Then it should return the correct total sum
   test("given an array with decimal/float numbers, return the correct total sum", () => {
-    expect(sum([1.0, 1.2, 1.5])).toBe(3.7);
+    expect(sum([1.0, 1.2, 1.5])).toBeCloseTo(3.7);
   });
   // Given an array containing non-number values
   // When passed to the sum function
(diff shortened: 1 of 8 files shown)


Sprint-1/implement/sum.test.js
```diff
@@ -10,27 +10,42 @@ const sum = require("./sum.js");
...
+  // When passed to the sum function
+  // Then it should still return the correct total sum
+  test("given an array containing negative numbers, return the correct total sum", () => {
+    expect(sum([-1, -7, -10])).toBe(-18);
+  });
+  // Given an array with decimal/float numbers
+  // When passed to the sum function
+  // Then it should return the correct total sum
+  test("given an array with decimal/float numbers, return the correct total sum", () => {
+    expect(sum([1.0, 1.2, 1.5])).toBe(3.7);
```
cjyuan, line 37: Decimal numbers in most programming languages (including JS) are internally represented in "floating point number" format. Floating point arithmetic is not exact. For example, the result of `46.5678 - 46 === 0.5678` is false because `46.5678 - 46` only yield a value that is very close to `0.5678`.

So to test a calculated result that involves decimal numbers, we should use `.toBeCloseTo()` instead of `.toBe()`.
See: https://jestjs.io/docs/expect#tobeclosetonumber-numdigits

cjyuan (commented):
Code is good. Just a few comments and suggestions.

PR #61, opened 2024-11-16
North West | Zabihollah Namazi | Module-Data-Groups | Sprint 1

ZabihollahNamazi:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.
I answered all questions and wrote all codes which were asked.

## Questions


Sprint-1/implement/max.test.js
```diff
@@ -16,28 +16,43 @@ const findMax = require("./max.js");
 // When passed to the max function
 // Then it should return -Infinity
 // Delete this test.todo and replace it with a test.
-test.todo("given an empty array, returns -Infinity");
+test("given an empty array, returns -Infinity", () => {
+    expect(findMax([])).toEqual(Infinity);
```
cjyuan, line 20: The expected value mentioned in the description of this test is `-Infinity`.

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,22 @@
...
+        return Infinity
+    }
+    for(let i =0; i < elements.length ; i++){
+        
+        if(typeof elements[i] == "number" && elements[i] > max){
+            max = elements[i];
+        }
+    }
+    if(max == -(Infinity)){
+        return "there is no number in that array!!";
```
cjyuan, line 13: There is also no numbers in an empty array, `[]`. Returning this value would somehow make the return value "not consistent".

Also, what do you expect from the following function calls (on extreme cases)?

```
findMax([-Infinity]);
findMax([NaN]);
```

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,12 @@
 function sum(elements) {
+    let total = 0;
+    for(let i = 0 ; i < elements.length ; i++){
+        if(typeof elements[i] == "number"){
+            total += elements[i];
+        }
+    }
+    return total;
```
cjyuan, line 8: What do you expect from the following function calls (on extreme cases)?

```
sum([Infinity, 1]);
sum([Infinity, Infinity]);
sum([NaN, 1]);
sum([Infinity, -Infinity]);
```

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,37 @@ const sum = require("./sum.js");
...
-
+test("Given an array containing negative numbers, Then it should return the correct total sum", () =>{
+    expect(sum([-2, -3])).toEqual(-5);
+});
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
-
+test("Given an array with decimal/float numbers, Then it should return the correct total sum", () =>{
+    expect(sum([1.2, 3.4])).toEqual(4.6);
```
cjyuan, line 36: Decimal numbers in most programming languages (including JS) are internally represented in "floating point number" format. Floating point arithmetic is not exact. For example, the result of `46.5678 - 46 === 0.5678` is false because `46.5678 - 46` only yield a value that is very close to `0.5678`.

Sprint-1/refactor/includes.js
```diff
@@ -1,13 +1,15 @@
...
-    }
-  }
-  return false;
+    // if (list.includes(target)) {
+    //   return true;
+    // }
+    // else{
+    //   return false;
+    // }
+    return list.includes(target)
```
cjyuan, line 10: While this solution works, the spec does have the following requirement:
`// Refactor the implementation of includes to use a for...of loop`

Sprint-1/stretch/aoc-2018-day1/solution.js
```diff
@@ -0,0 +1,22 @@
+
+const fs = require('fs');
+
+fs.readFile('/root/CYF/Module-Data-Groups/Sprint-1/stretch/aoc-2018-day1/input.txt', (err, data) => {
```
cjyuan, line 4: With the file path specified as an absolute path, the script probably cannot find `input.txt` when it is executed on a different computer.

Assuming `input.txt` will always be in the same folder as the script, how would you modify your code to read data from this `input.txt` file?

cjyuan (commented):
I left my comments and suggestions in the code.

Sprint-1/stretch/aoc-2018-day1/solution.js
```diff
@@ -0,0 +1,22 @@
...
+  if (err) throw err;
+  console.log(device(data.toString().split("\n")));
+});
+
+function device(array){
+    let result = 0;
+    for(let i =0; i < array.length; i++){
+        if(array[i].slice(0, 1) == "+"){
```
cjyuan, line 12: To test if the first character is a "+", you could also write `array[i][0] == "+"` or `array[0].charAt(0) == "+"`.

You can possibly simplify your code and make your code perform better.
Hint: What are the value of `Number("+10")` and `Number("-10")`?

cjyuan (commented):
Just 1 more comment.

PR #67, opened 2024-11-22
West Midlands|Azar Alampanah|Module-Data-Groups|week1

azaralampanah2:
## Learners, PR Template

Self checklist

- [yes ] I have committed my files one by one, on purpose, and for a reason
- [yes ] I have titled my PR with COHORT_NAME | FIRST_NAME LAST_NAME | REPO_NAME | WEEK 
- [yes ] My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/contributing/)
- [yes ] My changes meet the [requirements](./README.md) of this task

## Changelist

Briefly explain your PR.
This PR has been raised for my work on Sprint-1 exercises.

## Questions

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -3,9 +3,24 @@
 // If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  if (JSON.stringify(list) === JSON.stringify([1, 2, 3])) {
+    return [1, 2, 3];
+  }
+
+  if (list.length % 2 == 0) {
+    let sum = 0;
+    for (let i = 0; i < list.length; i++) {
+      sum += list[i];
+    }
+
+    avarage = sum / list.length;
+    return avarage;
+  } else {
+    const middleIndex = Math.floor(list.length / 2);
+    const median = list.splice(middleIndex, 1)[0];
+    return median;
+  }
 }
+//console.log(calculateMedian([1, 2, 3]));
 
 module.exports = calculateMedian;
--- a/Sprint-1/implement/max.js
+++ b/Sprint-1/implement/max.js
@@ -1,4 +1,19 @@
 function findMax(elements) {
-}
+  let maxNumber = elements[0];
+  if (elements.length === 0) {
+    return "-Infinity";
+  }
+  
 
+  for (let i = 1; i < elements.length; i++) {
+    if (maxNumber < elements[i]) {
+      maxNumber = elements[i];
+    }
+  }
+  return maxNumber;
+
+
+ 
+}
+console.log(findMax([]));
 module.exports = findMax;
--- a/Sprint-1/implement/max.js
+++ b/Sprint-1/implement/max.js
@@ -1,19 +1,23 @@
 function findMax(elements) {
-  let maxNumber = elements[0];
-  if (elements.length === 0) {
+  const numericValues = elements.filter(
+    (element) => typeof element === "number" && !isNaN(element)
+  );
+  let maxNumber = numericValues[0];
+  //console.log(numericValues);
+  if (
+    elements.length === 0 ||
+    elements.every((element) => typeof element != "number")
+  ) {
     return "-Infinity";
   }
-  
 
-  for (let i = 1; i < elements.length; i++) {
-    if (maxNumber < elements[i]) {
-      maxNumber = elements[i];
+  for (let i = 1; i < numericValues.length; i++) {
+    if (maxNumber < numericValues[i]) {
+      maxNumber = numericValues[i];
     }
   }
   return maxNumber;
-
-
- 
 }
-console.log(findMax([]));
+//console.log(findMax(["hi", "bye", "no", 10, 1]));
+
 module.exports = findMax;
--- a/Sprint-1/implement/max.test.js
+++ b/Sprint-1/implement/max.test.js
@@ -16,27 +16,49 @@ const findMax = require("./max.js");
 // When passed to the max function
 // Then it should return -Infinity
 // Delete this test.todo and replace it with a test.
-test.todo("given an empty array, returns -Infinity");
+test("an empty array should return -Infinity", ()=>{
+expect(findMax([])).toBe("-Infinity")
+});
+
 
 // Given an array with one number
 // When passed to the max function
 // Then it should return that number
+test("an array with one number should return the number",()=>{
+    expect(findMax([0])).toEqual(0);
+    expect(findMax([-4])).toEqual(-4);
+    expect(findMax([102])).toEqual(102);
+});
 
 // Given an array with both positive and negative numbers
 // When passed to the max function
 // Then it should return the largest number overall
+test("should return the largest number", ()=>{
+    expect([-4,-12,0,-100,-65]).toBe(0)
+    expect([-4,-12,0,-100,1,60,112]).toBe(112)
+});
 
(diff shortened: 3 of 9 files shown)


Sprint-1/implement/max.js
```diff
@@ -1,4 +1,23 @@
 function findMax(elements) {
+  const numericValues = elements.filter(
+    (element) => typeof element === "number" && !isNaN(element)
+  );
+  let maxNumber = numericValues[0];
+  //console.log(numericValues);
+  if (
+    elements.length === 0 ||
+    elements.every((element) => typeof element != "number")
```
cjyuan, line 9: If `elements` is `["A", "B", "C", "D"]`, what do you expect `numericValues` be after line 4?

The check at line 9 is very similar to the check at line 3. Instead of carrying out another check at line 9,
why not just take advantage of `numericValues`?

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,23 @@
...
+  ) {
+    return "-Infinity";
```
cjyuan, line 11: `Infinity` is a keyword in the JavaScript language. So you can return `-Infinity` (which is a value of type "number") instead of returning `"-Infinity"`.

Sprint-1/implement/max.test.js
```diff
@@ -16,28 +16,51 @@ const findMax = require("./max.js");
 // When passed to the max function
 // Then it should return -Infinity
 // Delete this test.todo and replace it with a test.
-test.todo("given an empty array, returns -Infinity");
+test("an empty array should return -Infinity", () => {
+  expect(findMax([])).toBe("-Infinity");
```
cjyuan, line 20: Should be
`expect(findMax([])).toBe(-Infinity);`

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,23 @@
 function sum(elements) {
+  let sum1 = 0;
+  if (elements.every((Element) => typeof Element != "number")) {
```
cjyuan, line 3: If a value is not of type "number", it can be `undefined`, `null`, `true`, an array, or an object, which does not have the method `.charCodeAt()`.

When `elements` contains both numbers and non-number values, how does your code deal with the non-number values?

When `elements` contains only non-number values, why not give them the same treatment?

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,41 @@ const sum = require("./sum.js");
...
 // Then it should still return the correct total sum
+test("for an array with negative numbers should return sum of them", () => {
+  expect(sum([-5, -9, -4, -10])).toBe(-28);
+});
 
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
+test("for an array with decimal numbers should return sum of them", () => {
+  expect(sum([5.2, 1.9, 8.3])).toEqual(15.4);
```

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,41 @@ const sum = require("./sum.js");
...
 // Then it should ignore the non-numerical values and return the sum of the numerical elements
+test("if the array has non-numerical values inside, it should ignore them and just add up the numbers", () => {
+  expect(sum(["hi", 5, 4, 3, "NO"])).toBe(12);
+});
 
 // Given an array with only non-number values
 // When passed to the sum function
 // Then it should return the least surprising value given how it behaves for all other inputs
+test("array with only non-number values will add up the ASCII of the values!", () => {
+  expect(sum(["hi", "bye", "hello"])).toBe(1061);
```
cjyuan, line 52: Can also include non-number values such as `undefined`, `null`, and `true` in the test.

cjyuan (commented):
I left few comments and suggestions.

Sprint-1/fix/median.js
```diff
@@ -3,9 +3,24 @@
 // If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  if (JSON.stringify(list) === JSON.stringify([1, 2, 3])) {
```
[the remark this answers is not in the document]
azaralampanah2: I used this line to check whether the array list has the same contents as the array [1, 2, 3] in exact order. I had a look at description and this was not required, so I removed them

Sprint-1/fix/median.js
```diff
@@ -3,9 +3,24 @@
...
+  }
+
+  if (list.length % 2 == 0) {
+    let sum = 0;
+    for (let i = 0; i < list.length; i++) {
+      sum += list[i];
+    }
+
+    avarage = sum / list.length;
+    return avarage;
```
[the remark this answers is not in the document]
azaralampanah2: yes, my function is doing wrong. it return the avarage of all numbers not the required, so I amended it. Now it return the expected value and all tests work properly.
thanks CJ

Sprint-1/fix/median.js
```diff
@@ -3,9 +3,24 @@
...
+  } else {
+    const middleIndex = Math.floor(list.length / 2);
+    const median = list.splice(middleIndex, 1)[0];
```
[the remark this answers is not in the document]
azaralampanah2: I tried to run the test with your suggestion and it works, so I changed line 20 to the suggestion.
Thanks CJ

Sprint-1/fix/median.js
```diff
@@ -3,9 +3,24 @@
 // If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  if (JSON.stringify(list) === JSON.stringify([1, 2, 3])) {
+    return [1, 2, 3];
+  }
+
```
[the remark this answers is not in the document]
azaralampanah2: sorry CJ, I thought the median is the middle index, and all my tests passed because all were sorted.
I searched and found what median is and amended my function and insert line 6 to sort the array and then find the middle index.

Sprint-1/implement/max.js
azaralampanah2: with the array of ["A","B","C","D"], numericValues will be:[]
At line 9 ,I removed the check statement and insert this: "if numericValues !=0"
Thanks

Sprint-1/implement/max.js
azaralampanah2: Amended. thanks

Sprint-1/implement/max.test.js
azaralampanah2: Amended. Thanks

Sprint-1/implement/sum.js
azaralampanah2: I amended my codes, as you mentioned charCodeAt() can not be used for undefined, null, true, an array, or an object, so the least surprising value will be "return 0".
and also inserted"const numericValues = elements.filter(
    (element) => typeof element === "number" && !isNaN(element)
  );" to remove non-numeric values from the array, so the function will return"0" for any nun-numeric values and for number values will return the asciicode

Sprint-1/implement/sum.test.js
azaralampanah2: i inserted some more tests to check if the input is undefined, null, and true to return "0"
Thanks

Sprint-1/implement/sum.test.js
azaralampanah2: Thanks for this precious hint!
I searched for that . I need to put "toBeCloseTo()" instead of "toEqual".
I put another test for decimal numbers in my test file.

PR #68, opened 2024-11-23
North West | Zabihollah Namazi | Module-Data-Groups | Sprint 2

ZabihollahNamazi:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-2/implement/lookup.js
+++ b/Sprint-2/implement/lookup.js
@@ -1,5 +1,11 @@
-function createLookup() {
+function createLookup(list) {
   // implementation here
+  let currency = {};
+  for(let i = 0; i < list.length; i++){
+    currency[list[i][0]] = list[i][1];
+  }
+  return currency;
 }
-
+console.log(createLookup([['US', 'USD'], ['CA', 'CAD']]));
 module.exports = createLookup;
+
--- a/Sprint-2/implement/querystring.js
+++ b/Sprint-2/implement/querystring.js
@@ -6,11 +6,12 @@ function parseQueryString(queryString) {
   const keyValuePairs = queryString.split("&");
 
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
-    queryParams[key] = value;
+    const [key, ...rest] = pair.split("=");
+    queryParams[key] = rest.join("=");
   }
 
   return queryParams;
 }
 
+console.log(parseQueryString("equation=x=y+1"));
 module.exports = parseQueryString;
--- a/Sprint-2/implement/tally.js
+++ b/Sprint-2/implement/tally.js
@@ -1,3 +1,17 @@
-function tally() {}
-
+function tally(list) {
+    let count = {};
+    if(list.length == 0){
+        return {};
+    }
+    if(!Array.isArray(list)){
+        throw new Error("invalid input, input must be an array");
+    }
+    else{
+        for (let item of list) {
+            count[item] = (count[item] || 0) + 1;
+        }
+    }return count;
+}
+console.log(tally(["a", "b", "c", "c", "b"]));
 module.exports = tally;
+
--- a/Sprint-2/implement/tally.test.js
+++ b/Sprint-2/implement/tally.test.js
@@ -19,16 +19,26 @@ const tally = require("./tally.js");
 // Given a function called tally
 // When passed an array of items
 // Then it should return an object containing the count for each unique item
-
+test("tally When passed an array of items returns an object containing the count for each unique item", () =>{
+    expect(tally(["a", "b", "c"])).toEqual({a : 1, b : 1, c : 1});
+}); 
 // Given an empty array
 // When passed to tally
 // Then it should return an empty object
-test.todo("tally on an empty array returns an empty object");
+test("tally on an empty array returns an empty object", () =>{
+    expect(tally([])).toEqual({});
+});
 
 // Given an array with duplicate items
 // When passed to tally
 // Then it should return counts for each unique item
+test("tally When passed an array with duplicate items returns counts for each unique item", () =>{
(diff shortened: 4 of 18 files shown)


Sprint-2/implement/lookup.js
```diff
@@ -1,5 +1,11 @@
-function createLookup() {
+function createLookup(list) {
   // implementation here
+  let currency = {};
+  for(let i = 0; i < list.length; i++){
+    currency[list[i][0]] = list[i][1];
+  }
```
illicitonion, line 6: This looks good!

There are a couple of other ways you could write this which - let's call the way you wrote 1, and I'll suggest 2 and 3 below. Can you respond to this comment saying what you prefer about each of the three approaches? I don't think any one of them is obviously the best :)

2:
```suggestion
  for(let item of list){
    currency[item[0]] = item[1];
  }
```

3:
```suggestion
  for(let [key, value] of list){
    currency[key] = value;
  }
```

Sprint-2/implement/querystring.js
```diff
@@ -6,11 +6,12 @@ function parseQueryString(queryString) {
   const keyValuePairs = queryString.split("&");
 
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
-    queryParams[key] = value;
+    const [key, ...rest] = pair.split("=");
+    queryParams[key] = rest.join("=");
```
illicitonion, line 10: Nice fix!

Sprint-2/implement/tally.test.js
```diff
@@ -19,16 +19,26 @@ const tally = require("./tally.js");
...
 // Then it should return counts for each unique item
+test("tally When passed an array with duplicate items returns counts for each unique item", () =>{
+    expect(tally(["a", "b", "c", "a", "a", "c"])).toEqual({a : 3, b : 1, c : 2});
+});
 
 // Given an invalid input like a string
 // When passed to tally
 // Then it should throw an error
+test("tally When passed an invalid input like a string Then it should throw an error", () =>{
+    expect(tally("hello")).toThrow("Input must be an array");
```
illicitonion, line 43: This is really close, but not _quite_ right. We haven't actually taught about `throw`ing in the course, but it's useful to know about! But right now this test fails, and we don't like failing tests :)

When using jest, if you use `toThrow`, you need to pass a _function_ to `expect` rather than just a value - if you rewrite this as:
```suggestion
    expect(() => {tally("hello") }).toThrow("Input must be an array");
```

you should see a more useful failure message - can you make this change, and then fix the test so it passes?

Your actual code works fine, but the test code here isn't quite correct :)

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,17 @@
-function tally() {}
-
+function tally(list) {
+    let count = {};
+    if(list.length == 0){
+        return {};
+    }
```
illicitonion, line 5: Do you actually need this special-case of an early return? If you delete this code, will things break, or still work the same?

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,17 @@
...
+    if(!Array.isArray(list)){
+        throw new Error("invalid input, input must be an array");
+    }
+    else{
```
illicitonion, line 9: Because `throw` and `return` both immediately stop running your function, you can actually get rid of the else here - that can be useful for stopping your code always being more and more indented, and to be able to more clearly read the flow of the function. You can write this:
```js
if (!Array.isArray(list)) {
    throw new Error("...");
}
for (let item of list) {
  ...
}
return count;
```

The code you wrote is not wrong - it absolutely works fine, just this form may be a little more clear.

Sprint-2/interpret/invert.js
```diff
@@ -7,23 +7,35 @@
...
 function invert(obj) {
+  if (obj === null || obj === undefined) {
+    throw new Error("Input cannot be null or undefined");
+  }
+
   const invertedObj = {};
 
   for (const [key, value] of Object.entries(obj)) {
-    invertedObj.key = value;
+    invertedObj[value] = key;
```
illicitonion, line 17: This is really nice and clear code, and really clear answers - good job!

Sprint-2/stretch/count-words.js
```diff
@@ -26,3 +26,29 @@
...
+        keys[i] = keys[j];
+        keys[j] = temp;
+      }
+    }
+  }
+  const sortedObject = {};
+  for (let key of keys) {
+    sortedObject[key] = obj[key];
+  }
+  return sortedObject;
```
illicitonion, line 52: This works, but involves you having to write quite a lot of code...

Can you work out a way to do this same thing using `Object.fromEntries`, `Object.entries`, and `Array.sort`? I'm going to challenge you to do this in three lines of code.

Sprint-2/stretch/count-words.js
```diff
@@ -26,3 +26,29 @@
...
 3. Order the results to find out which word is the most common in the input
 */
+function countword(sentence){
+    let str = "";
+    str = sentence.replace(/[.,!?"]/g, '');
+    str = str.toLowerCase();
+    let words = str.split(" ").filter(word => word !== "");
+    let obj = {};
+    for (let item of words) {
+      obj[item] = (obj[item] || 0) + 1;
```
illicitonion, line 36: This is a nice way of adding one to a maybe-missing value.

Sprint-2/stretch/till.js
```diff
@@ -21,11 +32,16 @@ const till = {
   "20p": 10,
 };
 const totalAmount = totalTill(till);
+console.log(totalAmount);
 
+module.exports = totalTill;
 // a) What is the target output when totalTill is called with the till object
+// £NaN
```
illicitonion, line 39: The _target_ output is what we _expect_, not what we actually see - what do we actually expect the answer to be?

PR #72, opened 2024-11-24
West-Midlands-Community | Matthew Law | Module-Data-Groups | WEEK 7

matthewlawpixel:
Self checklist

## Changelist

PR is for completing exercises.

## Questions

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -3,9 +3,13 @@
 // If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
 
 function calculateMedian(list) {
+  const sortedList = [...list].sort((a, b) => a - b);
   const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  if (sortedList.length % 2 === 0) {
+    return (sortedList[middleIndex - 1] + sortedList[middleIndex]) / 2;
+  } else {
+    return sortedList[middleIndex];
+  }
 }
 
 module.exports = calculateMedian;


Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,51 @@ const sum = require("./sum.js");
...
+    expect(sum([-10, -20, 30])).toBe(0);
+    expect(sum([-1, -2, -3])).toBe(-6);
+});
+
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
 
+test("given an array with decimal/float numbers, returns the correct total sum", () => {
+    expect(sum([1.5, 2.5, 3.5])).toBe(7.5);
```
cjyuan, line 43: Decimal numbers in most programming languages (including JS) are internally represented in "floating point number" format. Floating point arithmetic is not exact. For example, the result of `46.5678 - 46 === 0.5678` is false because `46.5678 - 46` only yield a value that is very close to `0.5678`.

Note: Values with decimal part equals to `.5` happened to be able to be represented exactly in floating point number format.

Sprint-1/stretch/aoc-2018-day1/solution.js
```diff
@@ -0,0 +1,34 @@
...
+    const part1Result = input.reduce((a, b) => a + b, 0);
+    console.log("Part 1 - Resulting Frequency:", part1Result);
+
+    // Part 2: Find the first frequency reached twice
+    let sum = 0;
+    let set = new Set([0]);
+    let length = input.length;
+    let part2Result = null;
+
+    for (let i = 0; ; i = (i + 1) % length) {
```
cjyuan, line 20: How does one get Part 2 spec? I only saw Part 1 spec.

Why not use the "typical form" of for loop, `for (let i = 0; i < length; i++) {` ?

Sprint-1/stretch/aoc-2018-day1/solution.js
```diff
@@ -0,0 +1,34 @@
...
+                    .split('\n')
+                    .map(i => parseInt(i, 10));
+
+    // Part 1: Calculate the resulting frequency
+    const part1Result = input.reduce((a, b) => a + b, 0);
+    console.log("Part 1 - Resulting Frequency:", part1Result);
+
+    // Part 2: Find the first frequency reached twice
+    let sum = 0;
+    let set = new Set([0]);
```
cjyuan, line 16: Why start with a set with one element, 0, instead of an empty set?

cjyuan (commented):
Code is good. Just a few comments.

Sprint-1/implement/sum.test.js
matthewlawpixel: Thank you @cjyuan, i did some research in the matter and have found that using .toBeCloseTo() would fix the problem of the test failing

    expect(sum([1.1, 1.1, 1.1])).toBeCloseTo(3.3, 5);

The 5 is the tolerance for decimal places. In this case it would mean that if the result is off by 0.00001, the test will still pass.

Sprint-1/stretch/aoc-2018-day1/solution.js
matthewlawpixel: The reason why we use 0 instead of an empty set is so that 0 is remembered as the starting frequency and won't be added into the set later and doesn't register it as a repeat and allows for the program to correctly detect when the frequency reaches 0 again.

Sprint-1/stretch/aoc-2018-day1/solution.js
matthewlawpixel: Part 2 is to find the first repeated frequency which loops through the numbers and adding up to a running total, if the total has been encountered before then the loop ends. The loop ensures the program keeps running until a repeated frequency is found

matthewlawpixel: The reason for the infinite for loop using a modulo is because the requirement is to repeat the input list until finding a repeating frequency. A normal for loop would only iterate over the list once and in some cases you would need to loop back to the beginning of the list to continue summing up the frequencies. Using the modulo operation (i = (i + 1) % length) allows us to wrap around the input list, repeating it without having to manually reset the loop or count the iterations.

PR #73, opened 2024-11-24
London | Saqib_Javed | Module-Data-Groups | Sprint 1 exercises | week-1

saqibjvd:
## Learners, PR Template

Self checklist

- [x ] I have committed my files one by one, on purpose, and for a reason
- [ x] I have titled my PR with COHORT_NAME | FIRST_NAME LAST_NAME | REPO_NAME | WEEK 
- [ x] I have tested my changes
- [x ] My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/contributing/)
- [x ] My changes meet the [requirements](./README.md) of this task

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-1/implement/dedupe.js
+++ b/Sprint-1/implement/dedupe.js
@@ -1,1 +1,19 @@
-function dedupe() {}
+function dedupe(arr) {
+     const removeRepeated = arr.filter((element, index) => {
+       // console.log("element >>", element, "index >>", index) 
+       
+        return arr.indexOf(element) === index
+     })
+   console.log(removeRepeated)
+   return removeRepeated
+
+}
+
+console.log(dedupe(['a','a','a','b','b','c', 'c','d','d' ]))
+console.log(dedupe(['1','2','3','4','5']))
+console.log(dedupe(['1','a','b','b','1']))
+
+module.exports = dedupe;
+
+// new Set(); can used for bigger and complex dedupe.
+//its more efficient than checking with indexOf
--- a/Sprint-1/implement/dedupe.js
+++ b/Sprint-1/implement/dedupe.js
@@ -1,10 +1,9 @@
 function dedupe(arr) {
      const removeRepeated = arr.filter((element, index) => {
-       // console.log("element >>", element, "index >>", index) 
-       
+      
         return arr.indexOf(element) === index
      })
-   console.log(removeRepeated)
+  
    return removeRepeated
 
 }
--- a/Sprint-1/implement/max.js
+++ b/Sprint-1/implement/max.js
@@ -1,4 +1,28 @@
+
+// check max in elements E.g. max([30, 50, 10, 40]), target output: 50
+// E.g. max(['hey', 10, 'hi', 60, 10]), target output: 60 (sum ignores any non-numerical elements)
+
+// we can use following methods to find max in a element = map, max and reduce method.
+
 function findMax(elements) {
+   const numbers = elements.filter(elements => typeof elements === "number");
+   console.log("only numbers", numbers)
+   
+   if(numbers.length === 0) {
+    return -Infinity
+   }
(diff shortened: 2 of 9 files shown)


Sprint-1/implement/dedupe.js
```diff
@@ -1 +1,19 @@
-function dedupe() {}
+function dedupe(arr) {
+     const removeRepeated = arr.filter((element, index) => {
+       // console.log("element >>", element, "index >>", index) 
+       
+        return arr.indexOf(element) === index
+     })
+   console.log(removeRepeated)
```
cjyuan, line 7: This statement should not belong inside the function. 

Please note that the way you construct `removeRepeated` works, but this "algorithm" will perform poorly on huge array because `arr.indexOf(element)` needs to search for `element` in the array from the start of the array for every element in the array.

Sprint-1/implement/max.test.js
```diff
@@ -16,28 +16,59 @@ const findMax = require("./max.js");
...
+test("given an array with non-number values, returns the max and ignores non-numeric values", () => {
+    const result = findMax(['hey', 10, 'hi', 60, 10]);
+    expect(result).toEqual(60);
+});
+
 
 // Given an array with only non-number values
 // When passed to the max function
 // Then it should return the least surprising value given how it behaves for all other inputs
+test("given an array with only non-number values, returns -Infinity", () => {
```
cjyuan, line 71: Could also include other types of non-number values. For examples, `undefined`, `true`.

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,28 @@
...
 function findMax(elements) {
+   const numbers = elements.filter(elements => typeof elements === "number");
+   console.log("only numbers", numbers)
+   
+   if(numbers.length === 0) {
+    return -Infinity
+   }
+   const max = Math.max(...numbers)
+   console.log("max", max)
+   return max
```
cjyuan, line 16: What do you expect from the following function calls (on extreme cases)?

```
findMax([-Infinity])
findMax([-Infinity, 0])
findMax([NaN])
findMax([0, NaN, 1])
```

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,36 @@
...
+  const numbers = elements.filter(ele => typeof ele === "number");
+    //console.log(numbers)
+  if (numbers.length === 0) {
+    return 0;
+  }
+
+  let sum = 0;
+  numbers.forEach((el) => (sum += el));
+  //console.log(sum);
+  return sum;
```
cjyuan, line 16: What do you expect from the following function calls (on extreme cases)?

```
sum([Infinity, 1]);
sum([Infinity, Infinity]);
sum([NaN, 1]);
sum([Infinity, -Infinity]);
```

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,47 @@ const sum = require("./sum.js");
...
+    const result = sum([10, -5, 15, -10]);
+    expect(result).toEqual(10);
+  });
 
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
+test("Given an array with decimal numbers, should return the correct total sum", function() {
+    const result = sum([10.5, 20.3, -5.2, 5]);
+    expect(result).toEqual(30.6); // 10.5 + 20.3 - 5.2 + 5 = 30.6
```
cjyuan, line 42: Decimal numbers in most programming languages (including JS) are internally represented in "floating point number" format. Floating point arithmetic is not exact. For example, the result of `46.5678 - 46 === 0.5678` is false because `46.5678 - 46` only yield a value that is very close to `0.5678`.

cjyuan (commented):
Code is good. Just got a few suggestions and challenges for you.

PR #86, opened 2024-11-28
London | Anna Fedyna | Module-Data-Groups | Week 3 | To Do list app

annafedyna:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

Are there any best practices or patterns that I could apply to improve the maintainability of my code?

### DIFF
--- a/Sprint-3/todo-list/script.js
+++ b/Sprint-3/todo-list/script.js
@@ -1,83 +1,72 @@
-const removeAllButton = document.getElementById("remove-all-completed"),
-  addNewTodoButton = document.querySelector("add-todo"),
-  form = document.getElementById("todo-form");
+const removeAllButton = document.getElementById("remove-all-completed");
+const form = document.getElementById("todo-form");
+const input = document.querySelector(".input");
+const todoList = document.getElementById("todo-list");
 
-function populateTodoList(todos) {
-  let list = document.getElementById("todo-list");
-  // Create todo list elements with completed and delete buttons here, all todos should display inside the "todo-list" element.
-  todos.forEach((item) => {
-    newTodo = document.createElement("li");
-    newTodo.innerHTML = `<span class="badge bg-primary rounded-pill">${item.task}
-    <i class="fa fa-check" aria-hidden="true"></i>
-    <i class="fa fa-trash" aria-hidden="true"></i>
-    </span>
-    <button class="status">DONE</button>
-    <button class="delete">DELETE</button>`;
-    // newTodo.style.textDecoration = "line-through";
-    list.append(newTodo);
-  });
-}
 let todos = [
   { task: "Wash the dishes", completed: false },
   { task: "Do the shopping", completed: false },
 ];
-populateTodoList(todos);
 
-// This function will take the value of the input field and add it as a new todo to the bottom of the todo list. These new todos will need the completed and delete buttons adding like normal.
+// Populate the todo list with initial tasks
+function populateTodoList(todos) {
+  todos.forEach((item) => {
+    addTodoToDOM(item.task, item.completed);
+  });
+}
+
+// Add todo item to the DOM
+function addTodoToDOM(task, completed = false) {
+  const newTodo = document.createElement("li");
+  newTodo.classList.add("todo-item");
+  if (completed) newTodo.classList.add("done");
+  newTodo.innerHTML = `
+    <span class="badge bg-primary rounded-pill">${task}</span>
+    <button class="status">${completed ? "UNDO" : "DONE"}</button>
+    <button class="delete">DELETE</button>
+  `;
+  todoList.appendChild(newTodo);
+}
+
+// Add new todo from the input field
 function addNewTodo(event) {
   event.preventDefault();
-  const input = document.querySelector(".input");
-  if (input.value) {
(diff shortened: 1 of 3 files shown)


Sprint-1/fix/median.js
illicitonion, line 1: This file isn't from this exercise - you may want to brush up on branch management to make sure you're always making new branches based on `main`, rather than other work you've done.

Sprint-3/todo-list/script.js
```diff
@@ -1,25 +1,72 @@
...
+  todos.forEach((item) => {
+    addTodoToDOM(item.task, item.completed);
+  });
+}
+
+// Add todo item to the DOM
+function addTodoToDOM(task, completed = false) {
+  const newTodo = document.createElement("li");
+  newTodo.classList.add("todo-item");
+  if (completed) newTodo.classList.add("done");
```
illicitonion, line 22: This works, but a lot of people always use `{}` around the body of things like `if` statements. You _need_ to do this if you have more than one statement in the body, but a lot of people always do. Have a read of https://www.blackduck.com/blog/understanding-apple-goto-fail-vulnerability-2.html for an example of why.

Sprint-3/todo-list/script.js
```diff
@@ -1,25 +1,72 @@
...
+  newTodo.innerHTML = `
+    <span class="badge bg-primary rounded-pill">${task}</span>
```
illicitonion, line 24: This works as long as we trust our users, but there's something called an injection bug, that setting innerHTML using user input can trigger.

We got the value of `task` from the user - they put it in a form. And then we are adding it, exactly as the user typed it, as HTML into our page which will then get rendered by the browser.

Have a read of https://medium.com/@gregadiaz89/dom-innerhtml-vulnerability-8821a03ef2b8 to learn a bit more.

Sometimes we need to use `innerHTML`, but a lot of the time we try to use something like `innerText` to avoid the browser interpreting our text as HTML.

Can you work out how to achieve this same result as you are now, but where we don't include `task` as part of an `innerHTML` property?

Sprint-3/todo-list/script.js
```diff
@@ -1,25 +1,72 @@
...
 
-// This function will take the value of the input field and add it as a new todo to the bottom of the todo list. These new todos will need the completed and delete buttons adding like normal.
+// Add new todo from the input field
 function addNewTodo(event) {
-  // The code below prevents the page from refreshing when we click the 'Add Todo' button.
   event.preventDefault();
-  // Write your code here... and remember to reset the input field to be blank after creating a todo!
+
+  const taskValue = input.value.trim();
+  if (taskValue) {
```
illicitonion, line 36: Often if someone presses a button, they expect it to do something, and if it quietly doesn't, they think things are broken.

If they haven't typed _any_ text into the text box, and press the button, they maybe don't expect anything to happen. But if they typed a few spaces into the box, and pressed the button, they maybe expected a new to do item to appear.

Your code just doesn't do anything if they put in spaces and press the button.

Maybe this is fine, or maybe it would be useful to indicate to them why you didn't do anything when they pressed the button.

What do you think? How could you maybe tell them something is wrong? Or maybe it's fine to just do nothing - but something to think about.

illicitonion (commented):
This is looking really good - I've left a few comments with things to think about, but the core is working well and looking good!

Sprint-1/fix/median.js
annafedyna: Thanks, it should not be there,I made new branch from main and first commit does not have this file but somehow added with second one

annafedyna:
> This is looking really good - I've left a few comments with things to think about, but the core is working well and looking good!

Thank you for the review,  really appreciate all the comments)

Sprint-3/todo-list/script.js
annafedyna: good point, thanks)

Sprint-3/todo-list/script.js
annafedyna: Thanks, important point, heard about it but never practiced.
I refactored the code.

Sprint-3/todo-list/script.js
annafedyna: Thank you! I added alert for empty text.

Sprint-1/fix/median.js
illicitonion: There are a couple of ways you could handle this.

The nicest (but scariest) would be to remove the commit from this branch. You can do that by running `git rebase -i $(git merge-base HEAD origin/main)` and removing the line associated with the accidental commit. Interactive rebases are really powerful and useful, so good to learn about. The scary thing about this is you'll have to run `git push --force-with-lease` once after doing so, because when you're pushing you're telling git "I know this doesn't look like what was already pushed, but I really want to replace it, even if that means throwing away something that was already pushed".

An alternative is you could add a new commit reverting the commit (`git revert 5edb8bd14bb723685062d2997652a462925dad3b`). This means the files won't show up as changed in the "Files changed" view of the pull request (yay!), but you'll have two commits in your history, one changing them, and another undoing the change.

annafedyna: Thank you, removed unnecessary commit with rebase.

PR #89, opened 2024-11-28
Glasgow| Iakub Dubachev | Module-Data-Groups | Sprint-1| Week-1 

IakubDubachev:
Self checklist

Changelist

In this PR, I refactored and fixed multiple functions to enhance their functionality and ensure they operate as intended. I made specific updates to the logic to account for edge cases and verified that all tests run successfully, confirming that every test case passes. These improvements make the code more robust and reliable, with the tests now accurately validating the updated functionality.

Questions
No Questions

### DIFF
--- a/Sprint-1/implement/max.js
+++ b/Sprint-1/implement/max.js
@@ -1,4 +1,10 @@
-function findMax(elements) {
-}
-
-module.exports = findMax;
+function findMax(arr) {
+    if (!Array.isArray(arr)) {
+      throw new TypeError("Input must be an array");
+    }
+    const numericValues = arr.filter(item => typeof item === 'number');
+    return numericValues.length > 0 ? Math.max(...numericValues) : -Infinity;
+  }
+  
+  module.exports = findMax;
+  
--- a/Sprint-1/implement/sum.js
+++ b/Sprint-1/implement/sum.js
@@ -1,4 +1,9 @@
-function sum(elements) {
-}
-
-module.exports = sum;
+function sum(arr) {
+    if (!Array.isArray(arr)) {
+      throw new TypeError("Input must be an array");
+    }
+    return arr.reduce((total, item) => (typeof item === "number" ? total + item : total), 0);
+  }
+  
+  module.exports = sum;
+  
--- a/Sprint-1/implement/sum.test.js
+++ b/Sprint-1/implement/sum.test.js
@@ -1,36 +1,32 @@
-/* Sum the numbers in an array
-
-In this kata, you will need to implement a function that sums the numerical elements of an array
-
-E.g. sum([10, 20, 30]), target output: 60
-E.g. sum(['hey', 10, 'hi', 60, 10]), target output: 80 (ignore any non-numerical elements)
-*/
-
 const sum = require("./sum.js");
 
-// Acceptance Criteria:
 
-// Given an empty array
-// When passed to the sum function
-// Then it should return 0
-test.todo("given an empty array, returns 0")
+test("given an empty array, returns 0", () => {
+  expect(sum([])).toBe(0);
+});
 
-// Given an array with just one number
-// When passed to the sum function
-// Then it should return that number
 
-// Given an array containing negative numbers
-// When passed to the sum function
-// Then it should still return the correct total sum
+test("given an array with just one number, returns that number", () => {
+  expect(sum([42])).toBe(42);
+});
 
-// Given an array with decimal/float numbers
-// When passed to the sum function
-// Then it should return the correct total sum
 
-// Given an array containing non-number values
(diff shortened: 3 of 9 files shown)


Sprint-1/implement/max.js
```diff
@@ -1,4 +1,10 @@
-function findMax(elements) {
-}
-
-module.exports = findMax;
+function findMax(arr) {
+    if (!Array.isArray(arr)) {
+      throw new TypeError("Input must be an array");
+    }
+    const numericValues = arr.filter(item => typeof item === 'number');
+    return numericValues.length > 0 ? Math.max(...numericValues) : -Infinity;
```
cjyuan, line 6: What do you expect from the following function calls (on extreme cases)?

```
findMax([-Infinity])
findMax([-Infinity, 0])
findMax([NaN])
findMax([0, NaN, 1])
```

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,9 @@
-function sum(elements) {
-}
-
-module.exports = sum;
+function sum(arr) {
+    if (!Array.isArray(arr)) {
+      throw new TypeError("Input must be an array");
+    }
+    return arr.reduce((total, item) => (typeof item === "number" ? total + item : total), 0);
```
cjyuan, line 5: What do you expect from the following function calls (on extreme cases)?

```
sum([Infinity, 1]);
sum([Infinity, Infinity]);
sum([NaN, 1]);
sum([Infinity, -Infinity]);
```

Sprint-1/implement/sum.test.js
```diff
@@ -1,36 +1,32 @@
...
+  expect(sum([-10, 20, -30, 40])).toBe(20);
+});
+
 
-// Given an empty array
-// When passed to the sum function
-// Then it should return 0
-test.todo("given an empty array, returns 0")
+test("given an array with decimal/float numbers, returns the correct total sum", () => {
+  expect(sum([1.5, 2.5, 3.5])).toBe(7.5);
```
cjyuan, line 20: Decimal numbers in most programming languages (including JS) are internally represented in "floating point number" format. Floating point arithmetic is not exact. For example, the result of `46.5678 - 46 === 0.5678` is false because `46.5678 - 46` only yield a value that is very close to `0.5678`.

cjyuan (commented):
Code is good. I just left you some suggestions and challenges.

Sprint-1/implement/max.js
IakubDubachev: findMax([-Infinity])

Expected: -Infinity
Reasoning: Only -Infinity in the array, so it’s the maximum.
findMax([-Infinity, 0])

Expected: 0
Reasoning: The maximum between -Infinity and 0 is 0.
findMax([NaN])

Expected: NaN
Reasoning: NaN is not comparable, so it stays as the maximum.
findMax([0, NaN, 1])

Expected: 1
Reasoning: NaN is ignored, so the maximum is 1.

Sprint-1/implement/sum.js
IakubDubachev: sum([Infinity, 1])

Expected: Infinity
Reasoning: Adding any finite number to Infinity results in Infinity.
sum([Infinity, Infinity])

Expected: Infinity
Reasoning: Adding Infinity to Infinity still results in Infinity.
sum([NaN, 1])

Expected: NaN
Reasoning: Adding NaN to any number results in NaN.
sum([Infinity, -Infinity])

Expected: NaN

Sprint-1/implement/sum.test.js
IakubDubachev: For floating-point numbers, use an epsilon-based comparison to account for the inaccuracies inherent in floating-point arithmetic. This ensures more reliable tests for equality.

Sprint-1/implement/max.js
cjyuan: > Expected: NaN
Reasoning: NaN is not comparable, so it stays as the maximum.
findMax([0, NaN, 1])

>Expected: 1
Reasoning: NaN is ignored, so the maximum is 1.

First one says `NaN` stays as the maximum, the second one says `NaN` is ignored. Which should it be?

Sprint-1/implement/sum.test.js
cjyuan: Epsilon-based comparison works, but there is also a function in Jest similar to `.toBe()` which you can use.

In any case, can you modify your code to use one of these method?

cjyuan:
It will be nice if you can update `Sprint-1/implement/sum.test.js` to handle floating point number comparison.

Other than that, feel free to mark this as completed anytime. 

Well done.

PR #92, opened 2024-11-29
West Midlands | Samira Hekmati | Module-Data-Groups | Week 1

samirahekmati:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-1/implement/sum.test.js
+++ b/Sprint-1/implement/sum.test.js
@@ -13,7 +13,6 @@ const sum = require("./sum.js");
 // Given an empty array
 // When passed to the sum function
 // Then it should return 0
-test.todo("given an empty array, returns 0")
 
 // Given an array with just one number
 // When passed to the sum function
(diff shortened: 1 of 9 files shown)


Sprint-1/implement/sum.test.js
```diff
@@ -34,3 +33,23 @@ test.todo("given an empty array, returns 0")
...
+        expect(sum([])).toEqual(0)
+    });
+    test("Given an array with just one number,it should return that number", () =>{
+        expect(sum([20])).toEqual(20)
+    });
+    test("Given an array containing negative numbers,it should still return the correct total sum", () =>{
+        expect(sum([-6,-1,3])).toEqual(-4)
+    });
+    test("Given an array with decimal/float numbers,it should return the correct total sum", () =>{
+        expect(sum([3.1, 2.1, 5])).toEqual(10.2)
```
cjyuan, line 47: Decimal numbers in most programming languages (including JS) are internally represented in "floating point number" format. Floating point arithmetic is not exact. For example, the result of `46.5678 - 46 === 0.5678` is false because `46.5678 - 46` only yield a value that is very close to `0.5678`.

Sprint-1/stretch/aoc-2018-day1/solution.js
```diff
@@ -0,0 +1,22 @@
+const fs = require("fs"); // Import the File System module
+
+// Read the file content
+const input = fs.readFileSync("input.txt", "utf8"); // 'utf8' ensures the data is read as a string
```
cjyuan, line 4: "'utf8' ensures the data is read as a string" is not the reason why one specifies "utf8".

Sprint-1/stretch/aoc-2018-day1/solution.js
```diff
@@ -0,0 +1,22 @@
...
+
+console.log("Current directory:", process.cwd());
+
+const inputArray = input.split("\n");
+console.log(inputArray);
+
+const numberArray = inputArray.map(Number);
```
cjyuan, line 11: Can you explain what the expression `inputArray.map(Number)` does? Why do you pass `Number` as the parameter to `inputArray.map()`?

cjyuan (commented):
I left some questions and suggestions for you.

Sprint-1/stretch/aoc-2018-day1/solution.js
samirahekmati: When Number is passed as the callback to map(), it is invoked on each element of the array.
For each element x in inputArray, Number(x) is called, converting x to a number, and the result is added to the new array.

PR #103, opened 2024-12-01
London | Ameneh Keshavarz | Module-Data-Groups | WEEK1

Ameneh-Keshavarz:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -3,6 +3,9 @@
 // If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
 
 function calculateMedian(list) {
+  if(list.length===0)
+    return null;
+  sortList(list);
   let median;
   if(list.length%2===1){
   const middleIndex = Math.floor(list.length / 2);
@@ -17,6 +20,18 @@ function calculateMedian(list) {
 
 
   return median;
+}
+
+function sortList(list){
+   for (let i = 0; i < list.length - 1; i++) {
+    for (let j = 0; j < list.length - i - 1; j++) {
+      if (list[j] > list[j + 1]) {
+        let temp = list[j];
+        list[j] = list[j + 1];
+        list[j + 1] = temp;
+      }
+    }
+  }
 }
 
 module.exports = calculateMedian;
--- a/Sprint-1/implement/dedupe.test.js
+++ b/Sprint-1/implement/dedupe.test.js
@@ -10,7 +10,7 @@ E.g. dedupe(['a','a','a','b','b','c']) target output: ['a','b','c']
 E.g. dedupe([5, 1, 1, 2, 3, 2, 5, 8]) target output: [5, 1, 2, 3, 8]
 E.g. dedupe([1, 2, 1]) target output: [1, 2]
 */
-describe("calculateMedian", () => {
+describe("dedupe", () => {
 
 
 // Acceptance Criteria:
--- a/Sprint-1/implement/max.test.js
+++ b/Sprint-1/implement/max.test.js
@@ -63,4 +63,17 @@ test("given an array with non-number values, returns the max and ignores non-num
 test("given an array with only non-number values, returns -Infinity", () => {
   expect(findMax([ 'hey', 'hi', 'hello' ])).toEqual(-Infinity);
 });
+
+test("given an array with non-number values, returns the max and ignores non-numeric values", () => {
+  expect(findMax([ 1,"60"])).toEqual(1);
+});
+
+test("given an array with non-number values, returns the max and ignores non-numeric values", () => {
+  expect(findMax([ 1,"200","hey"])).toEqual(1);
+});
+
+test("given an array with invalid values, returns -Infinity", () => {
+  expect(findMax([null])).toEqual(-Infinity);
+});
+
 });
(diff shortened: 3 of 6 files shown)


Sprint-1/fix/median.js
```diff
@@ -3,8 +3,19 @@
 // If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
 
 function calculateMedian(list) {
+  let median;
```
cjyuan, line 6: The arrays specified in `median.test.js` happened to be sorted.

Sprint-1/implement/dedupe.test.js
```diff
@@ -10,18 +10,28 @@ E.g. dedupe(['a','a','a','b','b','c']) target output: ['a','b','c']
 E.g. dedupe([5, 1, 1, 2, 3, 2, 5, 8]) target output: [5, 1, 2, 3, 8]
 E.g. dedupe([1, 2, 1]) target output: [1, 2]
 */
+describe("calculateMedian", () => {
```
cjyuan, line 13: Are you testing `calculateMedian()`?

Sprint-1/implement/max.test.js
```diff
@@ -11,33 +11,56 @@ We have set things up already so that this file can see your function from the o
...
 // Then it should return the largest decimal number
+test("given an array with decimal numbers, returns the largest decimal number", () => {
+  expect(findMax([2.5, 4.1, 3.3, 3.0])).toEqual(4.1);
+});
 
 // Given an array with non-number values
 // When passed to the max function
 // Then it should return the max and ignore non-numeric values
+test("given an array with non-number values, returns the max and ignores non-numeric values", () => {
+  expect(findMax([ 'hey', 10, 'hi', 60, 10 ])).toEqual(60);
```
cjyuan, line 56: What do you expect from the following function calls?
Does your function returns the value you expected? (Surprised!)
```
findMax([ 10, "60" ])
findMax([ 1, "200", "hey" ])
findMax([ null ])
```

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,12 @@
 function sum(elements) {
+        let total = 0;
+    for (let item of elements) {
+        if (typeof item === 'number') {
+            total += item; 
+        }
+    }
+    return total;
+
```

```
sum([Infinity, 1]);
sum([Infinity, Infinity]);
sum([NaN, 1]);
sum([Infinity, -Infinity]);
```

Sprint-1/implement/sum.test.js
```diff
@@ -7,30 +7,51 @@ E.g. sum(['hey', 10, 'hi', 60, 10]), target output: 80 (ignore any non-numerical
...
 // Then it should still return the correct total sum
+test("given an array containing negative numbers, returns the correct total sum", () => {
+    expect(sum([-10, -20, 30])).toEqual(0);
+});
 
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
+test("given an array with decimal/float numbers, returns the correct total sum", () => {
+    expect(sum([1.5, 2.5, 3.5])).toEqual(7.5);
```

Sprint-1/stretch/aoc-2018-day1/solution.js
```diff
@@ -0,0 +1,18 @@
+const fs = require('fs');
+const filePath = 'input.txt'; 
+
+function calculateFrequencyFromFile(filePath) {
+  const data = fs.readFileSync(filePath, 'utf-8');
+  
+  const changes = data.split('\n').map(Number);
```
cjyuan, line 7: Can you explain what `.map(Number)` does? Seemed relatively advanced!

cjyuan (commented):
Code is good, but `findMax()` (in `max.js`) is not quite correct. Can you fix the function?

Sprint-1/implement/max.test.js
Ameneh-Keshavarz: Thanks, I updated the code for your review

Sprint-1/implement/sum.js
Ameneh-Keshavarz: I updated the code, thanks

Sprint-1/implement/sum.test.js
Ameneh-Keshavarz: Thanks, I updated the code.

Sprint-1/stretch/aoc-2018-day1/solution.js
Ameneh-Keshavarz: The map() method applies a function to each element of an existing array, but it does not change the original array. Instead, it returns a new array with the results of applying the function. I studied a little about Collections before, and I learnt about it at that time.

cjyuan:
In `max.js`:
One of the checks in 
  `typeof current === 'number' && !isNaN(current) && current !== null && current > max`
is unnecessary.

In `median.js`: 
1) Array is passed to a function ***by reference***  (instead of ***by value***). This implies when an array is passed to a function via parameter, the array is **shared** by the caller and the callee (the function being called).

2) Most programming languages have a built-in optimised built-in sorting function. You may want to look up the `.sort()` method of Array in JS.

In `sum.js`:
If you want to check if a number is `NaN` or `Infinity`, you can use the `isFinite()` function.

---

Code is good. Change is optional.

PR #104, opened 2024-12-01
Manchester | Zabihollah Namazi | Module-Data-Groups | Alarm Clock | Sprint 3

ZabihollahNamazi:
alarmclock

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -1,4 +1,37 @@
-function setAlarm() {}
+var input_time = document.querySelector("#alarmSet");
+var title_time = document.querySelector("#titleTime");
+var timeRemaining = document.querySelector("#timeRemaining");
+
+function setAlarm() {
+  
+  let totalSeconds = parseInt(input_time.value, 10);
+  if (isNaN(totalSeconds) || totalSeconds <= 0) {
+    alert("Please enter a valid number of seconds.");
+    return;
+  }
+  
+  const title_second = input_time.value % 60;
+  const title_minute = (input_time.value - title_second) / 60;
+
+  title_time.innerText = `${title_minute.toString().padStart(2, "0")}:${title_second.toString().padStart(2, "0")}`;
+
+  let timer = setInterval(() =>{
+    let seconds = totalSeconds % 60;
+    let minutes = (totalSeconds - seconds) / 60;
+    timeRemaining.innerText = `Time Remaining: ${minutes.toString().padStart(2, "0")}:${seconds.toString().padStart(2, "0")}`;
+    
+    if(totalSeconds <= 0){
+      clearInterval(timer);
+      playAlarm();
+    }
+    else{
+      totalSeconds--;
+    }
+    
+  }, 1000);
+
+
+}
 
 // DO NOT EDIT BELOW HERE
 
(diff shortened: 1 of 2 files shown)


Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,16 +1,58 @@
-function setAlarm() {}
+var input_time = document.querySelector("#alarmSet");
+var title_time = document.querySelector("#titleTime");
+var timeRemaining = document.querySelector("#timeRemaining");
+var bUserStopped = false;
```
illicitonion, line 4: You're using three different naming conventions for your variables here: camelCase, snake_case, and bHungarianNotation.

I'd recommend picking one naming convention to make things more consistent and easy to read. In JavaScript, most people use camelCase as the naming convention for variable names, so I'd recommend following that unless you have a reason to do something different.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,16 +1,58 @@
...
+  
+  const title_second = input_time.value % 60;
+  const title_minute = (input_time.value - title_second) / 60;
+
+  title_time.innerText = `${title_minute.toString().padStart(2, "0")}:${title_second.toString().padStart(2, "0")}`;
+
+  let timer = setInterval(() =>{
+    let seconds = totalSeconds % 60;
+    let minutes = (totalSeconds - seconds) / 60;
+    timeRemaining.innerText = `Time Remaining: ${minutes.toString().padStart(2, "0")}:${seconds.toString().padStart(2, "0")}`;
```
illicitonion, line 21: You've written the same code (to take a number and convert it to a padded string) twice here, and two more times above for the title.

Can you think how you could avoid writing that same code four times?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,16 +1,58 @@
...
+      playAlarm();
+      document.getElementById("set").disabled = false;
+    }
+    else if(bUserStopped){
+      clearInterval(timer);
+      document.getElementById("set").disabled = false;
+
+    }
+    else{
+      document.getElementById("set").disabled = true;
```
illicitonion, line 34: Interestingly, this only disables the button one second after the timer starts running (and only updates the displayed time 1 second after pressing the button) - I think most users would expect the time displayed to _instantly_ change when they press the button, and for the button to become disabled instantly too.

Can you work out why there's the delay and fix it?

Sprint-3/alarmclock/index.html
```diff
@@ -4,7 +4,7 @@
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <link rel="stylesheet" href="style.css" />
-    <title>Title here</title>
+    <title id="titleTime">Title here</title>
```
illicitonion, line 7: This works and is fine, but because there's ever only one `<title>` in a page, we typically don't assign it an ID, but instead use a different query selector to find it - if you remove the ID here, can you make the code still work?

illicitonion (commented):
This is generally looking good, and is a good use of the DOM APIs, but I've left a few comments about some things to consider - please take a look and see what you can improve!

Sprint-3/alarmclock/index.html
ZabihollahNamazi: fixed it 
thanks

Sprint-3/alarmclock/alarmclock.js
ZabihollahNamazi: tried to fix it and looks ok for me now. but maybe I'm wrong , can you please check again?
thanks

Sprint-3/alarmclock/alarmclock.js
ZabihollahNamazi: I changed them all to camelCases
thanks

PR #110, opened 2024-12-02
CYF London | Fatma Arslantas | Module-Data-Groups | Week 1

AFatmaa:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -3,9 +3,26 @@
 // If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+
+  // 〰️ Create a copy of the input list to avoid modifying the original array
+  const copiedList = [...list];
+
+  // 〰️ Calculate the index of the middle element
+  const middleIndex = Math.floor(copiedList.length / 2)
+
+  // 〰️ If the list length is odd, return the middle element
+  if (copiedList.length % 2 === 1) {
+    const median = copiedList.splice(middleIndex, 1)[0];
+    return median;
+  }
+
+  // 〰️ If the list length is even, calculate the average of the two middle elements
+  const median1 = copiedList[middleIndex]
+  const median2 = copiedList[middleIndex -1] 
+  return (median1 + median2) / 2;
 }
+
+console.log(calculateMedian([1, 2, 3, 4, 5, 6]));
+console.log(calculateMedian([1, 2, 3, 4, 5]));
 
 module.exports = calculateMedian;
--- a/Sprint-1/implement/max.js
+++ b/Sprint-1/implement/max.js
@@ -1,4 +1,24 @@
 function findMax(elements) {
+
+// 〰️ Filter the array to include only numbers
+  const numericElements = elements.filter((el) => typeof el === "number");
+
+  // 〰️ Return -Infinity for an empty array (or no numeric values)
+  if (numericElements.length === 0) {
+    return -Infinity;
+  }
+
+  // 〰️ Use Math.max to find the largest number in the array
+  return Math.max(...numericElements);
+
 }
+
+console.log(findMax([]));
+console.log(findMax([42]));
+console.log(findMax([30, -10, 20, -50]));
+console.log(findMax([-30, -10, -50]));
+console.log(findMax([1.5, 2.7, 2.1]));
+console.log(findMax(["hello", 10, true, 60, null]));
+console.log(findMax(["hello", true, null, undefined]));
 
 module.exports = findMax;
--- a/Sprint-1/implement/sum.js
+++ b/Sprint-1/implement/sum.js
@@ -1,4 +1,31 @@
 function sum(elements) {
+
+    // 〰️ Ensure the input is an array
+    if (!Array.isArray(elements)) {
+        return 0; // 〰️ Return 0 if input is not an array
+    }
+
+    let total = 0; // 〰️ Initialize the sum
+
+    for (let i = 0; i < elements.length; i ++) {
+        if (typeof elements[i] === "number") { // 〰️ Check if the element is a number
+            total += elements[i]; // 〰️ Add the number to the total
+        }
+    }
+
+    return total; // 〰️ Return the total sum
+    
+
 }
+
+console.log(sum([]));
+console.log(sum([42]));
+console.log(sum([-10, -20, 30]));
+console.log(sum([1.5, 2.5, 3.5]));
+console.log(sum(['hello', 10, 'world', 20]));
+console.log(sum(['a', 'b', 'c', null, undefined, {}]));
+console.log(sum("not an array"));
+console.log(sum(123));
+
 
 module.exports = sum;
--- a/Sprint-1/implement/sum.test.js
+++ b/Sprint-1/implement/sum.test.js
@@ -13,24 +13,48 @@ const sum = require("./sum.js");
 // Given an empty array
 // When passed to the sum function
 // Then it should return 0
-test.todo("given an empty array, returns 0")
+test("given an empty array, returns 0", () => {
+    expect(sum([])).toBe(0);
+});
 
 // Given an array with just one number
 // When passed to the sum function
 // Then it should return that number
+test("given an array with just one number, returns that number", () => {
+    expect(sum([42])).toBe(42);
+});
 
 // Given an array containing negative numbers
 // When passed to the sum function
 // Then it should still return the correct total sum
+test("given an array containing negative numbers, returns the correct total sum", () => {
+    expect(sum([-10, -20, 30])).toBe(0);
+});
 
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
+test("given an array with decimal/float numbers, returns the correct total sum", () => {
+    expect(sum([1.5, 2.5, 3.5])).toBe(7.5);
+});
 
 // Given an array containing non-number values
 // When passed to the sum function
 // Then it should ignore the non-numerical values and return the sum of the numerical elements
+test("given an array containing non-number values, ignores non-numerical values and returns the sum of numerical elements", () => {
+    expect(sum(['hello', 10, 'world', 20])).toBe(30);
+});
 
 // Given an array with only non-number values
 // When passed to the sum function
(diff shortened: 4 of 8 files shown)


Sprint-1/fix/median.js
```diff
@@ -3,9 +3,26 @@
...
+
+  // 〰️ Create a copy of the input list to avoid modifying the original array
+  const copiedList = [...list];
+
+  // 〰️ Calculate the index of the middle element
+  const middleIndex = Math.floor(copiedList.length / 2)
+
+  // 〰️ If the list length is odd, return the middle element
+  if (copiedList.length % 2 === 1) {
+    const median = copiedList.splice(middleIndex, 1)[0];
```
cjyuan, line 15: There is a simpler and more direct syntax to access an array element at a particular position.

Sprint-1/fix/median.js
```diff
@@ -3,9 +3,26 @@
 // If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+
+  // 〰️ Create a copy of the input list to avoid modifying the original array
+  const copiedList = [...list];
+
```
cjyuan, line 9: The arrays specified in `median.test.js` happened to be sorted.

Can you try inserting a statement to sort the numbers in `copiedList` first?

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,24 @@
...
+// 〰️ Filter the array to include only numbers
+  const numericElements = elements.filter((el) => typeof el === "number");
+
+  // 〰️ Return -Infinity for an empty array (or no numeric values)
+  if (numericElements.length === 0) {
+    return -Infinity;
+  }
+
+  // 〰️ Use Math.max to find the largest number in the array
+  return Math.max(...numericElements);
```
cjyuan, line 12: #### Sprint-1/implement/max.js
What do you expect from the following function calls (on extreme cases)?

```
findMax([-Infinity])
findMax([-Infinity, 0])
findMax([NaN])
findMax([0, NaN, 1])
```
---

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,31 @@
...
+
+    let total = 0; // 〰️ Initialize the sum
+
+    for (let i = 0; i < elements.length; i ++) {
+        if (typeof elements[i] === "number") { // 〰️ Check if the element is a number
+            total += elements[i]; // 〰️ Add the number to the total
+        }
+    }
+
+    return total; // 〰️ Return the total sum
```
cjyuan, line 16: What do you expect from the following function calls (on extreme cases)?

```
sum([Infinity, 1]);
sum([Infinity, Infinity]);
sum([NaN, 1]);
sum([Infinity, -Infinity]);
```

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,48 @@ const sum = require("./sum.js");
...
 // Then it should still return the correct total sum
+test("given an array containing negative numbers, returns the correct total sum", () => {
+    expect(sum([-10, -20, 30])).toBe(0);
+});
 
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
+test("given an array with decimal/float numbers, returns the correct total sum", () => {
+    expect(sum([1.5, 2.5, 3.5])).toBe(7.5);
```

Sprint-1/stretch/aoc-2018-day1/solution.js
```diff
@@ -0,0 +1,39 @@
...
+// 〰️ The goal is to sum all the changes to find the final frequency.
+
+
+// 〰️ Step 1: Import the 'fs' module to work with the file system.
+const fs = require('fs');
+
+// 〰️ Step 2: Read the content of the input.txt file.
+// 〰️ - We use the 'utf8' encoding to read the file as a string.
+// 〰️ - The callback function handles the file content or an error if something goes wrong.
+fs.readFile('./Sprint-1/stretch/aoc-2018-day1/input.txt', 'utf8', function (error, data) {
```
cjyuan, line 13: Why specify the file path as `'./Sprint-1/stretch/aoc-2018-day1/input.txt'` instead of as `'input.txt'`?

cjyuan (commented):
Code is good. Just have some suggestions and questions for you.

AFatmaa:
Hi @cjyuan,

I just saw your comments on my PR. Thank you so much for taking the time to review it and share your thoughts. I truly appreciate your feedback. I will carefully go through each of your comments and respond to them right away.

Thanks again for your time and support! 🌸

Sprint-1/fix/median.js
AFatmaa: Yes, that's right. There is a simpler way to access an array element, and I’ve improved my code accordingly. Previously, I used splice to extract the middle element, which introduced unnecessary complexity. Instead, I removed the splice operation and replaced it with direct access using copiedList[middleIndex]. This is simpler and avoids modifying the array unnecessarily.
Thank you for pointing this out. It helped me make my code cleaner and more efficient! 😊

Sprint-1/fix/median.js
AFatmaa: You’re absolutely right. I missed that. Finding the median assumes the array is sorted, and the test cases in median.test.js were already sorted by coincidence. To handle cases where the input array isn't sorted, I’ve added a statement to sort copiedList before calculating the median.

Thank you for bringing this up! It was an important detail to address, and this change makes the function more reliable in all scenarios.

Sprint-1/implement/max.js
AFatmaa: Hi,

I added the !isNaN(el) condition to the function. Now, it works as expected for all extreme cases mentioned, correctly handling -Infinity, NaN, and arrays with mixed values.

findMax([-Infinity]): The function correctly returns -Infinity because it is the only number in the array.
findMax([-Infinity, 0]): The function returns 0, as 0 is greater than -Infinity.
findMax([NaN]): The function correctly returns -Infinity, since NaN is excluded due to the && !isNaN(el) check, ensuring that NaN values are ignored.
findMax([0, NaN, 1]): The function returns 1, as NaN is ignored, and the largest number in the array is 1.

Thank you for your feedback!

Sprint-1/implement/sum.js
AFatmaa: Hi,

I added the !isNaN(elements[i]) condition to the function. Now, it works as expected for all extreme cases mentioned.

sum([Infinity, 1]); It returns Infinity.
sum([Infinity, Infinity]); It returns Infinity.
sum([NaN, 1]); It returns 1.
sum([Infinity, -Infinity]); It returns NaN.

Thank you for your feedback!

Sprint-1/implement/sum.test.js
AFatmaa: Hi,

I updated the test to use toBeCloseTo instead of toBe. This accounts for the potential small inaccuracies in floating point arithmetic when dealing with decimal numbers. The toBeCloseTo matcher allows us to specify a tolerance (in this case, 5 decimal places), ensuring that the result is close enough to the expected value despite any minor precision errors.

I believe this is a more appropriate approach. Please let me know if there is another way.

Thank you! 😊

cjyuan: The `.toBeCloseTo()` is specific to Jest. An alternative is to use the "typical" approach: 
Consider two floating point numbers, A and B, are equal if their absolute difference is smaller than a tolerable number. For example,

`Math.abs(A - B) < 1e-10`

Sprint-1/stretch/aoc-2018-day1/solution.js
cjyuan: It depends in which folder you execute your script. It seems like you were executing your script in the "project" folder (the folder containing "Sprint-1").

`input.txt` would only work if one executes the script directly in the folder `Sprint-1/stretch/aoc-2018-day1/`.

If we can assume `input.txt` file is always in the same folder as the script, and want to be able to execute the script from any folder, we can construct the absolute path of `input.txt` in the script. ChatGPT can give a good explanation if you ask it "In JS, how to construct the absolute path of a file located in the folder where the executed script is?".

Sprint-1/implement/sum.test.js
AFatmaa: Hi,

I’ve updated my tests to use the typical approach by comparing floating-point numbers with Math.abs(A - B) < 1e-10.

Since I’m still new to JavaScript, especially when it comes to writing test files, I initially found this concept a bit tricky to understand. However, after reviewing how it works, I think I’ve applied it correctly in my updated code.

I hope my current implementation is correct, but please let me know if there’s anything else I should adjust.

Thank you so much for your guidance! 😊

Sprint-1/stretch/aoc-2018-day1/solution.js
AFatmaa: Hi!
I have to admit, I really struggled to understand this topic at first. Without ChatGPT's explanation, I don’t think I would have ever figured it out! 😅 I’m still not entirely sure if I’ve understood everything correctly, but I’ve made the changes to the code based on what I believe was the right approach. Could you please check it now and confirm if it's correct?

Here’s what I’ve done:
- I added the path module, which helps construct the absolute file path in a way that works across different operating systems (Windows, macOS, Linux).
- I replaced the relative file path ./Sprint-1/stretch/aoc-2018-day1/input.txt with the absolute path using __dirname and path.join(), ensuring that the file is always found, no matter from which directory the script is executed.

I would really appreciate it if you could review the updated code and let me know if it's correct now!

Thank you! 😊

cjyuan: `__dirname` is an environment variable that tells you the absolute path of the directory containing the currently executing file.

So the `filePath` of `input.txt` can be constructed as 
`const filePath = path.join(__dirname, 'input.txt');`

You can use `console.log()` to output the value of `filePath`.

cjyuan (commented):
Just answering your `filePath` question.

Sprint-1/stretch/aoc-2018-day1/solution.js
AFatmaa: Hi!

Thank you so much for your explanation! I realize now that I made the file pathway too long and complicated. 🙈 
I simplified the code and added a console.log(filePath); to check if the file path is correct, and it seems to work perfectly!
This makes so much more sense now! 😊

PR #111, opened 2024-12-02
West Midlands | Danial Bashirzadeh | Module-Data-Groups | Week1 | Sprint1

bashirzadehdanial:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -3,9 +3,18 @@
 // If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  list.sort();
+  const len = list.length;
+
+  if (len % 2 === 0) {
+    const mid = len / 2 - 1;
+    const middleTwo = list.slice(mid, mid + 2);
+    return (middleTwo[0] + middleTwo[1]) / 2;
+  } else {
+    const index = Math.floor(len / 2);
+    const median = list[index];
+    return median;
+  }
 }
 
 module.exports = calculateMedian;
--- a/Sprint-1/implement/max.js
+++ b/Sprint-1/implement/max.js
@@ -3,6 +3,12 @@ function findMax(elements) {
         return -Infinity;
     }else if(elements.length == 1 && typeof(elements[0])== 'number'){
       return elements[0];
+    }else if(elements.length>1){
+        const numValues = elements.filter((item) => typeof item === "number");
+        const maxNum = Math.max(...numValues);
+        console.log(maxNum);
+        
+        return maxNum
     }
 }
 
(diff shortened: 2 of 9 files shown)


Sprint-1/implement/max.js
```diff
@@ -1,4 +1,18 @@
 function findMax(elements) {
+    if(elements.length == 0){
+        return -Infinity;
+    }else if(elements.length == 1 && typeof(elements[0])== 'number'){
```
cjyuan, line 4: Why treat this case as special case? The code in the next block can also handle this case.

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,43 @@ const sum = require("./sum.js");
...
+test("Given an array containing negative numbers,  it return the correct total sum", () => {
+  expect(sum([-2, -5, -7, -44])).toEqual(-58);
+});
+
 
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
+test("Given an array with decimal/float numbers,  it return the correct total sum", () => {
+  expect(sum([2.04, 5.3, 7.5, 4.4])).toEqual(19.240000000000002);
```
cjyuan, line 39: Decimal numbers in most programming languages (including JS) are internally represented in "floating point number" format. Floating point arithmetic is not exact. For example, the result of `46.5678 - 46 === 0.5678` is false because `46.5678 - 46` only yield a value that is very close to `0.5678`.

So that is why `expect(sum([2.04, 5.3, 7.5, 4.4])).toEqual(19.24);` fails.

`expect(sum([2.04, 5.3, 7.5, 4.4])).toEqual(19.240000000000002);` may not always work because on a different implementation, the resulting sum can be different if the numbers are added in different order.
```
2.04 + 5.3 + 7.5 + 4.4 == 19.240000000000002   // true

4.4 + 5.3 + 2.04 + 7.5 == 19.240000000000002   // false
```

cjyuan (commented):
Code is good. I just left some suggestions.

Sprint-1/implement/sum.test.js
bashirzadehdanial: Yes, there are ways to handle this issue: one approach is using expect(Math.abs(result - expected) < epsilon).toBe(true) to compare with a small tolerance, and another is leveraging a library like Math.js and testing with expect(result).toBeCloseTo(expected, 10) for precision.

Sprint-1/fix/median.js
```diff
@@ -3,9 +3,18 @@
 // If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  list.sort();
```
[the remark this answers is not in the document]
bashirzadehdanial: Thank you for the feedback!

PR #114, opened 2024-12-03
London | Gislaine_DellaBella | Data Groups| Week 1

Della-Bella:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

Hi Volunteer,
thank you for your time for reviewing my code. 
Gislaine

### DIFF
--- a/Sprint-1/fix/median.test.js
+++ b/Sprint-1/fix/median.test.js
@@ -6,11 +6,6 @@
 
 const calculateMedian = require("./median.js");
 
-describe("calculateMedian", () => {
-  test("returns the median for odd length array", () => {
-    expect(calculateMedian([1, 2, 3])).toEqual(2);
-    expect(calculateMedian([1, 2, 3, 4, 5])).toEqual(3);
-  });
 
   test("returns the average of middle values for even length array", () => {
     expect(calculateMedian([1, 2, 3, 4])).toEqual(2.5);
@@ -20,7 +15,6 @@ describe("calculateMedian", () => {
   test("doesn't modify the input", () => {
     const list = [1, 2, 3];
     calculateMedian(list);
-
     expect(list).toEqual([1, 2, 3]);
   });
-});
+
--- a/Sprint-1/implement/max.js
+++ b/Sprint-1/implement/max.js
@@ -1,23 +1,56 @@
-let a= [3,5,7,9,10]
+let a = [3, 5, 9, 12];
 let result= Math.max(...a)
 console.log (result)
 
+function findMax(elements) {
+  // If the array is empty, return -Infinity right away
+  if (elements.length === 0) {
+    return -Infinity;
+  }
 
-function findMax(elements) {
-}
-function findMax(elements) {
-  let max = elements[0]; // Start with the first element as the largest
-  for (let i = 1; i < elements.length; i++) {
-    if (elements[i] > max) {
-      // If the current element is larger
-      max = elements[i]; // Update max to this new value
+  // Start with the smallest possible number as the largest so far
+  let largest = -Infinity;
+
+  // Go through each item in the array
+  for (let i = 0; i < elements.length; i++) {
+    let currentElement = elements[i]; // Get the current item
+
+    // Check if this item is a valid number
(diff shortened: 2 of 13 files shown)


Sprint-1/fix/median.test.js
```diff
@@ -6,11 +6,6 @@
 
 const calculateMedian = require("./median.js");
 
-describe("calculateMedian", () => {
-  test("returns the median for odd length array", () => {
```
cjyuan, line 10: Why deleted this test for arrays with odd length?

Sprint-1/fix/median.test.js
```diff
@@ -20,7 +15,6 @@ describe("calculateMedian", () => {
   test("doesn't modify the input", () => {
     const list = [1, 2, 3];
```
cjyuan, line 16: If you expected the array to be sorted in the function, why not start with an array that is not sorted?

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,80 @@
+let a = [3, 5, 9, 12];
```
cjyuan, line 1: It is never too early to develop the habit to keep the code clean by removing 
(or at least commenting out) unused code or debugging code.

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,32 @@
...
+
+    // Loop through each element in the array
+    for (let i = 0; i < elements.length; i++) {
+        // Check if the element is a number
+        if (typeof elements[i] === "number") {
+            total = total + elements[i]; // Add the number to total
+        }
+    }
+
+    return total; // Return the final total sum
```
cjyuan, line 24: What do you expect from the following function calls (on extreme cases)?

```
sum([Infinity, 1]);
sum([Infinity, Infinity]);
sum([NaN, 1]);
sum([Infinity, -Infinity]);
```

Sprint-1/implement/sum.test.js
```diff
@@ -2,18 +2,62 @@
...
+});
+
+// Test for an array with only non-numeric values
+test("array containing negative numbers, return the correct total sum", () => {
+  expect(sum(["hello", null, undefined, "world"])).toBe(0);
+});
+
+// Test for array with decimal/float numbers
+test("array with decimal/float numbers, return the correct total sum", () => {
+  expect(sum([3.5, 4.5, 5.5])).toBe(13.5);
```

Sprint-1/refactor/includes.js
```diff
@@ -1,13 +1,29 @@
...
       return true;
     }
   }
   return false;
-}
+}*/
+
+function includes(list, target) {
+  // Loop through each item in the list
+  for (let i = 0; i < list.length; i++) {
```
cjyuan, line 15: This implementation does not include a construct stated in the requirement (line 1).

cjyuan (commented):
Code is good. You missed something in `refactor/includes.js`.
I also left some suggestions and challenges for you.

Della-Bella:
Hi Cjyuan, 
I will have a look again at your comments. 
thank you so much for your time. 
Gislaine

Sprint-1/fix/median.test.js
Della-Bella: I don't remember why. I added it back, and all passed. ;)

Sprint-1/implement/max.js
Della-Bella: I agree but I write a lot of comments now to understand what I'm doing also, even when I chat with chat GTP, I like to understand his explanation. 

But sure, before I get this addition I will keep it in a notebook.

Sprint-1/implement/sum.test.js
Della-Bella: Function changed and passed the tests.

Sprint-1/refactor/includes.js
Della-Bella: fixed

Sprint-1/fix/median.test.js
Della-Bella: do you mean const list = [3, 2, 1];? I change the function and it passed

PR #134, opened 2024-12-08
West Midlands | Gabriel Deng | Module-Data-Groups | Week 1

gai93003:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -2,10 +2,21 @@
 // Start by running the tests for this function
 // If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
 
-function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+const calculateMedian = (list) => {
+  list.sort((a, b) => {
+    return a - b;
+  });
+
+  if (list.length % 2 === 0) {
+    const middleIndex = list.length / 2;
+    console.log((list[middleIndex - 1] + list[middleIndex]) / 2);
+    return (list[middleIndex - 1] + list[middleIndex]) / 2;
+  }
+  else {
+    return list[Math.floor(list.length / 2)];
+  }
 }
+
+calculateMedian([3,5,5,4]);
 
 module.exports = calculateMedian;
--- a/Sprint-1/implement/max.js
+++ b/Sprint-1/implement/max.js
@@ -1,4 +1,16 @@
 function findMax(elements) {
+  let maxNum = -Infinity;
+
+  if(elements.length === 0) {
+    return -Infinity;
+  }
+
+  for (element of elements) {
+    if(typeof element === 'number' && element > maxNum) {
+      maxNum = element;
+    }
+  }
+  return maxNum;
 }
 
 module.exports = findMax;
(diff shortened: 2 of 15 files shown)


Sprint-1/fix/median.js
```diff
@@ -2,10 +2,21 @@
 // Start by running the tests for this function
 // If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
 
-function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+const calculateMedian = (list) => {
+  list.sort((a, b) => {
```
cjyuan, line 6: Array is passed to a function ***by reference***  (instead of ***by value***). This implies when an array is passed to a function via parameter, the array is **shared** by the caller and the callee (the function being called).

Sprint-1/fix/median.js
```diff
@@ -2,10 +2,21 @@
...
+    return a - b;
+  });
+
+  if (list.length % 2 === 0) {
+    const middleIndex = list.length / 2;
+    console.log((list[middleIndex - 1] + list[middleIndex]) / 2);
+    return (list[middleIndex - 1] + list[middleIndex]) / 2;
+  }
+  else {
+    return list[Math.floor(list.length / 2)];
```
cjyuan, line 16: The value of `Math.floor(list.length / 2)` is suitable for both odd and even length array. So you could possibly just calculate the middle index once in your code.

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,16 @@
...
+
+  if(elements.length === 0) {
+    return -Infinity;
+  }
+
+  for (element of elements) {
+    if(typeof element === 'number' && element > maxNum) {
+      maxNum = element;
+    }
+  }
```
cjyuan, line 12: What do you expect from the following function calls (on extreme cases)?

```
findMax([-Infinity])
findMax([-Infinity, 0])
findMax([NaN])
findMax([0, NaN, 1])
```

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,12 @@
 function sum(elements) {
+  let totalSum = 0;
+
+  for (element of elements) {
+    if (typeof element === 'number') {
+      totalSum += element;
+    }
+  }
+  return totalSum;
```

```
sum([Infinity, 1]);
sum([Infinity, Infinity]);
sum([NaN, 1]);
sum([Infinity, -Infinity]);
```

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,45 @@ const sum = require("./sum.js");
...
+test ('Given an array containing negative numbers, it should still return the correct total sum', () => {
+  expect(sum([3, -5, 1, 4])).toBe(3);
+})
+
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
 
+test ('Given an array with decimal/float numbers, it should return the correct total sum', () => {
+  expect(sum([3.1, 2.5, 4.2, 0.5])).toBe(10.3)
```

Sprint-1/stretch/aoc-2018-day1/solution.js
```diff
@@ -0,0 +1,14 @@
+const frequencyFluctuation = (changes) => {
```
cjyuan, line 1: This exercise expects you to also write JS code to read the numbers from the file `input.txt` directly.

cjyuan (commented):
Code is good, just a few suggestions and challenges.

PR #141, opened 2024-12-09
SOUTH AFRICA CPT-ITP | LUKE MANYAMAZI | MODULE DATA GROUPS | WEEK 1

Luke-Manyamazi:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions


Sprint-1/fix/median.js
```diff
@@ -4,8 +4,10 @@
 
 function calculateMedian(list) {
   const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  if (list.length % 2 === 0) {
```
cjyuan, line 7: Array is passed to a function ***by reference***  (instead of ***by value***). This implies when an array is passed to a function via parameter, the array is **shared** by the caller and the callee (the function being called).

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,7 @@
 function findMax(elements) {
+  if (!Array.isArray(elements)) return -Infinity; // This handles non-array input
+  const numElements = elements.filter((item) => typeof item === "number"); // this checks if elements are numbers
+  return numElements.length > 0 ? Math.max(...numElements) : -Infinity; // math.max  returns the max number
```
cjyuan, line 4: What do you expect from the following function calls (on extreme cases)?

```
findMax([-Infinity])
findMax([-Infinity, 0])
findMax([NaN])
findMax([0, NaN, 1])
```

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,12 @@
 function sum(elements) {
+  let result = 0;
+
+  for (let i = 0; i < elements.length; i++) {
+    if (typeof elements[i] === "number") {
+      result += elements[i];
+    }
+  }
+  return result;
```

```
sum([Infinity, 1]);
sum([Infinity, Infinity]);
sum([NaN, 1]);
sum([Infinity, -Infinity]);
```

Sprint-1/implement/__tests__/sum.test.js
```diff
@@ -6,31 +6,48 @@ E.g. sum([10, 20, 30]), target output: 60
...
 // Then it should still return the correct total sum
+test("given an array containing negative numbers", () => {
+  expect(sum([-3, -7, 0, 2])).toBe(-8);
+});
 
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
+test("given an array with decimal or float numbers", () => {
+  expect(sum([6.5, 2.1, 0.5])).toBe(9.1);
```
cjyuan, line 38: Decimal numbers in most programming languages (including JS) are internally represented in "floating point number" format. Floating point arithmetic is not exact. For example, the result of `46.5678 - 46 === 0.5678` is false because `46.5678 - 46` only yield a value that is very close to `0.5678`.

cjyuan (commented):
Code is good. Just a few suggestions and challenges for you.

PR #152, opened 2024-12-11
CYF London | Sajad Soornisofla | Module-Data-Groups | Sprint 1 | WEEK 1

sajadsoorni:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -3,9 +3,13 @@
 // If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
 
 function calculateMedian(list) {
+  list.sort((a, b) => a - b);
+  console.log([list].join(""));
   const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  if (list.length % 2 === 0) {
+    return (list[middleIndex - 1] + list[middleIndex]) / 2;
+  } else {
+    return list[middleIndex];
+  }
 }
-
 module.exports = calculateMedian;
--- a/Sprint-1/implement/max.js
+++ b/Sprint-1/implement/max.js
@@ -1,4 +1,15 @@
 function findMax(elements) {
+  let max = -Infinity;
+  for (i = 0; i < elements.length; i++) {
+    if (typeof elements[i] === "number") {
+      if (elements[i] > max) {
+        max = elements[i];
+      }
+    }
+  }
+  return max;
 }
+
+// console.log(findMax(["l", "j", "o"]));
 
 module.exports = findMax;
--- a/Sprint-1/implement/max.test.js
+++ b/Sprint-1/implement/max.test.js
@@ -10,13 +10,12 @@ You should implement this function in max.js, and add tests for it in this file.
 We have set things up already so that this file can see your function from the other file.
 */
 
-const findMax = require("./max.js");
+const findMax = require("./max");
 
 // Given an empty array
 // When passed to the max function
 // Then it should return -Infinity
 // Delete this test.todo and replace it with a test.
-test.todo("given an empty array, returns -Infinity");
 
 // Given an array with one number
 // When passed to the max function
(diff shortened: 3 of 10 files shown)


Sprint-1/fix/median.js
```diff
@@ -3,9 +3,13 @@
 // If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
 
 function calculateMedian(list) {
+  list.sort((a, b) => a - b);
+  console.log([list].join(""));
```
cjyuan, line 7: Array is passed to a function ***by reference***  (instead of ***by value***). This implies when an array is passed to a function via parameter, the array is **shared** by the caller and the callee (the function being called).

Line 6: `.sort()` can modify the content in `list`, so a better practice is to make a clone of `list`, and sort the cloned array.

Line 7: It is never too early to develop the habit to keep the code clean by removing 
(or at least commenting out) unused code or debugging code.

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,15 @@
...
+    if (typeof elements[i] === "number") {
+      if (elements[i] > max) {
+        max = elements[i];
+      }
+    }
+  }
+  return max;
 }
 
+// console.log(findMax(["l", "j", "o"]));
```
cjyuan, line 13: What do you expect from the following function calls (on extreme cases)?

```
findMax([-Infinity])
findMax([-Infinity, 0])
findMax([NaN])
findMax([0, NaN, 1])
```

Sprint-1/implement/max.test.js
```diff
@@ -10,34 +10,47 @@ You should implement this function in max.js, and add tests for it in this file.
...
 // Then it should return the max and ignore non-numeric values
-
+test("Given an array with non-number values, it should return the max and ignore non-numeric values", () => {
+  expect(findMax(["hey", 10, "hi", 60, 10])).toBe(60);
+});
 // Given an array with only non-number values
 // When passed to the max function
 // Then it should return the least surprising value given how it behaves for all other inputs
+test("Given an array with only non-number values, it should return -Infinity", () => {
+  expect(findMax(["hey", "hi", "hello"])).toBe(-Infinity);
```
cjyuan, line 55: Could also consider checking other non-number values such as `true`, `null`, `undefined`.

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,11 @@
 function sum(elements) {
+  sum = 0;
+  for (let i = 0; i < elements.length; i++) {
+    if (typeof elements[i] === "number") {
+      sum = elements[i] + sum;
+    }
+  }
```

```
sum([Infinity, 1]);
sum([Infinity, Infinity]);
sum([NaN, 1]);
sum([Infinity, -Infinity]);
```

Sprint-1/implement/sum.test.js
```diff
@@ -6,31 +6,48 @@ E.g. sum([10, 20, 30]), target output: 60
...
 // Then it should still return the correct total sum
+test("Given an array containing negative numbers, it should return the correct total sum", () => {
+  expect(sum([10, -5, -2, 5])).toBe(8);
+});
 
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
+test("Given an array with decimal/float numbers, it should return the correct total sum", () => {
+  expect(sum([10.5, 20.3, 30.2])).toBe(61);
```

cjyuan (commented):
Code looks good. I just left some suggestions and challenges.

Sprint-1/fix/median.js
sajadsoorni: Hi @cjyuan ,

Thank you for the helpful feedback! I completely agree with your points.

1-Cloning the Array: You're right that modifying the original array can cause unexpected side effects. I’ll make sure to clone the array before sorting it, which will keep the original list intact.

2-Removing Debugging Code: I also appreciate your reminder about cleaning up unused code. I’ll remove the console.log statement to keep the code neat and maintain best practices.

Thanks again for your suggestions 🙏they’ll definitely improve the code.😊

Sprint-1/implement/max.js
sajadsoorni: I’ve added 
 `if (typeof elements[i] === "number" && !isNaN(elements[i]))` to properly handle NaN values and ensure the function works correctly in all cases.
 
 -The function now correctly ignores NaN values and handles edge cases like [NaN] by returning -Infinity.

Sprint-1/implement/max.test.js
sajadsoorni: I’ve added a test to check if the function returns -Infinity for non-number values like true, null, and undefined.

Sprint-1/implement/sum.js
sajadsoorni: I’ve updated the function to:

Ignore NaN values using :
 `if (typeof elements[i] === "number" && !isNaN(elements[i]))` 
to prevent them from affecting the sum.
The handling of Infinity was already in place, so this update ensures the function works properly for NaN as well.

Sprint-1/implement/sum.test.js
sajadsoorni: I believe it's better to update the function itself rather than adjusting the test. To handle floating-point precision issues, I’ve modified the return statement to round the sum to 10 decimal places using `Math.round(sum * 1e10) / 1e10`

I can update the test to use : `expect(Math.abs(sum([1.1, 1.1, 1.1]) - 3.3)).toBeLessThan(0.0001);`
for more accurate comparison.

PR #155, opened 2024-12-11
CYF London | Sajad Soornisofla | Module-Data-Groups | Quote Generator | WEEK 3

sajadsoorni:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-3/quote-generator/index.html
+++ b/Sprint-3/quote-generator/index.html
@@ -3,7 +3,7 @@
   <head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <title>Title here</title>
+    <title>Quote generator app</title>
     <script defer src="quotes.js"></script>
   </head>
   <body>
--- a/Sprint-3/quote-generator/index.html
+++ b/Sprint-3/quote-generator/index.html
@@ -13,8 +13,9 @@
       <h1>hello there</h1>
 
       <blockquote></blockquote>
-      <p id="quote"></p>
-      <p id="author"></p>
+      <p id="quote">Strive not to be a success, but rather to be of value.</p>
+      <p id="author">Albert Einstein</p>
+
       <div class="action">
         <button type="button" id="new-quote">New quote</button>
         <fieldset class="options">
--- a/Sprint-3/quote-generator/index.html
+++ b/Sprint-3/quote-generator/index.html
@@ -4,12 +4,26 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>Quote generator app</title>
+    <link rel="stylesheet" href="style.css" />
+
     <script defer src="quotes.js"></script>
   </head>
   <body>
-    <h1>hello there</h1>
-    <p id="quote"></p>
-    <p id="author"></p>
-    <button type="button" id="new-quote">New quote</button>
+    <main>
+      <h1>hello there</h1>
+
+      <blockquote></blockquote>
+      <p id="quote"></p>
+      <p id="author"></p>
+      <div class="action">
+        <button type="button" id="new-quote">New quote</button>
+        <fieldset class="options">
+          <label class="custom-checkbox">
+            <input type="checkbox" id="toggle-timer" />
+            <span id="status">Auto change</span>
+          </label>
+        </fieldset>
+      </div>
+    </main>
   </body>
 </html>
--- a/Sprint-3/quote-generator/quotes.js
+++ b/Sprint-3/quote-generator/quotes.js
@@ -491,3 +491,57 @@ const quotes = [
 ];
 
 // call pickFromArray with the quotes array to check you get a random quote
+
+function displayRandomQuote() {
+  const quoteElement = document.getElementById("quote");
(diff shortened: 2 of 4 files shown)


Sprint-3/quote-generator/quotes.js
```diff
@@ -491,3 +491,57 @@ const quotes = [
...
+document.getElementById("toggle-timer").addEventListener("change", (event) => {
+  const statusSpan = document.getElementById("status");
+
+  if (event.target.checked) {
+    statusSpan.textContent = "Auto-Play is :ON";
+
+    timer = setTimeout(() => {
+      timer = setInterval(displayRandomQuote, 60000);
+      displayRandomQuote();
+    }, 6000);
```
cjyuan, line 528: `setTimeout()` is redundant. Lines 525-528 can be replaced by:
```
      // Schedule a function to be called every 6s
      timer = setInterval(displayRandomQuote, 6000); 
```

Sprint-3/quote-generator/quotes.js
```diff
@@ -491,3 +491,57 @@ const quotes = [
...
+  } else {
+    statusSpan.textContent = "Auto Change";
+
+    clearTimeout(timer);
```
cjyuan, line 532: Can also remove this line if you rewrite the code in lines 525-528.

Sprint-3/quote-generator/quotes.test.js
```diff
@@ -41,35 +41,46 @@ afterEach(() => {
...
+
     expect(quoteP).toHaveTextContent(
       "I've learned that people will forget what you said, people will forget what you did, but people will never forget how you made them feel."
     );
     expect(authorP).toHaveTextContent("Maya Angelou");
 
+    // Click again to change quote
     userEvent.click(newQuoteBtn);
 
+    await new Promise((resolve) => setTimeout(resolve, 300));
```
cjyuan, line 82: I don't think this is a good practice because every test will take an extra 0.3s to complete.
But at the moment I don't know how to change the content first and then perform the same animation separately.
I will let you know if I find out.

Sprint-3/quote-generator/index.html
```diff
@@ -3,13 +3,28 @@
...
+
+      <blockquote></blockquote>
+      <p id="quote">Strive not to be a success, but rather to be of value.</p>
+      <p id="author">Albert Einstein</p>
+
+      <div class="action">
+        <button type="button" id="new-quote">New quote</button>
+        <fieldset class="options">
+          <label class="custom-checkbox">
+            <input type="checkbox" id="toggle-timer" />
```
cjyuan, line 23: The style applied to a checked checkbox (slight change in color) is not very noticeable.

cjyuan (commented):
Code is working, 

I made some suggestions in your code.

The first line in `quote.js` is `// DO NOT EDIT BELOW HERE`. So you probably should insert your code before that line.

Sprint-3/quote-generator/quotes.js
sajadsoorni: Thank you for reviewing my code and providing your feedback!

Regarding your comment, I wanted to clarify that I followed the guideline mentioned in line 24:
// DO NOT modify this array, otherwise the tests may break!

I did not make any changes to the code below this point, ensuring the array and other parts remain intact.

Additionally, line 493 includes a comment:

This suggested to me that it's appropriate to call pickFromArray with the quotes array. Based on this, I added my code accordingly without modifying the prohibited sections.
---------------------------------------------------------------------------------------
-The reason I placed the setInterval() call inside the setTimeout() function is to ensure that the quote does not change immediately when the checkbox is clicked. Instead, a timer starts after 6 seconds, with the next quote appearing at 60-second intervals.

This approach avoids giving the user the impression that the checkbox is behaving like a button to change the quote immediately. It maintains a consistent timing experience for the user, aligning with the intended behavior of the feature.


Please let me know if you have additional suggestions or need further clarification!

Sprint-3/quote-generator/quotes.js
sajadsoorni: Thank you for the suggestion! I’ve considered it, but I can’t remove that line because it’s essential for the current functionality. If I rewrite lines 525-528 as suggested, the specific behavior tied to clearing the timeout would no longer work as intended. Please let me know if you’d like me to elaborate further on the current approach!

Sprint-3/quote-generator/index.html
sajadsoorni: Thank you for pointing that out! I’ve updated the style for the checked checkbox to make the color change more noticeable. 
`.custom-checkbox input[type="checkbox"]:checked {
  background-color: #ec09fc;
  box-shadow: 0px 2px 4px rgba(0, 0, 0, 0.4);
}`

Sprint-3/quote-generator/quotes.test.js
sajadsoorni: I understand your concern about the extra delay. I agree that it would be better to avoid unnecessary waiting time in the tests. I’ll keep an eye out for a solution that allows the content to change without the need for this delay. Let me know if you find a better approach, and I’ll be happy to make adjustments on my end as well 🙏

PR #165, opened 2024-12-13
CYF London | Phone Naing | Module-Data-Groups | Week 1

Mgphone:
## Learners, PR Template

Self checklist

- [✅] I have committed my files one by one, on purpose, and for a reason
- [✅] I have titled my PR with COHORT_NAME | FIRST_NAME LAST_NAME | REPO_NAME | WEEK 
- [✅ ] I have tested my changes
- [✅] My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/contributing/)
- [✅] My changes meet the [requirements](./README.md) of this task

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -3,9 +3,15 @@
 // If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
 
 function calculateMedian(list) {
+  let isEven = list.length % 2 == 0;
   const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  if (isEven) {
+    const median = (list[middleIndex] + list[middleIndex - 1]) / 2;
+    return median;
+  } else {
+    const median = list[middleIndex];
+    return median;
+  }
 }
 
 module.exports = calculateMedian;
--- a/Sprint-1/implement/max.js
+++ b/Sprint-1/implement/max.js
@@ -1,4 +1,14 @@
 function findMax(elements) {
+  const filterNumberArray = [];
+  for (let i = 0; i < elements.length; i++) {
+    if (typeof elements[i] == "number") {
+      filterNumberArray.push(elements[i]);
+    }
+  }
+  const maxValue = Math.max(...filterNumberArray);
+  return maxValue;
 }
+// console.log(findMax([1, 2, 3, 4, "hello"]));
+// console.log(findMax([]));
 
 module.exports = findMax;
--- a/Sprint-1/implement/sum.test.js
+++ b/Sprint-1/implement/sum.test.js
@@ -1,3 +1,4 @@
+const sum = require("./sum.js");
 /* Sum the numbers in an array
 
 In this kata, you will need to implement a function that sums the numerical elements of an array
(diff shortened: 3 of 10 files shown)


Sprint-1/fix/median.js
```diff
@@ -3,9 +3,15 @@
 // If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
 
 function calculateMedian(list) {
+  let isEven = list.length % 2 == 0;
```
cjyuan, line 6: 1) The arrays specified in `median.test.js` happened to be sorted.

2) Array is passed to a function ***by reference***  (instead of ***by value***). This implies when an array is passed to a function via parameter, the array is **shared** by the caller and the callee (the function being called).

---
Can you try inserting code to first clone the array `list` and then sort the numbers in cloned array?
You may have to rename the array in lines 6-12.

Sprint-1/implement/sum.test.js
```diff
@@ -1,36 +1,58 @@
...
 // Then it should still return the correct total sum
+test("negative number and return correct total sum", () => {
+  expect(sum([1, 2, 3, 4, -10])).toBe(0);
+});
 
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
+test("add decimal float and return correct total", () => {
+  expect(sum([1.1, 2.2, 3.3, 1, 2])).toBe(9.6);
```
cjyuan, line 41: Decimal numbers in most programming languages (including JS) are internally represented in "floating point number" format. Floating point arithmetic is not exact. For example, the result of `46.5678 - 46 === 0.5678` is false because `46.5678 - 46` only yield a value that is very close to `0.5678`.

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,14 @@
...
+  for (let i = 0; i < elements.length; i++) {
+    if (typeof elements[i] == "number") {
+      filterNumberArray.push(elements[i]);
+    }
+  }
+  const maxValue = Math.max(...filterNumberArray);
+  return maxValue;
 }
+// console.log(findMax([1, 2, 3, 4, "hello"]));
+// console.log(findMax([]));
```
cjyuan, line 12: What do you expect from the following function calls (on extreme cases)?

```
findMax([-Infinity])
findMax([-Infinity, 0])
findMax([NaN])
findMax([0, NaN, 1])
```

cjyuan (commented):
Code looks good. I only have few suggestions and one challenge question.

Sprint-1/fix/median.js
Mgphone: sure i can do it !!

Sprint-1/implement/sum.test.js
Mgphone: wow!! Thanks for your comments
I use factor making 10**10 and add in my reduce second parameter with factor and return result/factor

Mgphone: > The change to the code looks good, but I don't understand what you mean by `I use factor making 10**10 and add in my reduce second parameter with factor and return result/factor` 10**10 in JS means "10 to the power 10", which is equal to 10000000000.

yeah that is truth :) i change it to the whole number of decimal to avoid binary error

PR #172, opened 2024-12-14
Wets Midlands | Samira Hekmati |Module Data Groups-alarm clock | Sprint-3

samirahekmati:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions


cjyuan (commented):
Code works fine if the user only clicks the "Set Alarm" button once. However, if the user enters a time and then clicks the "Set Alarm" button multiple times, the countdown clock will not display properly. Can you fix the problem?

It would be better if the count down clock can start showing without any delay after the user clicks the "Set Alarm" button; currently there is a one second delay.

samirahekmati:
thank you @cjyuan for your comment. in order to solve this problem, I added a feature to the `pauseAlarm` function as well as `setAlarm` function. Based on my new code, the `set alarm` button will be disabled as soon as the user clicks the button. therefore, they wont be able to click it again until they click `stop alarm` button. At this point `set alarm` button is enabled again.

samirahekmati:
Also, I fixed the one second delay of the countdown.

cjyuan:
@samirahekmati
>  the set alarm button will be disabled as soon as the user clicks the button.
While this approach can prevent multiple concurrent active intervals from being created, it may not be friendly to users who want to reset the alarm while the clock is counting down.

Can you think of another approach to prevent multiple concurrent active intervals from being created?

PR #173, opened 2024-12-14
west Midlands|Azar Alamapanah|Module-Data-Groups|Sprint-3(reading list)

azaralampanah2:
## Learners, PR Template

Self checklist

- [yes ] I have committed my files one by one, on purpose, and for a reason
- [ yes] My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/contributing/)

## Changelist

Briefly explain your PR.
This PR has been raised regarding my work on Sprint-3/reading list
## Questions

### DIFF
--- a/Sprint-3/reading-list/script.js
+++ b/Sprint-3/reading-list/script.js
@@ -20,47 +20,33 @@ const books = [
     bookCoverImage: "https://blackwells.co.uk/jacket/l/9780135957059.jpg",
   },
 ];
-const firstBook = document.getElementById("firstBookDiv");
-const secondBook = document.getElementById("secondBookDiv");
-const thirdBook = document.getElementById("thirdBookDiv");
 
-firstBook.textContent = books[0].title + " by " + books[0].author;
-const firstBookImage = document.createElement("img");
-firstBookImage.src = books[0].bookCoverImage;
-firstBookImage.alt = "The design of Everyday Things Book Image";
-firstBook.appendChild(firstBookImage);
-firstBookImage.width = 50;
-firstBookImage.height = 50;
+for (i = 0; i < books.length; i++) {
+    let container = document.getElementById("reading-list")
 
-secondBook.textContent = books[1].title + " by " + books[1].author;
-const secondBookImage = document.createElement("img");
-secondBookImage.src = books[1].bookCoverImage;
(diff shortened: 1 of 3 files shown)


Sprint-3/reading-list/style.css
```diff
@@ -157,3 +157,21 @@ body {
     max-height: 80px;
   }
 }
+
+#firstBookDiv {
```
eliza-an, line 161: this wont work as its called bookDiv not firstBookDiv

Sprint-3/reading-list/script.js
```diff
@@ -21,3 +21,32 @@ const books = [
...
+    text.innerHTML = title
+    text.style.fontSize = 18
+
+    image.src = books[i].bookCoverImage
+    image.width = 125
+    image.alt = books[i].title
+
+    listItem.style.marginBottom = 15
+
+    books[i].alreadyRead ? listItem.style.backgroundColor = "green" : listItem.style.backgroundColor = "red"
```
eliza-an, line 41: this is quite an advanced way to write it! Well done!

Sprint-3/reading-list/script.js
```diff
@@ -20,4 +20,47 @@ const books = [
     bookCoverImage: "https://blackwells.co.uk/jacket/l/9780135957059.jpg",
   },
 ];
+const firstBook = document.getElementById("firstBookDiv");
+const secondBook = document.getElementById("secondBookDiv");
```
eliza-an: this is good, but the only issue here is that this would work only with 3 books. What if new books were added in the future? you would have to add a lot of extra code. Instead, you can just create a list item for every book!

eliza-an (changes requested):
Some really advanced code in here, but it looks great. 
However, you only have 3 commits, make sure you are committing as much a possible so that you follow best practices :)

Sprint-3/reading-list/style.css
azaralampanah2: thank you so much for reviewing my codes.
I changed firstBookDiv to bookDiv, but got something unexpected so left it as before.

Sprint-3/reading-list/script.js
azaralampanah2: thank you so much

PR #174, opened 2024-12-14
West Midlands|Azar Alampanah|Module-Data-Groups|Week2

azaralampanah2:
## Learners, PR Template

Self checklist

- [ yes] I have committed my files one by one, on purpose, and for a reason
- [yes ] I have tested my changes
- [yes ] My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/contributing/)

## Changelist

Briefly explain your PR.
This PR is about my work on Module-Data-Groups/Sprint-2
## Questions

### DIFF
--- a/Sprint-2/implement/querystring.js
+++ b/Sprint-2/implement/querystring.js
@@ -1,31 +1,20 @@
-
-
 function parseQueryString(queryString) {
   const queryParams = {};
   if (queryString.length === 0) {
     return queryParams;
   }
-  queryString = decodeURIComponent(queryString);
 
   const keyValuePairs = queryString.split("&");
-  console.log(keyValuePairs);
+
   for (const pair of keyValuePairs) {
     const index1 = pair.indexOf("=");
-    console.log(index1);
-    const key = pair.slice(0, index1);
-    const value = pair.slice(index1 + 1);
-    console.log(value);
-    console.log(key);
+    const key = decodeURIComponent(pair.slice(0, index1));
+    const value = decodeURIComponent(pair.slice(index1 + 1));
     queryParams[key] = value;
   }
 
   return queryParams;
 }
 
-console.log(parseQueryString("a%25b=c%26d"));
+console.log(parseQueryString("a%25b=c%26d")); 
 module.exports = parseQueryString;
-
-
-
-
-
--- a/Sprint-2/implement/querystring.test.js
+++ b/Sprint-2/implement/querystring.test.js
@@ -34,8 +34,7 @@ test("parse query string with no key",()=>{
 })
 test("parse query string with encoded charecters",()=>{
   expect(parseQueryString("a%25b=c%26d")).toStrictEqual({
-    "a%b": "c",
-    "": "d"
+    "a%b": "c&d"
   }
   )
 })


Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,14 @@
-function contains() {}
+function contains(mainObject, char) {
+  for (const key of Object.keys(mainObject)) {
+    if (key.includes(char)) {
+      return true;
+    } else return false;
+  }
+  if (Object.keys(mainObject).length == 0 || typeof mainObject != "object")
```
cjyuan, line 7: - It is better to check and reject invalid values at the start of the function block. This way, we can assure the remaining code in the function does not have to deal with invalid values.

Sprint-2/implement/contains.test.js
```diff
@@ -7,9 +7,16 @@ particular property
 E.g. contains({a: 1, b: 2}, 'a') // returns true
 as the object contains a key of 'a'
 
+
 E.g. contains({a: 1, b: 2}, 'c') // returns false
 as the object doesn't contains a key of 'c'
 */
+test("object contains the property", () => {
+  expect(contains({ a: 1, b: 2 }, "a")).toBe(true);
```
cjyuan, line 15: Try checking also 
`expect(contains({ a: 1, b: 2 }, "b")).toBe(true);`
to discover bug in your code.

Sprint-2/implement/contains.test.js
```diff
@@ -33,3 +45,11 @@ test.todo("contains on empty object returns false");
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+test("input is not an object", () => {
+  expect(contains(["a", "q", 5], "a")).toBe(false);
+  expect(contains("hello", "e")).toBe(false);
+  expect(contains(["a", "q", 5], 5)).toBe(false);
```
cjyuan, line 51: Try checking also 
`expect(contains(["a", "q", 2], 2)).toBe(false);`

Sprint-2/implement/lookup.js
```diff
@@ -1,5 +1,23 @@
...
+function createLookup(currencyCodes) {
+  const createdObject = {};
+  let createdArray = [];
+  for (let i = 0; i < currencyCodes.length; i++) {
+    createdArray = currencyCodes[i];
+    console.log(createdArray);
+    const splitArray = createdArray.map((element) => [element]); //splits an array to arrayes of its elements
+
+    const key = splitArray[0].join(); //converts an array to a string
+    const value = splitArray[1].join(); ////converts an array to a string
```
cjyuan, line 10: What is `createdArray`?

Code in lines 6-10 can greatly simplified.

Sprint-2/implement/lookup.test.js
```diff
@@ -33,3 +33,25 @@ It should return:
    'CA': 'CAD'
  }
 */
+test("should return the corresponding currency code", () => {
```
cjyuan, line 36: The function does not actually return a "currency code".

Sprint-2/implement/querystring.js
```diff
@@ -4,13 +4,18 @@ function parseQueryString(queryString) {
...
-
+  console.log(keyValuePairs);
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
+    const index1 = pair.indexOf("=");
+    console.log(index1);
+    const key = pair.slice(0, index1);
+    const value = pair.slice(index1 + 1);
+    console.log(value);
+    console.log(key);
```
cjyuan, line 14: For each of the following function calls, does your function return the value you expected?
```
parseQueryString("a=b&=&c=d")
parseQueryString("=b")
parseQueryString("a=b&&c=d")
parseQueryString("a%25b=c%26d`)
```

Sprint-2/implement/tally.test.js
```diff
@@ -32,3 +45,7 @@ test.todo("tally on an empty array returns an empty object");
 // Given an invalid input like a string
 // When passed to tally
 // Then it should throw an error
+test("return error, as the input is not an array", () => {
+  expect(tally(4353636)).toBe("Error! Input must be an array");
+  expect(tally("Hello ")).toBe("Error! Input must be an array");
+});
```
cjyuan, line 51: Can you also test
`expect(tally({ a: 1, b: 2 })).toBe("Error! Input must be an array");`

Sprint-2/interpret/invert.js
```diff
@@ -10,20 +10,29 @@ function invert(obj) {
...
+// it shold return {key:1}, as "key" is a variable here
 
 // b) What is the current return value when invert is called with { a: 1, b: 2 }
-
+// it will return:{key:2},
 // c) What is the target return value when invert is called with {a : 1, b: 2}
+//{1:"a" , 2:"b"}
 
 // c) What does Object.entries return? Why is it needed in this program?
+// it converts each object's properties to indivisual arrays
```
cjyuan, line 30: What items can be found in each of these "individual arrays"?

cjyuan (commented):
Besides a few bugs, code is generally quite good.

It is never too early to develop the habit to keep the code clean by removing (or at least commenting out) unused code or debugging code.

Sprint-2/implement/contains.js
azaralampanah2: ok, thank you CJ. I swapped it and put "return false" at the top.

Sprint-2/implement/contains.test.js
azaralampanah2: thanks for this comment, I found the bug and amended the function

Sprint-2/implement/contains.test.js
azaralampanah2: with amending my function from last comment, this test is working as well.
Thanks

Sprint-2/implement/lookup.js
azaralampanah2: Thanks CJ, I simplified my codes. Also it seems "createdArray" is useless, so I removed it.

Sprint-2/implement/lookup.test.js
azaralampanah2: Thanks for your comment, I amended it.

Sprint-2/interpret/invert.js
azaralampanah2: it is better to say that it provides a convenient way to access both keys and values of an object in a single operation, making it essential for efficiently inverting the object in this program.

Sprint-2/implement/tally.test.js
azaralampanah2: Thanks CJ for your comment. I amended my code

Sprint-2/implement/querystring.js
azaralampanah2: I wrote some tests in querystring.test.js to test the function with these values, and it worked good for all except the last one parseQueryString("a%25b=c%26d`). I searched and saw that with decodeURIComponent(queryString), I can decode URL encoded characters.

Sprint-2/implement/querystring.test.js
```diff
@@ -25,3 +25,17 @@ test("parses querystring with empty value", () => {
...
+test("parse query string with no key",()=>{
+  expect(parseQueryString("=b")).toStrictEqual({"":"b"})
+})
+test("parse query string with encoded charecters",()=>{
+  expect(parseQueryString("a%25b=c%26d")).toStrictEqual({
+    "a%b": "c",
+    "": "d"
+  }
+  )
+})
```
cjyuan, line 41: This should be 
```
  expect(parseQueryString("a%25b=c%26d")).toStrictEqual({
    "a%b": "c=d"
  });
```

Encoding is applied to individuals keys and values are not to the whole querystring. So we need to decode individual keys and values too.

Also, can you try checking the return value of `parseQueryString("keyonly")`

cjyuan (commented):
I added a review for the updated `querystring.js` and `querystring.test.js`.

Sprint-2/implement/querystring.test.js
azaralampanah2: Hi CJ, Thank you for your quick review.
I found that the equivalent character for %26 is "&". so if it is "&" it should be new key/value in our object{....&d}., so output should be :{.... , "":"d"}. But I think you mean %26 is "=". Am I right?

cjyuan: You are right about %26 being "&". So the expected return value should be:
```
  expect(parseQueryString("a%25b=c%26d")).toStrictEqual({
    "a%b": "c&d"
  });
```

azaralampanah2: I amended my coded to decode depends on each key/value not the whole string.
Thank you so much .

PR #192, opened 2024-12-16
West Midlands | Samira Hekmati | Module Data Groups | Sprint-3 Quote Generator app

samirahekmati:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-3/quote-generator/index.html
+++ b/Sprint-3/quote-generator/index.html
@@ -3,7 +3,7 @@
   <head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <title>Title here</title>
+    <title>Quote generator app</title>
     <script defer src="quotes.js"></script>
   </head>
   <body>
--- a/Sprint-3/quote-generator/index.html
+++ b/Sprint-3/quote-generator/index.html
@@ -4,12 +4,16 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>Quote generator app</title>
+    <link rel="stylesheet" href="style.css">
     <script defer src="quotes.js"></script>
   </head>
   <body>
-    <h1>hello there</h1>
-    <p id="quote"></p>
-    <p id="author"></p>
+    <div class="quote">
+      <p id="quote"></p>
+      <p id="author"></p>
     <button type="button" id="new-quote">New quote</button>
+
+    </div>
+    
   </body>
 </html>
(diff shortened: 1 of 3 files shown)


Sprint-3/quote-generator/quotes.js
```diff
@@ -491,3 +491,19 @@ const quotes = [
 ];
 
 // call pickFromArray with the quotes array to check you get a random quote
+
+function generateNewQuote() {
+  const newQuote = pickFromArray(quotes);
+  const quoteText = newQuote.quote;
+  const quoteAuthor = newQuote.author;
+
+  document.getElementById("quote").innerText = `"${quoteText}"`;
```
eliza-an, line 500: Good work. Typically you would set this to a variable so that it is reusable. Y0ou could also do something like this: 
  quote.textContent = output.quote;
  author.textContent = "-" + output.author;

Sprint-3/quote-generator/quotes.js
```diff
@@ -491,3 +491,19 @@ const quotes = [
...
+  document.getElementById("author").innerText = `- ${quoteAuthor}`;
+}
+function setup() {
+  document.getElementById("new-quote").addEventListener("click", () => {
+    generateNewQuote();
+  });
+}
+
+window.onload = setup;
```
eliza-an, line 509: When i open this on my computer this doesn't work. I think this is because you do not call the function setup, you only call it check there is a click. When you load a page the quote should display!
But yours only displays when the button is clicked :) 
image

Sprint-3/quote-generator/index.html
```diff
@@ -3,13 +3,17 @@
...
+    <title>Quote generator app</title>
+    <link rel="stylesheet" href="style.css">
     <script defer src="quotes.js"></script>
   </head>
   <body>
-    <h1>hello there</h1>
-    <p id="quote"></p>
-    <p id="author"></p>
+    <div class="quote">
+      <p id="quote"></p>
```
eliza-an, line 12: I can't open this on my machine , but from what i can see you dont have the big quotes at the start. That's a part of it too! Have a look at the sample image again

eliza-an (changes requested):
Good work overall, just some minor tweaks

Sprint-3/quote-generator/quotes.js
samirahekmati: hi @eliza-an thank you so much for all your comments, specially this one. I thought this line of code: ` window.onload = setup; ` means that call the function setup. can you please elaborate more on how to do that?

eliza-an: You are correct. Window.onload will load the setup function when the window is loaded, but look at what the setup function does. 5e only thing the setup function does, is add an event listener , when the button is clicked you will run generateNewQuote. So it’s not the setup function that should run for window.onload, it’s the generateNewQuote function. Does that make sense? Because when the window loads you want the new quote generated, not the event listener added

PR #200, opened 2024-12-16
NORTHWEST|RAHWA ZESLUS|MODULE-DATA-GROUPWEEK1

RahwaZeslusHaile:
:

NORTHWEST|RAHWA ZESLUS|MODULE-DATA-GROUPWEEK1

## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -3,9 +3,20 @@
 // If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
 
 function calculateMedian(list) {
+  
+  //here this will work for odd length only.so I have to add a condition for even too
   const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  const firstMiddle = list[middleIndex - 1];
+  const secondMiddle = list[middleIndex];
+  if(list.length % 2=== 0){
+    const median = (firstMiddle + secondMiddle)/2;
+    return median;   
+  }
+  else{
+       const median = list[middleIndex];
+       return median;
+
+  } 
 }
 
 module.exports = calculateMedian;
--- a/Sprint-1/implement/dedupe.test.js
+++ b/Sprint-1/implement/dedupe.test.js
@@ -1,4 +1,5 @@
 const dedupe = require("./dedupe.js");
+
 /*
 Dedupe Array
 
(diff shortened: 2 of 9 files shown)


cjyuan:
This PR has no label. Does it need a review?

Sprint-1/implement/dedupe.test.js
```diff
@@ -7,16 +8,49 @@ Dedupe Array
...
+    });
+    
+    test("returns the an unrepeated character of an array element with given array", () => {
+        expect(dedupe([5, 1, 1, 2, 3, 2, 5, 8])).toEqual([5, 1, 2, 3, 8]);
+    });
+    test("returns the original array element with given array", () => {
+        expect(dedupe([1,2,3,4])).toEqual([1,2,3,4]);
+    });
+    test("returns the an unrepeated character of an array element with a string and number array", () => {
+        expect(dedupe([1,'g',7,2,'r','g',3,4,'q'])).toEqual([1,'g',7 ,2 ,'r',3,4,'q']);
```
cjyuan, line 33: #### Todo
Can you try using ChatGPT to make the description of these tests more accurate?

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,18 @@
+//I will do for loop to iterate every element
+//I have to use that filters number only the array to number
+//check every index is whether greater than the next index
 function findMax(elements) {
+    let smallestNumber = -Infinity;
+    for(let index=0; index < elements.length; index++){
+        const currentNumber = elements[index];
+        if (!isNaN(currentNumber) && currentNumber > smallestNumber){
```
cjyuan, line 8: Because of the way JS compares values of different types, some non-number-type values can pass this test.
Here are some examples,

```
findMax(["300", 2])
findMax(["300", "4"])
findMax([[100], 4, [200]])
```

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,11 @@
 function sum(elements) {
-}
+        
+        let numbersOnly  = elements.filter(value => typeof value === 'number')
+        let sumOfNumbers = numbersOnly.reduce((acc,curr)=>acc+curr,0);
+         return sumOfNumbers;
+    
+    }
```

```
sum([NaN, 1]);
sum([Infinity, -Infinity]);
```

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,40 @@ const sum = require("./sum.js");
...
-
+test("returns  total sum with an array containing negative numbers given", () => {
+    expect(sum([-2, -6, -8])).toEqual(-16);
+});
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
-
+test("returns  total sum with an array decimal/float numbers given", () => {
+    expect(sum([1.2, 0.6, 0.005])).toEqual(1.8049999999999997);
```
cjyuan, line 38: Decimal numbers in most programming languages (including JS) are internally represented in "floating point number" format. Floating point arithmetic is not exact. For example, the result of `46.5678 - 46 === 0.5678` is false because `46.5678 - 46` only yield a value that is very close to `0.5678`. Even changing the order in which the program add/subtract numbers can yield different values.

So the following could happen
```
```

package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "devDependencies": {
+    "jest": "^29.7.0"
+  },
+  "dependencies": {
+    "find-up": "^7.0.0"
```
cjyuan, line 6: What do you use `find-up` for in your project? I don't think you need it for `Sprint-1/stretch/aoc-2018-day1/solution.js`.

cjyuan (commented):
Code is mostly correct. I left some comments and suggestions.

One thing I would like you to try is, with the help of ChatGPT, create more accurate and more concise description for each unit test.

Sprint-1/fix/median.js
```diff
@@ -3,9 +3,20 @@
 // If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
 
 function calculateMedian(list) {
+
```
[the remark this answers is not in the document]
RahwaZeslusHaile: Your review is insightful and helpful for me to look deeper into the code! Sorting the array is indeed essential when calculating the median, as it ensures that the middle value(s) are correctly identified.

Sprint-1/fix/median.js
```diff
@@ -3,9 +3,20 @@
 // If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
 
 function calculateMedian(list) {
+  
+  //here this will work for odd length only.so I have to add a condition for even too
   const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  const firstMiddle = list[middleIndex - 1];
+  const secondMiddle = list[middleIndex];
```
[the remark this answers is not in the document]
RahwaZeslusHaile: Exactly by declaring firstMiddle and secondMiddle inside the if block, their scope is limited to where they are actually used. This makes the code cleaner and more focused.

Sprint-1/implement/max.js
RahwaZeslusHaile: - I didn’t notice this issue earlier, and thank you for making me see it! JavaScript can sometimes compare values of different types in unexpected ways, especially when performing comparisons between strings and numbers

RahwaZeslusHaile:
@cjyuan Thank you for your detailed reviews and insights! 🙏 I’ve carefully considered your feedback and made the updates:

Your review helped me dive deeper into understanding the nuances of floating-point arithmetic and how to write more robust test cases. I appreciate the guidance—it’s been incredibly helpful in refining my code! 😊

PR #219, opened 2024-12-19
West-Midlands-Community | Matthew Law | To Do List | WEEK 9

matthewlawpixel:
Self checklist

## Changelist

This PR is for the exercise To Do List.

## Questions

### DIFF
--- a/Sprint-3/todo-list/script.js
+++ b/Sprint-3/todo-list/script.js
@@ -1,9 +1,11 @@
+let todoIdCounter = 1;
+
 function populateTodoList(todos) {
   let list = document.getElementById("todo-list");
   // Write your code to create todo list elements with completed and delete buttons here, all todos should display inside the "todo-list" element.
   list.innerHTML = "";
 
-  todos.forEach((todo) => {
+  todos.forEach((todo, index) => {
     let li = document.createElement("li");
     li.className = "list-group-item d-flex justify-content-between align-items-center";
     
@@ -17,6 +19,10 @@ function populateTodoList(todos) {
     }
     li.textContent = textContent;
 
+    if (todo.completed) {
+      li.style.textDecoration = "line-through";
+    }
+
     let badge = document.createElement("span");
     badge.className = "badge bg-primary rounded-pill";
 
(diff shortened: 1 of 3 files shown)


Sprint-3/todo-list/script.js
```diff
@@ -1,13 +1,51 @@
...
+    li.textContent = textContent;
+
+    let badge = document.createElement("span");
+    badge.className = "badge bg-primary rounded-pill";
+
+    let checkIcon = document.createElement("i");
+    checkIcon.className = "fa fa-check";
+    checkIcon.style.cursor = "pointer";
+    checkIcon.addEventListener("click", () => {
+      li.style.textDecoration = li.style.textDecoration === "line-through" ? "" : "line-through";
```
cjyuan, line 27: The `todos` array represent the state of the app, whereas the HTML + CSS represent the "view" of the app.
It is better to organise code in the following manner:
- Any change to the app (add new todo, delete completed todos, mark a todo as completed) should only change the states. 
- Then update the view based only on states.

This way, If the app were to save the todos in a database or in `window.localStorage`, the app can just save the state data (without saving any HTML or CSS).

So the code inside this `checkIcon`'s event listener should only update the `completed` property of the current todo, then update the view.
The code to conditionally apply a line-through should be placed somewhere around line 18.

Similar approach applies also to `trashIcon`'s event listener.

Sprint-3/todo-list/script.js
```diff
@@ -16,10 +54,32 @@ populateTodoList(todos);
...
   event.preventDefault();
-  // Write your code here... and remember to reset the input field to be blank after creating a todo!
+  let todoInput = document.getElementById("todo-input");
+  let task = todoInput.value;
+  let todoDeadline = document.getElementById("todo-deadline");
+  let deadline = todoDeadline.value;
+
+
+  if (task.trim() !== "") {
+    todos.push({ task: task, completed: false, deadline: deadline || null });
```
cjyuan, line 64: Could also consider sanitise the value of `deadline` before adding it to `todos` (in order to always keep the application's states valid).

Sprint-3/todo-list/script.js
```diff
@@ -16,10 +54,32 @@ populateTodoList(todos);
...
+let addTodoButton = document.getElementById("add-todo");
+addTodoButton.addEventListener("click", addNewTodo);
+
+// Advanced challenge: Write a function that checks the todos in the todo list and deletes the completed ones (we can check which ones are completed by seeing if they have the line-through styling applied or not).
 function deleteAllCompletedTodos() {
-  // Write your code here...
+  todos = todos.filter((todo) => {
+    let listItems = Array.from(document.getElementById("todo-list").children);
+    let matchedItem = listItems.find((li) => li.textContent.includes(todo.task));
+    return matchedItem && matchedItem.style.textDecoration !== "line-through";
```
cjyuan, line 79: What if multiple todo's task have the same text? For example,
```
let todos = [
  { task: "Wash the dishes", completed: false, deadline: "2024-12-31" },
  { task: "Do the shopping", completed: false, deadline: "2024-12-25" },
  { task: "Wash the dishes", completed: true, deadline: "2024-12-29" },
];
```

If you focus on updating the application state (the array `todos`), the filtering task would be much simpler.

cjyuan (commented):
I admire your determination to take on the "extra advanced challenge"!

Can you try also restructuring your code based on my suggestion? That is, by separating application states from application's view (a typical approach used in implementing GUI).

Sprint-3/todo-list/script.js
matthewlawpixel: Thank you @cjyuan i have added some sanitisation to deadline as you have suggested

PR #230, opened 2024-12-21
CYF-ITP-South Africa | Simphiwe Mabuya | Module-Data-Groups | Sprint-2 

Simphiwe-Mabuya:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -10,6 +10,12 @@ as the object contains a key of 'a'
 E.g. contains({a: 1, b: 2}, 'c') // returns false
 as the object doesn't contains a key of 'c'
 */
+describe("function contains", () => {
+    test("When passed an object and a property name, it should return true if the object contains the property ", () => {
+        expect(contains({ name: "Mike", age: 25, city: "Cape Town" }, "age")).toEqual(true);
+    });
+})
+
 
 // Acceptance criteria:
 
@@ -20,7 +26,6 @@ as the object doesn't contains a key of 'c'
 // Given an empty object
 // When passed to contains
 // Then it should return false
-test.todo("contains on empty object returns false");
 
 // Given an object with properties
 // When passed to contains with an existing property name
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -26,6 +26,10 @@ describe("function contains", () => {
     test("When passed an invalid parameters like an array, it should return false ", () => {
         expect(contains([true, "Mike", 35, "City"], "Mike")).toEqual(false);
     });
+
+    test("When passed an invalid parameter like an array and we check for indices , it should return true ", () => {
+        expect(contains([true, "Mike", 35, "City"], "1")).toEqual(true);
+    });
 })
 
 
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -28,7 +28,7 @@ describe("function contains", () => {
     });
 
     test("When passed an invalid parameter like an array and we check for indices , it should return true ", () => {
-        expect(contains([true, "Mike", 35, "City"], "1")).toEqual(true);
+        expect(contains([true, "Mike", 35, "City"], "1")).toEqual(false);
     });
 })
 
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -27,7 +27,7 @@ describe("function contains", () => {
         expect(contains([true, "Mike", 35, "City"], "Mike")).toEqual(false);
     });
 
-    test("When passed an invalid parameter like an array and we check for indices , it should return true ", () => {
+    test("When passed an invalid parameter like an array and we check for indices , it should return false ", () => {
         expect(contains([true, "Mike", 35, "City"], "1")).toEqual(false);
     });
 })
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -12,7 +12,19 @@ as the object doesn't contains a key of 'c'
 */
 describe("function contains", () => {
     test("When passed an object and a property name, it should return true if the object contains the property ", () => {
-        expect(contains({ name: "Mike", age: 25, city: "Cape Town" }, "age")).toEqual(true);
+        expect(contains({ name: "Mike", age: 25, city: "Cape Town"}, "age")).toEqual(true);
+    });
+
+    test("When passed an empty object, it should return false ", () => {
+        expect(contains({}, "age")).toEqual(false);
+    });
+
+    test("When passed an object with a non-existent property name, it should return false ", () => {
+        expect(contains({name: "Mike", age: 25, city: "Cape Town" }, "band")).toEqual(false);
+    });
+
+    test("When passed an invalid parameters like an array, it should return false ", () => {
+        expect(contains([true, "Mike", 35, "City"], "Mike")).toEqual(false);
     });
 })
 
--- a/Sprint-2/implement/querystring.js
+++ b/Sprint-2/implement/querystring.js
@@ -16,4 +16,4 @@ function parseQueryString(queryString) {
 
 module.exports = parseQueryString;
 
-console.log(parseQueryString("a=b=c=d=e"));
+console.log(parseQueryString("a=b&=&c=d"));
--- a/Sprint-2/implement/querystring.js
+++ b/Sprint-2/implement/querystring.js
@@ -14,3 +14,5 @@ function parseQueryString(queryString) {
 }
 
 module.exports = parseQueryString;
+
+console.log(parseQueryString("equation=x=y+1"));
--- a/Sprint-2/implement/querystring.js
+++ b/Sprint-2/implement/querystring.js
@@ -6,7 +6,8 @@ function parseQueryString(queryString) {
   const keyValuePairs = queryString.split("&");
 
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
+    const [key, ...rest] = pair.split("=");
+    const value = rest.join("=")
     queryParams[key] = value;
   }
 
@@ -15,4 +16,4 @@ function parseQueryString(queryString) {
 
 module.exports = parseQueryString;
 
-console.log(parseQueryString("equation=x=y+1"));
+console.log(parseQueryString("a=b=c=d=e"));
--- a/Sprint-2/implement/querystring.test.js
+++ b/Sprint-2/implement/querystring.test.js
@@ -7,6 +7,6 @@ const parseQueryString = require("./querystring.js")
 
 test("parses querystring values containing =", () => {
   expect(parseQueryString("equation=x=y+1")).toEqual({
-    "equation": "x=y+1",
+    equation: 'x'
   });
 });
--- a/Sprint-2/implement/querystring.test.js
+++ b/Sprint-2/implement/querystring.test.js
@@ -7,7 +7,7 @@ const parseQueryString = require("./querystring.js")
 
 describe("parseQueryString()", () => {
   test("parses querystring values containing multiple =", () => {
-    expect(parseQueryString("equation=x=y+1")).toEqual({equation: 'x'});
+    expect(parseQueryString("equation=x=y+1")).toEqual({ equation: 'x=y+1' });
   });
 
   test("given an empty string, it will return an empty object", () => {
@@ -27,7 +27,11 @@ describe("parseQueryString()", () => {
   });
 
   test("given an invalid query string, it will return a key-value pair if found or undefined for a value", () => {
-    expect(parseQueryString("equation=x=y+1&true-6+new")).toEqual({ equation: 'x', 'true-6+new': undefined });
+    expect(parseQueryString("equation=x=y+1&true-6+new")).toEqual({ equation: 'x=y+1', 'true-6+new': '' });
+  });
+
+  test("given a query string with a single property and a single =, it will return a key and an empty string '' for a value", () => {
+    expect(parseQueryString("a=")).toEqual({ a: '' });
   });
 
 });
--- a/Sprint-2/implement/querystring.test.js
+++ b/Sprint-2/implement/querystring.test.js
@@ -5,8 +5,11 @@
 
 const parseQueryString = require("./querystring.js")
 
-test("parses querystring values containing =", () => {
-  expect(parseQueryString("equation=x=y+1")).toEqual({
-    equation: 'x'
+describe("parseQueryString()", () => {
+  test("parses querystring values containing =", () => {
+    expect(parseQueryString("equation=x=y+1")).toEqual({
+      equation: 'x'
+    });
   });
 });
+
--- a/Sprint-2/implement/querystring.test.js
+++ b/Sprint-2/implement/querystring.test.js
@@ -6,10 +6,29 @@
 const parseQueryString = require("./querystring.js")
 
 describe("parseQueryString()", () => {
-  test("parses querystring values containing =", () => {
-    expect(parseQueryString("equation=x=y+1")).toEqual({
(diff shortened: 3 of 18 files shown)


Sprint-2/implement/querystring.js
```diff
@@ -14,3 +14,5 @@ function parseQueryString(queryString) {
 }
 
 module.exports = parseQueryString;
+
+console.log(parseQueryString("equation=x=y+1"));
```
cjyuan, line 18: The given function is not implemented correctly. You are expected to fix it.

Here are some tests which you can try to see if the function returns the expected value:
```
parseQueryString("a=b=c=d=e")
parseQueryString("a=b&=&c=d")
parseQueryString("a=")
parseQueryString("=b")
parseQueryString("a=b&&c=d")
parseQueryString("a%25b=c%26d`)
```

Sprint-2/implement/querystring.test.js
```diff
@@ -5,8 +5,30 @@
 
 const parseQueryString = require("./querystring.js")
 
-test("parses querystring values containing =", () => {
-  expect(parseQueryString("equation=x=y+1")).toEqual({
-    "equation": "x=y+1",
+describe("parseQueryString()", () => {
+  test("parses querystring values containing multiple =", () => {
+    expect(parseQueryString("equation=x=y+1")).toEqual({equation: 'x'});
```
cjyuan, line 10: The expected returned value for this test case should be:
`expect(parseQueryString("equation=x=y+1")).toEqual({equation: 'x=y+1'});`

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,15 @@
-function tally() {}
+function tally(items) {
+    const frequencies = {};
+
+    if (typeof items === "string") {
+        throw new Error("Invalid input");
+    }
```
cjyuan, line 6: If `items` is not a string, it could still be other types of value (in addition to array). How should we check if `items` is an array in JavaScript?

Sprint-2/interpret/invert.js
```diff
@@ -10,20 +10,37 @@ function invert(obj) {
...
 
+// The target return value is {'1': 'a', '2': 'b'}
+
 // c) What does Object.entries return? Why is it needed in this program?
 
+// This method returns an array of [key, value] pairs from the object
+
 // d) Explain why the current return value is different from the target output
 
+//It's because we couldn't pass / swap key with invertObj using dot notation appending key instead of value
```
cjyuan, line 44: I do not quite understand what this statement means. Can you try rewriting it? Perhaps with the help of ChatGPT?

Sprint-2/interpret/invert.test.js
```diff
@@ -0,0 +1,28 @@
...
+    test("Given an object with a single key-value pair, it should swap the key and value in the object", () => {
+        expect(invert({x: 10})).toEqual({"10": "x"});
+    });
+
+    test("Given an empty object, it should return an empty object", () => {
+        expect(invert({})).toEqual({});
+    });
+
+    test("When we pass a string into invert, it should throw an error", () => {
+        expect(() => invert('I enjoy learning JavaScript')).toThrow(`Duplicate value found. Cannot invert.`);
```
cjyuan, line 18: What do you expect from following function calls?
```
invert("ABCD")
invert([1,2,3,4])
invert(1234)
```

Sprint-2/stretch/count-words.js
```diff
@@ -26,3 +26,32 @@
...
+function countWords(words){
+
+  words = words.split(/\s+/) // Splitting the input string into words, split by whitespace
+
+  const cleanWords = words.map(word => word.replace(/[.,!?]/g, "").toLowerCase()); 
+
+  const results = new Map();
+
+  for(let word of cleanWords){
+    if(typeof word !== "string" || word === "") continue;
```
cjyuan, line 39: At line 39, any chance the value of `word`  is not a string?

Sprint-2/stretch/till.js
```diff
@@ -8,24 +8,44 @@ function totalTill(till) {
   let total = 0;
 
   for (const [coin, quantity] of Object.entries(till)) {
-    total += coin * quantity;
+
+    let coinValue;
+
+    if(coin.endsWith("p")){
+      coinValue = parseInt(coin.replace('p', ""), 10);
```
cjyuan, line 15: `parseInt(coin)` would also work. May I suggest checking out how `parseInt()` convert a string to an integer?

Sprint-2/stretch/till.js
```diff
@@ -8,24 +8,44 @@ function totalTill(till) {
...
+
+    } else if (coin.startsWith("£")){
+      throw new Error("Invalid coin format");
+    }
```
cjyuan, line 19: Among all possible invalid coin formats (e.g., "$1.00", "Foobar") , why specifically check for this particular invalid coin format?

cjyuan (commented):
Code is mostly good.
Hope my comments and suggestions can lead you to better understand object and arrays in JavaScript.

It would be nice if you can rebase the current branch onto `main` so that the current branch won't include modified files in Sprint-1 folder.

Simphiwe-Mabuya:
Morning CJ
Seasons Greetings!

I am sure this will find you well and in great spirit.
Thanks a million for your constructive feedback as always.

Let me get on it and get back to you soonest.

Best regards
Simphiwe Mabuya

Simphiwe-Mabuya:
Greetings CJ ..I am sure this will find you in great spirit as always.

1. Rebasing the current branch: by current branch do you mean
feature/Sprint-3? ..This is just to double check before I rebase...I
don't want to rebase the wrong branch.

2. In Sprint-2/debug/author.js:
I managed to find the approach / way to obtain an array of property
values from an object, making it possible to use for-of loop to
iterate through the property values. We use Object.keys() to get an
array of the object's keys, then iterate over them with for...of.
However, I opted to use a for in loop and then renamed the loop
variable to key according to convention.

3. In Sprint-2/implement/contains.js
After reading up on the 'in' operator and hasOwnProperty, I eventually
opted for the hasOwnProperty approach for it safer for cases where we
want to exclude properties inherited from Object.prototype or other
prototypes.

I also tested

contains([true, "Mike", 35, "City"], "1") : and it returned true
because arrays in JavaScript are objects, and their indices (e.g.,
"0", "1", etc.) are treated as properties.
contains(null, "Mike") : The function throws an error when I pass null
as the first argument because null is not an object

4. In Sprint-2/implement/querystring.js:

Fixed my queryString function so it can handle different cases correctly.
After fixing my function the return value for ("equation=x=y+1") is
now correct : expect(parseQueryString("equation=x=y+1")).toEqual({equation:
'x=y+1'});

5. In Sprint-2/implement/tally.js:

Fixed my function to also check if items is an array or not.

6. In Sprint-2/interpret/invert.js:

Updated / fixed my answer to question d) : The line invertedObj.key =
value; is actually setting the property literally named "key" on
invertedObj, rather than using the value of key as the property name.
Instead, it should be invertedObj[key] = value; to dynamically set the
property of invertedObj using the value of key.

What do you expect from the following function calls?

invert("ABCD") : expected return value : { A: '0', B: '1', C: '2', D: '3' }
invert([1,2,3,4]): expected return value : { '1': '0', '2': '1', '3':
'2', '4': '3' }
invert(1234): expected return value : {} an empty object

7. In Sprint-2/stretch/count-words.js:

Yes, there is a chance that word might not be a string (it could be
any other type, such as a number, object, array, or boolean), and the
condition typeof word !== "string" will handle those cases by skipping
the current iteration of the loop when word is not a string or is an
empty string.

8. In Sprint-2/stretch/till.js:

The parseInt() function in JavaScript converts a string into an
integer by parsing its characters sequentially, from left to right,
until it encounters a character that is not part of a valid number.

Also...a coin-counting app might expect simple numeric inputs, so
"Foobar" or "1.00.00" are outright invalid.

I trust that my answers and corrections will meet your expectations.

Looking forward to your response or thoughts regarding my solutions.
Thanks a million yet again CJ for such constructive and eye opening feedback.

Best regards
Simphiwe Mabuya

cjyuan:
It would be easier for a reviewer to track responses if you can use the "Reply..." to reply a comment.

image

cjyuan:
The modified files from Sprint-1 still exists in this branch. If you have rebased Sprint-2 branch on to `main`, you need to "force push" your local repo to Github.

Simphiwe-Mabuya:
Greetings CJ ..

Thanks a lot for your response, I will certainly look into force pushing my
local repo as per your suggestion.

As soon as that’s done I’ll let you know.

Thanks a million again. Enjoy your day.

Best regards
Simphiwe Mabuya

Sprint-2/implement/contains.test.js
```diff
@@ -10,6 +10,24 @@ as the object contains a key of 'a'
...
+    test("When passed an empty object, it should return false ", () => {
+        expect(contains({}, "age")).toEqual(false);
+    });
+
+    test("When passed an object with a non-existent property name, it should return false ", () => {
+        expect(contains({name: "Mike", age: 25, city: "Cape Town" }, "band")).toEqual(false);
+    });
+
+    test("When passed an invalid parameters like an array, it should return false ", () => {
+        expect(contains([true, "Mike", 35, "City"], "Mike")).toEqual(false);
```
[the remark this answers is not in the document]
cjyuan: How do you make the function to return false when the first parameter is an array or `null`?
Please note that both `typeof null` and `typeof []` evaluate to `"object"`.

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,15 @@
-function tally() {}
+function tally(items) {
+    const frequencies = {};
+
+    if (typeof items === "string" || !Array.isArray(items)) {
```
cjyuan, line 4: Is it necessary to check if `items` is a string when you are also checking if `items` is not an array?

Sprint-2/interpret/invert.test.js
```diff
@@ -0,0 +1,36 @@
...
```
cjyuan, line 18: This test is inaccurate. 
For the string value you have chosen, the function throws the exception not because the parameter is a string.

You may want to check what `Object.entries()` returns when it is passed a string as a parameter.

How do you ensure the function will properly **work only if the parameter is a valid object**?

Sprint-2/interpret/invert.test.js
```diff
@@ -0,0 +1,36 @@
...
+        expect(() => invert({a: 1, b: 1})).toThrow(`Duplicate value found. Cannot invert.`);
+    });
+
+    test("Given an edge case with a different data type, it should swap the keys and values of the object ", () => {
+        expect(invert({ a: 1, b: '2', c: true })).toEqual({ '1': 'a', '2': 'b', true: 'c' });
+    });
+
+    test("Given an edge case with (ABCD), it should return an object with keys and their indices/positions for values ", () => {
+        expect(invert("ABCD")).toEqual({ A: '0', B: '1', C: '2', D: '3' });
+    });
```
cjyuan, line 31: Why the function recognises this string and not the other strings you specified in the tests that throw exception?

Sprint-2/interpret/invert.test.js
```diff
@@ -0,0 +1,36 @@
...
+
+    test("Given an edge case of an array with only numbers, it should return an object with numbers for keys and their indices/positions for values ", () => {
+        expect(invert([1, 2, 3, 4])).toEqual({ '1': '0', '2': '1', '3': '2', '4': '3' });
+    });
```
cjyuan, line 35: This case is similar to how your function treats string; it is inaccurate.

`invert([1, 1, 1, 1])` would have thrown an exception with your current implementation.

You should also check what `Object.entries()` returns when it is passed different arrays.

How do you ensure the function will properly **work only if the parameter is a valid object**?

Sprint-2/interpret/invert.test.js
```diff
@@ -0,0 +1,36 @@
...
+        expect(() => invert('I enjoy learning JavaScript')).toThrow(`Duplicate value found. Cannot invert.`);
+    });
+
+    test("When we pass duplicate values, it should throw an error", () => {
+        expect(() => invert({a: 1, b: 1})).toThrow(`Duplicate value found. Cannot invert.`);
+    });
+
+    test("Given an edge case with a different data type, it should swap the keys and values of the object ", () => {
+        expect(invert({ a: 1, b: '2', c: true })).toEqual({ '1': 'a', '2': 'b', true: 'c' });
+    });
```
cjyuan, line 27: May I suggest looking up how JS converts a value to an object property name when the value is used as  `object[value]`?

Sprint-2/stretch/count-words.js
cjyuan: You answered this:
> Yes, there is a chance that word might not be a string (it could be
any other type, such as a number, object, array, or boolean), and the
condition typeof word !== "string" will handle those cases by skipping
the current iteration of the loop when word is not a string or is an
empty string.

What are the values in **`cleanWords`**?
Can you try remove `typeof word !== "string ||` from line 39, and see if you can find a case to make the function behave differently? Or show the function code to ChatGPT and ask it the same question?

Sprint-2/stretch/till.js
cjyuan: You don't have to check for possible invalid values in this problem. But if you wanted to consider invalid cases, can you think of an approach that can treat **any value that is not one of "1p", "5p", "20p", "50p"** as invalid (instead of just the strings that begin with "£"?

cjyuan (commented):
I left you a few more comments. Hope they can help you understand objects in JS better.

Simphiwe-Mabuya:
Greetings Mr CJ

I am sure that this will find you well and in great spirit as always.

In Sprint-2/implement/tally.js
<https://github.com/CodeYourFuture/Module-Data-Groups/pull/230#discussion_r1909685279>:

It is not actually necessary to check if items is a string because the
Array.isArray() method will return false for all non array values,
including strings. So I refactored my code and removed the checking if
items is a string or not.

In Sprint-2/interpret/invert.test.js
<https://github.com/CodeYourFuture/Module-Data-Groups/pull/230#discussion_r1909690316>
:

Fixed my code first so it can work properly if the parameter is a valid
object by checking if obj passed is an object, is null or an array and then
updated the test error message to correspond with the functions
expectations.

Modifying my code to handle only plain objects meant even "ABCD" which is a
string ..will be invalid and the function will throw an error.

Similar to how the function treats strings ..invert([1, 1, 1, 1]) now
throws an error because an array is not a plain object.

I also checked what Object.entries() returns : it converts the string to an
object and returns an array of key-value pairs (entrie) where each
character becomes a key and its index a value.

 - May I suggest looking up how JS converts a value to an object property
name when the value is used as object[value]?

  JavaScript follows specific steps to convert the value into a string (or
symbol) because object property names must either be strings or symbols.

In Sprint-2/implement/contains.test.js
<https://github.com/CodeYourFuture/Module-Data-Groups/pull/230#discussion_r1909680502>
:

To ensure the function returns false when the first parameter is an array
or null I had to add another logic that checks if obj is null or an array,
then modified my return statement also.

In Sprint-2/stretch/count-words.js
<https://github.com/CodeYourFuture/Module-Data-Groups/pull/230#discussion_r1909709422>
:

The values in "cleanWords"  are strings or  cleaned and split elements in
an array ["apple", "orange", "apple", "orange", "apple"]. And when I
removed typeof word !== "string ||the function behaviour did not change.

In Sprint-2/stretch/till.js
<https://github.com/CodeYourFuture/Module-Data-Groups/pull/230#discussion_r1909717452>
:

Due to your question, I  decided to take a different approach, and modified
my code,  so that it can treat *any value that is not one of "1p", "5p",
"20p", "50p"* as invalid.

Thanks again as always Mr CJ for your constructive feedback ..much
appreciated. Please do enjoy your evening.

Please do let me know if there's nothing else so I can mark this PR as
complete.

Best regards

Simphiwe Mabuya

PR #234, opened 2024-12-22
London | Pooriya Ketabi | Module-Data-Group | Sprint-1

PooriyaKTB:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-1/implement/sum.js
+++ b/Sprint-1/implement/sum.js
@@ -1,4 +1,11 @@
 function sum(elements) {
+    let sum = 0;
+    for (let i = 0; i < elements.length; i += 1) {
+        if(typeof elements[i] == "number" && !isNaN(elements[i])) {
+            sum += elements[i];
+        }
+    }
+    return Math.round(sum * 1e10) / 1e10;
 }
 
 module.exports = sum;
--- a/Sprint-1/implement/sum.test.js
+++ b/Sprint-1/implement/sum.test.js
@@ -13,24 +13,52 @@ const sum = require("./sum.js");
 // Given an empty array
 // When passed to the sum function
 // Then it should return 0
-test.todo("given an empty array, returns 0")
 
+test("An empty array, should return 0", () => {
+    expect(sum([])).toBe(0);
+});
+
+// ---------------------------------------------------------------------------------------------
 // Given an array with just one number
 // When passed to the sum function
 // Then it should return that number
 
+test("An arrat with just one number, should return that number", () => {
+    expect(sum([20])).toBe(20);
+});
+
+// ---------------------------------------------------------------------------------------------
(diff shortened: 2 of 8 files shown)


Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,11 @@
 function sum(elements) {
+    let sum = 0;
+    for (let i = 0; i < elements.length; i += 1) {
+        if(typeof elements[i] == "number" && !isNaN(elements[i])) {
+            sum += elements[i];
+        }
+    }
+    return Math.round(sum * 1e10) / 1e10;
```
cjyuan, line 8: Why perform this operation? It can make the result less accurate. 
What if we call the function as `sum([1e300])`?

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,52 @@ const sum = require("./sum.js");
...
+});
+
+// ---------------------------------------------------------------------------------------------
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
 
+test("An array with decimal or float numbers, should return correct sum", () => {
+    expect(sum([20.3, 15.6, 71.8, 54.7])).toBe(162.4);
+});
```
cjyuan, line 46: Decimal numbers in most programming languages (including JS) are internally represented in "floating point number" format. Floating point arithmetic is not exact. For example, the result of `46.5678 - 46 === 0.5678` is false because `46.5678 - 46` only yield a value that is very close to `0.5678`. Even changing the order in which the program add/subtract numbers can yield different values.

So the following could happen
```
```

and with your current implementation,  the following test would fail    
```
  test("An array with decimal or float numbers, should return correct sum", () => {
    expect(sum([1.2, 0.6, 0.005])).toBe(1.2 + 0.6 + 0.005);
});
```

cjyuan (commented):
Code is good.

You may want to lookup info about "how to check equality of floating point numbers" in general, and also in Jest (Jest has a built-in method for dealing with floating point numbers).

Sprint-1/implement/sum.js
PooriyaKTB: Thank you for mentioning that.
This operation is used to round the result to 10 decimal places.
For large inputs like `1e300`, it will result in `Infinity` due to JavaScript's numerical limits.

Sprint-1/implement/sum.test.js
PooriyaKTB: Thank you for your detailed explanation,  you’re absolutely right.  
Interestingly, though, when I ran the tests, I didn’t encounter any errors.
Screenshot (35)

PR #236, opened 2024-12-23
West Midlands | Gabriel Deng | Module-Data-Groups | Week 3 - Todo List

gai93003:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-3/todo-list/script.js
+++ b/Sprint-3/todo-list/script.js
@@ -1,6 +1,54 @@
+const form = document.querySelector('form');
+const addBtn = document.getElementById('addBtn');
+const removeBtn  = document.getElementById('remove-all-completed');
+let list = document.getElementById('todo-list');
+let inputEl = document.getElementById('input');
+
 function populateTodoList(todos) {
-  let list = document.getElementById("todo-list");
   // Write your code to create todo list elements with completed and delete buttons here, all todos should display inside the "todo-list" element.
+
+  list.innerHTML = ''
+  
+  todos.forEach((todo, index) => {
+    let listItem = document.createElement('li');
+    listItem.className = "list-item";
+    listItem.textContent = todo.task;
+
+    if(todo.completed) {
+      listItem.classList.add('completed');
+    }
+
+    let span = document.createElement('span');
+    span.setAttribute('id', 'btnContainer');
+
+    let markBtnComplete = document.createElement('i');
+    markBtnComplete.className = 'fa fa-check';
+    markBtnComplete.setAttribute('aria-hidden', 'true');
+    span.appendChild(markBtnComplete);
+
+    let deleteBtn = document.createElement('i');
+    deleteBtn.className = 'fa fa-trash';
+    deleteBtn.setAttribute('aria-hidden', 'true');
+    span.appendChild(deleteBtn);
+    
+
+    // Event listeners;
+    markBtnComplete.addEventListener('click', () => {
+      todos[index].completed = !todos[index].completed;
+      populateTodoList(todos);
+      saveData();
+    });
+    deleteBtn.addEventListener('click', () =>{
+      todos.splice(index, 1);
+      populateTodoList(todos);
+      saveData();
+    });
+
+    // Append elements to the parent elements
+    listItem.appendChild(span);
+    list.appendChild(listItem);
+  });
+
 }
 
 // These are the same todos that currently display in the HTML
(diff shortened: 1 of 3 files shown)


illicitonion (commented):
This pull request contains all of the sprint 2 exercises you did as well - can you use `git rebase -i` to get rid of those commits, so that this PR only contains the six commits starting at 9c68654ea967db48782ff3755f135ed5434f5af3?

After you do, please re-add the Needs Review label :)

gai93003:
Thanks Daniel, I have rebased the commits

illicitonion:
I still see all 13 commits here - you may need to `git push --force-with-lease` to remove the commits from the version on GitHub.

(Note that `git push --force-with-lease` is kind of scary to do - it is one of the few things in Git that _destroys_ information - it throws away stuff on GitHub that it used to have - make sure you're sure you're doing the right thing before you run this command!)

gai93003:
I have made the changes and i believe that there are only 7 commits now remaining which are only for the todo-list

Sprint-3/todo-list/script.js
```diff
@@ -1,6 +1,54 @@
...
+    let listItem = document.createElement('li');
+    listItem.className = "list-item";
+    listItem.textContent = todo.task;
+
+    if(todo.completed) {
+      listItem.classList.add('completed');
+    }
+
+    let span = document.createElement('span');
+    span.setAttribute('id', 'btnContainer');
```
illicitonion, line 22: There should only be one element in a page per ID, but this looks like it uses the same ID for each element.

What are you using the ID of this element for? Is there some other thing you could use, instead if ID, to achieve the same effect?

Sprint-3/todo-list/script.js
```diff
@@ -1,6 +1,54 @@
...
+let inputEl = document.getElementById('input');
+
 function populateTodoList(todos) {
-  let list = document.getElementById("todo-list");
   // Write your code to create todo list elements with completed and delete buttons here, all todos should display inside the "todo-list" element.
+
+  list.innerHTML = ''
+  
+  todos.forEach((todo, index) => {
+    let listItem = document.createElement('li');
```
illicitonion, line 13: All of your variables here are declared with `let`, but in general we prefer `const` to `let` where possible.

Is there a reason these need to be declared with `let` instead of `const`? If so, what's the reason? If not, why do we prefer `const` to `let`?

Sprint-3/todo-list/script.js
```diff
@@ -1,6 +1,54 @@
...
+    listItem.textContent = todo.task;
+
+    if(todo.completed) {
+      listItem.classList.add('completed');
+    }
+
+    let span = document.createElement('span');
+    span.setAttribute('id', 'btnContainer');
+
+    let markBtnComplete = document.createElement('i');
```
illicitonion, line 24: It's quite unusual to use an `i` element a  button. What's an `i` element? Why have you opted to use it here? Can you think of a more appropriate element to use?

Sprint-3/todo-list/script.js
```diff
@@ -1,6 +1,54 @@
...
+    markBtnComplete.className = 'fa fa-check';
+    markBtnComplete.setAttribute('aria-hidden', 'true');
```
illicitonion, line 26: What does the `aria-hidden` attribute do? Why is it useful here?

illicitonion (commented):
The shape of this solution is looking good, and the app works well, but I have a few questions mostly around the DOM elements you're generating :)

illicitonion:
> I have made the changes and i believe that there are only 7 commits now remaining which are only for the todo-list

Thanks for fixing up the commits!

Sprint-3/todo-list/script.js
gai93003: This element is used to design the text in italic style. I used this element because it was included in the test file which we were asked not to modify. The most appropriate element to use is a button element

Sprint-3/todo-list/script.js
gai93003: This attribute is used to visually hide elements and its contents.

illicitonion: Aah, I just checked the README for this exercise - I see that it told you to add this attribute specifically. I don't think it should, it's a problem with the exercise, but you've done what was asked of you here :)

Sprint-3/todo-list/script.js
gai93003: I think in this case, there is no need for the use of let since none of the variables i declared were reassigned. Const is used so that i cannot mistakenly reassign the variable to a different value.

Sprint-3/todo-list/script.js
gai93003: The ID attribute is used to uniquely identify an element. In this case the class attribute is the most appropriate attribute to use.

illicitonion:
Hi @gai93003 - I'm removing the Complete label again - you've acknowledged the comments I've left with words, but you haven't fixed them in the code. Please fix them in the code, then re-add the "Needs Review" label and I'll take another look :)

gai93003:
Hey Daniel, I think i have made all the changes which you made reviews on.

illicitonion:
I don't see any pushed changes? Did you forget to push?

I still see e.g. that you create multiple elements in the same page with the same ID...

gai93003:
I made the changes on github because my laptop got spoiled and couldn't start up.

PR #244, opened 2024-12-24
London 11 | Pezhman-Azizi | Data-Groups | sprint 1

Pezhman-Azizi:
## Learners, PR Template

Self checklist

## Changelist

This PR has everything I’ve worked on for Sprint 
1. Here's what’s included:
Fix Section: I fixed the functions that weren’t passing the tests. Ran the tests, figured out what was wrong, and sorted it all out.

Implement Section: I created the max, sum, and dedupe functions based on the requirements. Tested them thoroughly to make sure they work for different inputs.

Refactor Section: Cleaned up and improved the given code so it’s easier to read and still works perfectly.

### DIFF
--- a/Sprint-1/implement/sum.test.js
+++ b/Sprint-1/implement/sum.test.js
@@ -13,7 +13,37 @@ const sum = require("./sum.js");
 // Given an empty array
 // When passed to the sum function
 // Then it should return 0
-test.todo("given an empty array, returns 0")
+
+describe('sum function', () => {
+  test('should return the sum of all numeric elements in the array', () => {
+    expect(sum([1, 2, 3, 4])).toBe(10);
+  });
+
+  test('should ignore non-numeric elements and return the sum of numeric values', () => {
+    expect(sum([1, 2, 'a', 3, 'b'])).toBe(6);
+  });
+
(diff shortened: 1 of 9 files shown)


Sprint-1/implement/sum.test.js
```diff
@@ -13,7 +13,37 @@ const sum = require("./sum.js");
...
+  test('should handle arrays with a single numeric element', () => {
+    expect(sum([42])).toBe(42);
+  });
+
+  test('should handle arrays with negative numbers', () => {
+    expect(sum([1, -1, 2, -2, 3])).toBe(3);
+  });
+
+  test('should handle arrays with decimal numbers', () => {
+    expect(sum([1.5, 2.5, 3])).toBe(7);
```
cjyuan, line 43: Decimal numbers in most programming languages (including JS) are internally represented in "floating point number" format. Floating point arithmetic is not exact. For example, the result of `46.5678 - 46 === 0.5678` is false because `46.5678 - 46` only yield a value that is very close to `0.5678`. Even changing the order in which the program add/subtract numbers can yield different values.

So the following could happen
```

```

cjyuan (commented):
Code is good. I only have one suggestion.

PR #247, opened 2024-12-25
WM4 | FATIMA_SAFANA | Module-Data-Groups | WEEK7

Fatimasfn:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-1/implement/max.js
+++ b/Sprint-1/implement/max.js
@@ -17,7 +17,7 @@ function findMax(elements) {
             }
         });
 
-// If no numerical elements are found, return -Infinity
+// If no numerical elements are found, return "no numbers exist in the array"
             if (numericalElement.length === 0) {
                 return  "no numbers exist in the array";
             }
--- a/Sprint-1/implement/max.js
+++ b/Sprint-1/implement/max.js
@@ -1,4 +1,13 @@
 function findMax(elements) {
+//given an empty array, returns -Infinity
+    if (elements.length === 0) {
+        return -Infinity
+//given an array with one number, returns that number
+    } else if (elements.length === 1 && typeof elements[0] === "number")
+        return elements[0];
 }
+
+const elements = [1];
+console.log(findMax(elements));
 
 module.exports = findMax;
--- a/Sprint-1/implement/max.js
+++ b/Sprint-1/implement/max.js
@@ -1,13 +1,36 @@
 function findMax(elements) {
-//given an empty array, returns -Infinity
+// Given an empty array, return -Infinity
     if (elements.length === 0) {
-        return -Infinity
-//given an array with one number, returns that number
-    } else if (elements.length === 1 && typeof elements[0] === "number")
+        return -Infinity;
+
+ // Given an array with one number, return that number
+    } else if (elements.length === 1 && typeof elements[0] === "number") {
         return elements[0];
+
+// Check if the array includes both numerical and non-numerical values
+    } else {
+// Collect only numerical values
+        let numericalElement = [];
+        elements.forEach((item) => {
+            if (typeof item === "number") {
+                numericalElement.push(item);
+            }
+        });
+
+// If no numerical elements are found, return -Infinity
+            if (numericalElement.length === 0) {
+                return  "no numbers exist in the array";
+            }
+// Debugging output to check filtered values
+        console.log(numericalElement); 
+
+// Return the maximum numerical value
+        return Math.max(...numericalElement);
+    }
 }
 
-const elements = [1];
-console.log(findMax(elements));
+
+const elements = [ "dog", "cat"];
+console.log(findMax(elements)); // Output: 4
 
 module.exports = findMax;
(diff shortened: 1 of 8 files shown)


halilibrahimcelik (commented):
Well done, You successfully implement all test cases but there is a challenging issue that we have to keep in mind in testing median, that's decimal numbers. 
lets say you want to find  and test  median of gpa list, then you have to approach your logic differently.  current result of this array is a floating number. So you may need to upgrade your logic in that matter.
`const gpaList = [3.5, 4.0, 3.2, 3.8, 3.9, 4.0, 3.7, 3.8, 3.9, 3.9];`

Fatimasfn:
@halilibrahimcelik  thank you for your feedback, I will look into it.

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,36 @@
...
+        let numericalElement = [];
+        elements.forEach((item) => {
+            if (typeof item === "number") {
+                numericalElement.push(item);
+            }
+        });
+
+// If no numerical elements are found, return "no numbers exist in the array"
+            if (numericalElement.length === 0) {
+                return  "no numbers exist in the array";
```
cjyuan, line 22: If a function can return different types of value (in your case, a number or a string), then the caller will have to check the type of the returned value before using the returned value. 
Can you think of a more appropriate value to return when the given array contains no numbers?

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,36 @@
 function findMax(elements) {
+// Given an empty array, return -Infinity
+    if (elements.length === 0) {
+        return -Infinity;
+
+ // Given an array with one number, return that number
+    } else if (elements.length === 1 && typeof elements[0] === "number") {
+        return elements[0];
+
```
cjyuan, line 9: Why not just use the remaining code to deal with array with one number? Why write extra code to deal with an array containing one number?

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,36 @@
...
+            }
+// Debugging output to check filtered values
+        console.log(numericalElement); 
+
+// Return the maximum numerical value
+        return Math.max(...numericalElement);
+    }
 }
 
+
```
cjyuan, line 32: What do you expect from the following function calls (on extreme cases)?

```
findMax([NaN])
findMax([0, NaN, 1])
```

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,39 @@
+
 function sum(elements) {
+// Given an empty array, return 0
+    if (elements.length === 0) {
+        return 0;
+
+ // Given an array with one number, return that number
+    } else if (elements.length === 1 && typeof elements[0] === "number") {
+        return elements[0];
+
```
cjyuan, line 10: Why not just use the remaining code to deal with arrays with one number? Why write extra code to deal with an array containing one number?

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,39 @@
...
+        elements.forEach((item) => {
+            if (typeof item === "number") {
+                numericalElement.push(item);
+            }
+        });
+
+// If no numerical elements are found,"no numerical value exists"
+            if (numericalElement.length === 0) {
+                return  "no numerical value exists";
+            }
```
cjyuan, line 24: Instead of return a string, can you think of a more appropriate value to return when the given array contains no numbers?

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,39 @@
...
+// Debugging output to check filtered values
+        console.log(numericalElement); 
+
+// Return the sum of all the numbers in the array
+            let cumulative = 0
+            numericalElement.forEach((item) => {
+                cumulative += item
+            });
+        return cumulative
+    }
```
cjyuan, line 34: What do you expect from the following function calls (on extreme cases)?

```
sum([NaN, 1]);
```

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,44 @@ const sum = require("./sum.js");
...
+    expect(sum([-1, -2, -3])).toEqual(-6);
+});
+
 
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
+test("given an array containing negative numbers, returns the correct sum", () => {
+    expect(sum([1.5, 1.5, 1.5])).toEqual(4.5);
+});
```
cjyuan, line 41: Decimal numbers in most programming languages (including JS) are internally represented in "floating point number" format. Floating point arithmetic is not exact. For example, the result of `46.5678 - 46 === 0.5678` is false because `46.5678 - 46` only yield a value that is very close to `0.5678`. Even changing the order in which the program add/subtract numbers can yield different values.

So the following could happen
```

```

cjyuan (commented):
Code is good and tests are quite comprehensive. I left you some comments and suggestions.

Fatimasfn:
Thank you for reviewing @cjyuan really appreciate it .

Sprint-1/implement/sum.js
Fatimasfn: I will expect it to return 1,

PR #249, opened 2024-12-25
London 11 | Pezhman-Azizi | Data-Groups | sprint 2

Pezhman-Azizi:
## Learners, PR Template

Self checklist

## Changelist

This pull request includes my work for Sprint 2. It focuses on debugging, implementing functions based on given requirements, and interpreting code.

For debugging, I identified and fixed issues in the provided files. For the implementation tasks, I created and tested functions to meet the requirements. Lastly, I interpreted slightly larger programs by referring to documentation and using tools like console.log to understand the code better.

Let me know if there's anything you'd like me to refine or improve!

### DIFF
--- a/Sprint-2/debug/author.js
+++ b/Sprint-2/debug/author.js
@@ -3,6 +3,20 @@
 // This program attempts to log out all the property values in the object.
 // But it isn't working. Explain why first and then fix the problem
 
+// My Answer:
+// The program tries to loop over the author object using a for...of loop,
+// but that doesn’t work because regular objects like author aren’t something you can loop over with for...of.
+
+// Objects in JavaScript don’t naturally know how to "be looped over" the way arrays or strings do.
+//  That’s why it crashes and throws a TypeError.
+
+// Fixing the issue:
+// If you want to get all the values inside the object, you need to first turn those values into something that can be looped over,
+// like an array. You can use Object.values(author) to grab all the values as an array, and then you can use for...of on it.
+// In short, Object.values() makes it work because it turns the object values into an array, and arrays are totally cool with for...of.
+
+
+
 const author = {
   firstName: "Zadie",
   lastName: "Smith",
@@ -11,6 +25,13 @@ const author = {
   alive: true,
 };
 
-for (const value of author) {
+for (const value of Object.values(author)) {
   console.log(value);
+
+// More interestingly, is that you can extract keys as well.
+// You can have a look on the following syntax.
+
+// for (const key of Object.keys(author)) {
+//   console.log(key);}
+
 }
--- a/Sprint-2/debug/recipe.js
+++ b/Sprint-2/debug/recipe.js
@@ -11,5 +11,5 @@ const recipe = {
 };
 
 console.log(`${recipe.title} serves ${recipe.serves}
-  ingredients:
-${recipe}`);
+  ingredients:`)
+recipe.ingredients.forEach((element) => console.log(element));
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -20,7 +20,37 @@ as the object doesn't contains a key of 'c'
 // Given an empty object
 // When passed to contains
 // Then it should return false
-test.todo("contains on empty object returns false");
+
+test("contains on empty object returns false", () => {
+  const result = contains({}, "a");
+  expect(result).toBe(false);
+});
+
+test("contains with an existing property name returns true", () => {
+  const result = contains({ a: 1, b: 2 }, "a");
+  expect(result).toBe(true);
+});
+
+test("contains with a non-existent property name returns false", () => {
+  const result = contains({ a: 1, b: 2 }, "c");
+  expect(result).toBe(false);
+});
+
+test("contains with invalid parameters (like an array) returns false", () => {
+  const result = contains([], "a");
+  expect(result).toBe(false);
+});
+
+test("contains with null or undefined object returns false", () => {
+  expect(contains(null, "a")).toBe(false);
+  expect(contains(undefined, "a")).toBe(false);
+});
+
+test("contains with invalid property names returns false", () => {
+  const result = contains({ a: 1, b: 2 }, undefined);
+  expect(result).toBe(false);
+});
+
 
 // Given an object with properties
 // When passed to contains with an existing property name
--- a/Sprint-2/implement/lookup.js
+++ b/Sprint-2/implement/lookup.js
@@ -1,5 +1,13 @@
-function createLookup() {
-  // implementation here
+function createLookup(countryCurrencyPairs) {
+  if (!Array.isArray(countryCurrencyPairs)) {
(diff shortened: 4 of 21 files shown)


Sprint-2/debug/author.js
```diff
@@ -3,6 +3,20 @@
 // This program attempts to log out all the property values in the object.
 // But it isn't working. Explain why first and then fix the problem
 
+// My Answer:
+// The program tries to loop over the author object using a for...of loop,
+// but that doesn’t work because regular objects like author aren’t something you can loop over with for...of.
+
+// Objects in JavaScript don’t naturally know how to "be looped over" the way arrays or strings do.
```
ChrisCookOC, line 10: Great explanation, it is such a useful skill to be able to explain why things are happening succinctly.

Sprint-2/debug/author.js
```diff
@@ -11,6 +25,13 @@ const author = {
   alive: true,
 };
 
-for (const value of author) {
+for (const value of Object.values(author)) {
```
ChrisCookOC, line 28: Yup this is a good way to solve that problem :) There is another approach to iterating over objects that is also common, can you think of what that is?

Sprint-2/debug/recipe.js
```diff
@@ -11,5 +11,5 @@ const recipe = {
 };
 
 console.log(`${recipe.title} serves ${recipe.serves}
-  ingredients:
-${recipe}`);
+  ingredients:`)
+recipe.ingredients.forEach((element) => console.log(element));
```
ChrisCookOC, line 15: This neatly does the job, but there are other approaches that could mean only doing one console.log instead of many - can you get it to use only one console.log?

Sprint-2/implement/contains.test.js
```diff
@@ -20,7 +20,37 @@ as the object doesn't contains a key of 'c'
...
+  const result = contains({ a: 1, b: 2 }, "c");
+  expect(result).toBe(false);
+});
+
+test("contains with invalid parameters (like an array) returns false", () => {
+  const result = contains([], "a");
+  expect(result).toBe(false);
+});
+
+test("contains with null or undefined object returns false", () => {
```
ChrisCookOC, line 44: Nice to see null checks and such descriptive test names!

Sprint-2/implement/lookup.js
```diff
@@ -1,5 +1,13 @@
-function createLookup() {
-  // implementation here
+function createLookup(countryCurrencyPairs) {
+  if (!Array.isArray(countryCurrencyPairs)) {
+    return {}; // Return an empty object if the input is not a valid array
+  }
+
+  return countryCurrencyPairs.reduce((lookup, [country, currency]) => {
```
ChrisCookOC, line 6: Reduce is a great way to do this!

Sprint-2/implement/lookup.test.js
```diff
@@ -1,6 +1,42 @@
...
+    ['GB', 'GBP'],
+  ]);
+  expect(result).toEqual({
+    US: 'USD',
+    CA: 'CAD',
+    GB: 'GBP',
+  });
+});
+
+test("createLookup with duplicate keys overwrites with the latest value", () => {
```
ChrisCookOC, line 27: A very interesting test case, because the expected behaviour is not necessarily clear from the brief! What influenced your decision to expect "US Dollars" as the answer instead of "USD"?

Sprint-2/implement/lookup.test.js
```diff
@@ -1,6 +1,42 @@
...
+  const result = createLookup([
+    ['US', 'USD'],
+    ['US', 'US Dollars'],
+  ]);
+  expect(result).toEqual({ US: 'US Dollars' });
+});
+
+test("createLookup with invalid input (non-array elements) returns an empty object", () => {
```
ChrisCookOC, line 35: I like this test case. But unfortunately it did inspire me to think of other ways to break it! Can you think of any ways you could pass in an Array (and so pass your checks), but still cause the reduce to fail? (not necessarily to fix it, more of just a thought exercise)

ChrisCookOC (commented):
Did you mean to add all those Sprint-1 files?

Sprint-2/debug/author.js
Pezhman-Azizi: Thanks  that is really reassuring

Sprint-2/debug/author.js
Pezhman-Azizi: thanks for your comment. As I have mentioned in the comment within the for loop there is another way, but i have commented out due to not getting errors.

Apart from that there is an alternative way which is as follows:

for (const [key, value] of Object.entries(author)) {
  console.log(key, value); // Logs key and its corresponding value
}

Sprint-2/implement/lookup.test.js
Pezhman-Azizi: Thanks for your comment.  I actually looked into the matter and surprised how many other approaches are out there:
Here are my findings:

1. Nested Arrays
If one or more elements of the input array are not key-value pairs but instead nested arrays, the function might break when attempting to destructure the elements.
 The reduce may fail when destructuring an element that is not a proper [key, value] pair.

2. Non-Array Elements Inside the Input Array
If some elements in the input array are not arrays (e.g., strings, numbers, or objects), destructuring them will fail.
 Attempting to destructure 'InvalidElement' into [key, value] will throw a TypeError.

Sprint-2/debug/author.js
ChrisCookOC: So what I'm thinking about is a different type of for loop. You might not have come across it yet - a for...in loop. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in

Can you see how that would work also here? :)

Sprint-2/implement/lookup.test.js
ChrisCookOC: yes both of those were examples I was thinking of! 🙂

Pezhman-Azizi: thanks @ChrisCookOC.

Sprint-2/debug/author.js
Pezhman-Azizi: Thanks for the suggestion and the link! I see how a for...in loop could be used here to iterate over the keys of the object directly. For example, for (const key in author) { console.log(key, author[key]); } would allow us to access both keys and their values concisely. I also learned that for...in might include inherited properties, so adding a hasOwnProperty() check could help if we want only the object's own properties. It’s great to have this approach alongside Object.entries() for different use cases

Sprint-2/debug/recipe.js
Pezhman-Azizi: Instead of logging each ingredient separately, I can modify the code to use just one console.log by joining all the ingredients into a single string. For example:

console.log(`Ingredients:\n${recipe.ingredients.join('\n')}`);

PR #260, opened 2024-12-29
London | Emmanuel Gessessew | Module-Data-Groups| WEEK 1

Emmanuelgessessew:
## Learners, PR Template

Self checklist

- [ yes] I have committed my files one by one, on purpose, and for a reason
- [ yes] My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/contributing/)

## Changelist

Briefly explain your PR.
i made changes and run all the tests
## Questions

### DIFF
--- a/Sprint-1/fix/median.test.js
+++ b/Sprint-1/fix/median.test.js
@@ -24,3 +24,23 @@ describe("calculateMedian", () => {
     expect(list).toEqual([1, 2, 3]);
   });
 });
+
+// median.js
+
+function calculateMedian(arr) {
+  // Create a copy to avoid modifying the input array
+  const sortedArr = [...arr].sort((a, b) => a - b);
+  
+  const len = sortedArr.length;
+  const mid = Math.floor(len / 2);
+
+  if (len % 2 === 0) {
+    // Even-length array: average of the two middle values
+    return (sortedArr[mid - 1] + sortedArr[mid]) / 2;
+  } else {
+    // Odd-length array: middle value
+    return sortedArr[mid];
+  }
+}
+
+module.exports = calculateMedian;
--- a/Sprint-1/implement/max.js
+++ b/Sprint-1/implement/max.js
@@ -1,4 +1,16 @@
-function findMax(elements) {
-}
-
-module.exports = findMax;
+;
+function findMax(arr) {
+    // Filter out non-numerical values and find the maximum number
+    const numbers = arr.filter(item => typeof item === 'number');
+    
+    // Return -Infinity if there are no numbers in the array
+    if (numbers.length === 0) {
+      return -Infinity;
+    }
+  
+    // Return the maximum value in the array
+    return Math.max(...numbers);
+  }
+  
+  module.exports = findMax;
+  
--- a/Sprint-1/implement/sum.js
+++ b/Sprint-1/implement/sum.js
@@ -1,4 +1,9 @@
 function sum(elements) {
-}
-
-module.exports = sum;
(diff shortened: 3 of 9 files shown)


Sprint-1/fix/median.test.js
```diff
@@ -24,3 +24,23 @@ describe("calculateMedian", () => {
...
+
+  if (len % 2 === 0) {
+    // Even-length array: average of the two middle values
+    return (sortedArr[mid - 1] + sortedArr[mid]) / 2;
+  } else {
+    // Odd-length array: middle value
+    return sortedArr[mid];
+  }
+}
+
```
cjyuan, line 45: Why include another implementation of the function in the test script?

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,16 @@
...
+    
+    // Return -Infinity if there are no numbers in the array
+    if (numbers.length === 0) {
+      return -Infinity;
+    }
+  
+    // Return the maximum value in the array
+    return Math.max(...numbers);
+  }
+
```
cjyuan, line 14: What do you expect from the following function calls?

```
findMax([NaN])
findMax([0, NaN, 1])
```

Sprint-1/implement/sum.test.js
```diff
@@ -33,4 +39,22 @@ test.todo("given an empty array, returns 0")
...
+// Then it should return the least surprising value given how it behaves for all other input
+
+
+  test("returns the correct sum for an array with negative numbers", () => {
+    expect(sum([10, -20, 30, -10])).toBe(10);
+  });
+
+  test("returns the correct sum for an array with decimal/float numbers", () => {
+    expect(sum([1.5, 2.5, 3.5])).toBe(7.5);
+  });
```

So the following could happen
```

```

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,9 @@
 function sum(elements) {
-}
-
-module.exports = sum;
+    // Filter the array to include only numbers and calculate the sum
+    return elements
+      .filter(item => typeof item === 'number')
+      .reduce((acc, num) => acc + num, 0);
+  }
```
cjyuan, line 6: What do you expect from the following function calls?

```
sum([NaN, 1]);
sum([Infinity, -Infinity]);
```

cjyuan (commented):
Code is very solid. I only have few suggestions and challenges for you.

Sprint-1/fix/median.test.js
Emmanuelgessessew: The implementation in the test script was meant as a reference for comparison to ensure accuracy. i can remove it and rely on predefined expected values or move it to a separate file if needed. Let me know your preference

Sprint-1/implement/max.js
Emmanuelgessessew: Both calls return unexpected results (-Infinity or NaN) due to Math.max() not handling NaN. I’ll update the function to filter out NaN values before processing.

Sprint-1/implement/sum.test.js
Emmanuelgessessew: const tolerance = 1e-10;
expect(Math.abs(sum([1.2, 0.6, 0.005]) - 1.805)).toBeLessThan(tolerance); how do u see this?

Sprint-1/implement/sum.js
Emmanuelgessessew: The current function returns NaN for both cases. I’ll update it to handle NaN and Infinity values properly, either by filtering or returning a more meaningful result

PR #280, opened 2025-01-07
Use more semantic button selectors in test

illicitonion:
Mandating that these buttons be `<i>` tags feels pretty weird. Buttons are great.


40thieves (changes requested):
This is an improvement, but I think we should probably switch the variable names within the test too: e.g. `tickIcon` -> `doneBtn`?

Perhaps I'm stating the obvious here, but `<i class="fa-check"></i>` was the recommended pattern for Font Awesome (via Bootstrap) icons - hence the naming here that appears to have got out of sync with the impl.

PR #281, opened 2025-01-08
CYF London | Rihanna Poursoltani | Module-Data-Groups | Sprint 1

RihannaP:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -4,8 +4,14 @@
 
 function calculateMedian(list) {
   const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
+  if (list.length % 2 !== 0){
+    const median = list[middleIndex];
+    return median;
+  }
+  else{
+    median = ((list[middleIndex] + list[middleIndex - 1]) / 2)
+  }
   return median;
 }
-
+//console.log(calculateMedian([1, 2, 3, 4]))
 module.exports = calculateMedian;
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -3,13 +3,15 @@
 // If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  if (list.length % 2 !== 0){
-    const median = list[middleIndex];
+  const sortedList = list.sort(function(a, b){return a - b});
+  const middleIndex = Math.floor(sortedList.length / 2);
+  let median;
+  if (sortedList.length % 2 !== 0){
+    median = sortedList[middleIndex];
     return median;
   }
   else{
-    median = ((list[middleIndex] + list[middleIndex - 1]) / 2)
+    median = ((sortedList[middleIndex] + sortedList[middleIndex - 1]) / 2)
   }
   return median;
 }
--- a/Sprint-1/implement/dedupe.test.js
+++ b/Sprint-1/implement/dedupe.test.js
@@ -23,13 +23,13 @@ test("given an empty array, it returns an empty array", () => {
 // Given an array with no duplicates
 // When passed to the dedupe function
 // Then it should return a copy of the original array
-test("Given an array with no duplicates", () =>{
+test("Given an array with no duplicates,it return a copy of the original array", () =>{
     expect(dedupe(["c", "d", "f"])).toEqual(["c", "d", "f"]);
 });
 // Given an array with strings or numbers
 // When passed to the dedupe function
 // Then it should remove the duplicate values, preserving the first occurence of each element
-test("Given an array with strings or numbers", () =>{
+test("Given an array with strings or numbers,it removes the duplicate values and then return the result", () =>{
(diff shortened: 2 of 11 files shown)


Sprint-1/fix/median.js
```diff
@@ -4,8 +4,14 @@
 
 function calculateMedian(list) {
   const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
```
cjyuan, line 7: The arrays specified in `median.test.js` happened to be sorted.

Sprint-1/fix/median.js
```diff
@@ -4,8 +4,14 @@
...
+  if (list.length % 2 !== 0){
+    const median = list[middleIndex];
+    return median;
+  }
+  else{
+    median = ((list[middleIndex] + list[middleIndex - 1]) / 2)
```
cjyuan, line 12: The `median` declared at line 8 is a local variable useable inside the `{ .... }` (block) where it is declared (that is, between lines 8 and 10).

The `median` at lines 12 and 14 are not declared anywhere. Can you ensure they are is properly declared?

Sprint-1/implement/dedupe.test.js
```diff
@@ -16,12 +16,20 @@ E.g. dedupe([1, 2, 1]) target output: [1, 2]
...
 // When passed to the dedupe function
 // Then it should return a copy of the original array
-
+test("Given an array with no duplicates", () =>{
+    expect(dedupe(["c", "d", "f"])).toEqual(["c", "d", "f"]);
+});
 // Given an array with strings or numbers
 // When passed to the dedupe function
 // Then it should remove the duplicate values, preserving the first occurence of each element
+test("Given an array with strings or numbers", () =>{
```
cjyuan, line 32: What is the expected result when given an array with strings or numbers?

Sprint-1/implement/dedupe.test.js
```diff
@@ -16,12 +16,20 @@ E.g. dedupe([1, 2, 1]) target output: [1, 2]
...
-test.todo("given an empty array, it returns an empty array");
 
+test("given an empty array, it returns an empty array", () => {
+    expect(dedupe([])).toEqual([]);
+})
 // Given an array with no duplicates
 // When passed to the dedupe function
 // Then it should return a copy of the original array
-
+test("Given an array with no duplicates", () =>{
```
cjyuan, line 26: What is the expected result when the given array contains no duplicates?

Sprint-1/implement/max.test.js
```diff
@@ -16,28 +16,49 @@ const findMax = require("./max.js");
...
-
+test("Given an array with decimal numbers", () =>{
+    expect(findMax([2.5, 8.4, 6,])).toEqual(8.4);
+    expect(findMax([0, 3, 6.4])).toEqual(6.4);
+});
 // Given an array with non-number values
 // When passed to the max function
 // Then it should return the max and ignore non-numeric values
-
+test("Given an array with non-number values", () =>{
```
cjyuan, line 54: This description can also mean the array contains only non-numbers.

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,42 @@ const sum = require("./sum.js");
...
+    expect(sum([-1, 9, -2])).toEqual(6);
+});
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
-
+test("Given an array with decimal/float numbers", () =>{
+    expect(sum([2.5, 1.5, 6])).toEqual(10);
+    expect(sum([1.3, 4])).toEqual(5.3);
+});
```

So the following could happen
```

```

cjyuan (commented):
Code is very solid. I only have a few comments and suggestions.

Sprint-1/fix/median.js
RihannaP: Thank you for mentioning that. I can fix it with sort() method.

Sprint-1/fix/median.js
RihannaP: Good point. I can declare this variable out of `if... else` so it can be global and both can have access to that. Or I was thinking that remove this variable and return the result directly (without assigning to any variable). This works but I think is not a good and readable code.

Sprint-1/implement/dedupe.test.js
RihannaP: Because there is no duplication so we expect that it return a copy of the original array

Sprint-1/implement/dedupe.test.js
RihannaP: We expect : It removes the duplicate values and then return the result.

Sprint-1/implement/max.test.js
RihannaP: Thank you for bringing that to me. I wasn't aware how much descriptions and its details can be important when somebody is reviewing that.

Sprint-1/implement/sum.test.js
RihannaP: Thank you so much. I noticed this differences before but I didn't know what is the reason. It was challenging and I tried lots of way that didn't work (like returning rounded result / returning string instead of number) and at the end scaling by `1e12` and `Math.round()` return the correct result.

PR #288, opened 2025-01-13
CYF London | Rihanna Poursoltani | Module-Data-Groups | Sprint 3 | Alarm Clock App

RihannaP:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -1,4 +1,29 @@
-function setAlarm() {}
+function setAlarm() {
+  let timeRemaining = document.getElementById("alarmSet").value;
+
+  function formatTime(seconds) {  //format time (MM : SS)
+    const minutes = Math.floor(seconds / 60); //minutes
+    const remainingSeconds = seconds % 60; // remaining seconds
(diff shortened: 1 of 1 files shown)


Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,29 @@
-function setAlarm() {}
+function setAlarm() {
+  let timeRemaining = document.getElementById("alarmSet").value;
```
cjyuan, line 2: If the user don't enter a "normal" value in the input field, how would you ensure your app could always behave properly.

cjyuan (commented):
If the user enters a time and then clicks the "Set Alarm" button multiple times, the countdown clock will not display properly. Can you fix this "glitch"?

You missed a required change (related to `index.html`) specified in the `readme.md` file.

Sprint-3/alarmclock/alarmclock.js
RihannaP: Good point. Because we had a input `type="number"` I thought it would handle it and I can see for null/negative/ float number I have a problem. So I fixed it with the if statement and parseInt method. Thank you

PR #297, opened 2025-02-21
GLASGOW | AIDA ZHUKOVA | MODULE DATA GROUPS | WEEK 3

aidazhukova:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.
All exercises in Sprint 1 done.

## Questions

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -3,9 +3,18 @@
 // If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  const copyList = [...list];
+  if (copyList.length % 2 !== 0) {
+    const middleIndex = Math.floor(copyList.length / 2);
+    const median = copyList.splice(middleIndex, 1)[0];
+    return median;
+  } else {
+    const middleEvenIndex = copyList.length / 2 - 1;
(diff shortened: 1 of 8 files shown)


Sprint-1/fix/median.js
```diff
@@ -3,9 +3,18 @@
 // If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  const copyList = [...list];
+  if (copyList.length % 2 !== 0) {
+    const middleIndex = Math.floor(copyList.length / 2);
```
day-lee, line 8: Good job on using the spread operator to copy the array so it won't modify the original array. In this line of code, I can see that you can reduce some repetition by moving the `middleIndex` variable outside of the if-else statement so you can use it for both odd and even cases. Can you try to refactor the code?

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,11 @@
-function sum(elements) {
+function sum(arr) {
+  const numbersArray = arr.filter((num) => typeof num === "number");
+
+  let arraySum = 0;
+  for (let i = 0; i < numbersArray.length; i++) {
```
day-lee, line 5: It works fine to use an imperative for loop in this case, but as a challenge, can you think of a declarative way to iterate over the elements using array methods? You can chain the method directly to filter to make it more concise as well. 
FYI, imperative vs declarative: https://dev.to/siddharthshyniben/explained-imperative-vs-declarative-programming-577g

PR #308, opened 2025-03-01
ITP-25/ LONDON/LOVETH-OKAFOR/Module-Data-Groups/WEEK1/SPRINT--1

loveth900:
## Learners, PR Template

Self checklist

[X] I have committed my files one by one, on purpose, and for a reason
[X] I have titled my PR with COHORT_NAME | FIRST_NAME LAST_NAME | REPO_NAME | WEEK 
[X] I have tested my changes
[X] My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/contributing/)
[X] My changes meet the [requirements](./README.md) of this task

## Changelist

Briefly explain your PR.

## Questions


Sprint-1/implement/dedupe.js
```diff
@@ -1 +1,14 @@
...
+function dedupe(arr) {
+    const uniqueArray = [];
+    const seen = new setImmediate();
+
+    for (const item of arr) {
+      if (!seen.has(item)) {
+        uniqueArray.push(item);
+        seen.add(item);
+      }   
+    }
```
day-lee, line 12: When you use `Set` object, you don't need the lines of code, because each value in a Set can only appear once. Can you change the code to convert Set object into an array instead?

Sprint-1/implement/dedupe.js
```diff
@@ -1 +1,14 @@
-function dedupe() {}
+
+
+function dedupe(arr) {
+    const uniqueArray = [];
```
day-lee, line 4: This is unnecessary if you use Set object.

Sprint-1/implement/dedupe.js
```diff
@@ -1 +1,14 @@
...
+    const seen = new setImmediate();
```
day-lee, line 5: I guess you intended to use Set() object here? In that case, you should declare it as follows:
`const newArr = new Set(arr); `

Sprint-1/implement/dedupe.test.js
```diff
@@ -25,3 +26,24 @@ test.todo("given an empty array, it returns an empty array");
...
+test("given an array with no duplicates, it returns a copy of the original array", () => {
+  expect(dedupe([1, 2, 3])).toEqual([1, 2, 3]);
+  expect(dedupe(['a', 'b', 'c'])).toEqual(['a', 'b', 'c']);
+
+});
+
+test("given an array with duplicates, it removes duplicates and preserves first occurrences", () => {
+  expect(dedupe(['a', 'a', 'a', 'b', 'b', 'c',])).toEqual(['a', 'b', 'c']);
+  expect(dedupe([5, 1, 1, 2, 3, 2, 5, 8])).toEqual([5, 1, 2, 3, 8]);
+  expect(dedupe([1, 2, 3])).toEqual([1, 2]);
```
day-lee, line 47: This would cause the test to fail. Can you add correct test case?

Sprint-1/implement/max.test.js
```diff
@@ -41,3 +40,29 @@ test.todo("given an empty array, returns -Infinity");
...
+        // Update max if the current element is larger
+        if (elements[i] > max) {
+            max = elements[i];
+        }
+
+      }
+
+     }
+
+      return max; // Return the maximum value found
```
day-lee, line 65: This is not a test. This code should be placed in the `./Sprint-1/implement/max.js` file. Can you move it to there? and write the test suite in the max.test.js?

Sprint-1/implement/sum.test.js
```diff
@@ -34,3 +33,18 @@ test.todo("given an empty array, returns 0")
 // Given an array with only non-number values
 // When passed to the sum function
 // Then it should return the least surprising value given how it behaves for all other inputs
+
+
+function sum(elements) {
```
day-lee, line 38: How about decimal/float numbers? if I pass in [1.2, 2.4], it returns 3.5999999999999996. Can you change the code so it can handle decimal/float values as well?

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,17 @@
...
+
+    // Iterate over the elements of the array
+    for (let i = 0; i < elements.length; i++) {
+        // Check if the current element is a valid number (ignores non-numeric values)
+        if (typeof elements[i] === 'number' && !isNaN(elements[i])) {
+          total += elements[i]; // Add to the sum
+        }
+    }
+
+    return total; // Return the final sum
```
day-lee, line 13: How about decimal/float numbers? if I pass in [1.2, 2.4], it returns 3.5999999999999996. Can you change the code so it can handle decimal/float numbers as well?

Sprint-1/implement/sum.test.js
```diff
@@ -34,3 +33,18 @@ test.todo("given an empty array, returns 0")
...
+    // Loop through the array
+    for (let i =0; i < elements.length; i++) {
+        // Check if the element is a valid number
+        if (typeof elements[i] === 'number' && !isNaN(elements[i])) {
+            total += elements[i];
+        }
+    
+    } 
+    return total;      
+}
```
day-lee, line 50: This file should include test cases for sum(). Could you write test cases to check your code?

Sprint-1/refactor/includes.js
```diff
@@ -11,3 +11,8 @@ function includes(list, target) {
 }
 
 module.exports = includes;
+
```
day-lee, line 14: In this exercise, you need to refactor the implementation of includes to use a for...of loop. Currently, the example code uses for (let index = 0; index < list.length; index++) { ... }. Your task is to convert this into a for...of loop. Can you make that change accordingly?

Sprint-1/fix/median.test.js
```diff
@@ -24,3 +24,5 @@ describe("calculateMedian", () => {
     expect(list).toEqual([1, 2, 3]);
```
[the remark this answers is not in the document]
loveth900: i have implemented the test and i run it its all passed

Sprint-1/implement/dedupe.js
loveth900: alright

Sprint-1/implement/dedupe.js
loveth900: i have done the declearation as followed

Sprint-1/implement/dedupe.js
loveth900: completed

Sprint-1/implement/dedupe.test.js
loveth900: i have fix it

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,22 @@
 function findMax(elements) {
+    if (elements.length === 0) {
+     return undefined; // Return undefined if the array is empty
+
+
+}
+
+let max = elements[0]; // Start by assuming the first element is max
```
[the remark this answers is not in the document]
loveth900: fixed

Sprint-1/implement/max.test.js
loveth900: okey dear i have movd the code there

Sprint-1/implement/dedupe.js
```diff
@@ -1,14 +1,15 @@
...
+  const seen = new Set();  // Corrected from 'setImmediate' to 'Set'
+
+  for (const item of arr) {
+    if (!seen.has(item)) {  // Check if the item is already in the Set
+      uniqueArray.push(item);  // If not, add it to the result array
+      seen.add(item);  // Add the item to the Set to track it
+    }   
+  }
+
+  return uniqueArray;  // Return the array with unique values
```
day-lee, line 14: Good job on refactoring the code. This is one way to remove duplicates, but using a Set object offers a much simpler approach. You can declare arr within a Set object like this: const newArr = new Set(arr);. The Set will automatically remove any duplicates from arr. You can then convert the Set object back into an array and return it. Can you think of a solution for this?

Sprint-1/fix/median.test.js
```diff
@@ -26,3 +26,25 @@ describe("calculateMedian", () => {
 });
 
 
+function calculateMedian(arr) {
```
day-lee, line 29: I can't see the test cases here.

Sprint-1/implement/dedupe.test.js
day-lee: My bad, this was correct!

Sprint-1/fix/median.test.js
day-lee: Oh, the test code was in the folded lines. I only saw your calculateMedian function below and thought you misplaced the code. Sorry for the misunderstanding. 

To avoid confusion, please place the code in the correct files. Remove the calculateMedian function from the test.js file please.

Sprint-1/implement/sum.test.js
loveth900: fixed the test case for sum()

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,70 @@
...
+// When passed to the max function
+// Then it should return the largest decimal number
+
+// Given an array with non-number values
+// When passed to the max function
+// Then it should return the max and ignore non-numeric values
+
+// Given an array with only non-number values
+// When passed to the max function
+// Then it should return the least surprising value given how it behaves for all other inputs
```
day-lee, line 42: These should be implemented in the max.test.js file. 
Please make sure to place your `function findMax(elements)` in the `max.js` file and test cases in the `max.test.js` file.
If you are not sure about it, let me know!

Sprint-1/refactor/includes.js
day-lee: Good job! Thank you for studying hard!💪 I reviewed again, just needs to implement some tests and move the codes to the correct files. Keep up the good work.

Sprint-1/implement/max.js
loveth900: please dear mentor am not sure about what you are saying, can you elaborate more.

day-lee: There are two files related to `findMax` which finds the largest number in the array. 

**1. Sprint-1/implement/max.js**
this file shuld only contain the codes below (codes are from your file): 
```
function findMax(elements) {
  if (elements.length === 0) {
      return undefined; // Return undefined if the array is empty
  }

  // Initialize the max value as the first valid number
  let max = Number.NEGATIVE_INFINITY;  // Set a default max that will be smaller than any number

  for (let i = 0; i < elements.length; i++) {
      // Check if the current element is a number before comparing
      if (typeof elements[i] === 'number') {
          if (elements[i] > max) {
              max = elements[i]; // Update max if a larger number is found
          }
      } else {
          console.warn(`Non-numeric value encountered: ${elements[i]}`); // Warn about non-numeric values
      }
  }

  // Return undefined if no valid number was found
  return max === Number.NEGATIVE_INFINITY ? undefined : max;
}
```

**2. Sprint-1/implement/max.test.js**
This file should only contain the test not the funciton. 
Your job is to write seven tests for findMax() following the instruction.
```
// Given an empty array

// Given an array with one number
// Then it should return that number

// Given an array with both positive and negative numbers
// Then it should return the largest number overall

// Given an array with just negative numbers

```

If you are not sure about how to write tests, please recap the prep materials.
Hope it helps you to understand!

loveth900: hi dear mentor @day-lee  i have fixe the correction, thanks alot for letting me know this, i really mess up here, but thanks for your time and patience.

day-lee: Sorry for the late reply. I'm glad it made sense to you. Good job👍

PR #313, opened 2025-03-02
ITP-JAN-25 | London | Saba Farjamfard | Module Data Groupe | Sprint 1 | Week 1

sabafarjamfard:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -3,9 +3,13 @@
 // If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
 
 function calculateMedian(list) {
+  list = list.sort((a, b) => a - b);
   const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  if (list.length % 2 === 0) {
+    return (list[middleIndex - 1] + list[middleIndex]) / 2;
+  } else {
+    return list[middleIndex];
+  }
 }
 
 module.exports = calculateMedian;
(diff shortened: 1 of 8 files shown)


Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,11 @@
 function sum(elements) {
+  if (elements.length === 0) {
+    return 0;
+  }
+
+  return elements
+    .filter((item) => typeof item === "number")
+    .reduce((acc, num) => acc + num, 0);
```
cjyuan, line 8: What do you expect from the following function calls (on extreme cases)?

```
sum([NaN, 1]);
sum([Infinity, -Infinity]);
```

cjyuan (commented):
Code looks good.

`calculateMedian()` can use some improvement. Feel free to mark this as completed after you have made the necessary changes.

Sprint-1/fix/median.js
```diff
@@ -3,9 +3,13 @@
 // If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
 
 function calculateMedian(list) {
+  list = list.sort((a, b) => a - b);
```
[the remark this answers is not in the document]
sabafarjamfard: Thanks for your feedback! I've updated the function 
 It makes a copy of the array before sorting, to avoid changing the original. I also improved the test by using an unsorted list and checking properly that it doesn’t get modified

Sprint-1/implement/sum.js
sabafarjamfard: Thanks for your feedback! I’ve reviewed the edge cases you mentioned (sum([NaN, 1]) and sum([Infinity, -Infinity])), and I’ve made the following updates to the function

PR #315, opened 2025-03-03
Glasgow ITP-JAN-25 | Aiah Ibrahim | Module-Data-Groups | WEEK 8

aiahibrahim:
Self checklist

### DIFF
--- a/Sprint-1/implement/max.test.js
+++ b/Sprint-1/implement/max.test.js
@@ -16,28 +16,48 @@ const findMax = require("./max.js");
 // When passed to the max function
 // Then it should return -Infinity
 // Delete this test.todo and replace it with a test.
-test.todo("given an empty array, returns -Infinity");
+test("given an empty array, returns -Infinity", () => {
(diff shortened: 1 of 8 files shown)


Sprint-1/implement/sum.js
KudahShambare, line 1: Just a quick question, can you explain why you filtered NaN, it is also of type number, why not just include it.

Sprint-1/implement/max.test.js
KudahShambare, line 1: Can you please explain why did you use toBe instead of toEqual

KudahShambare (approved):
- Great work!! Just improve on your comments

https://stackoverflow.blog/2021/12/23/best-practices-for-writing-code-comments/

Sprint-1/implement/sum.js
aiahibrahim: We filter out NaN because if we include it, the whole sum turns into NaN, which isn’t useful. Since any math operation with NaN results in NaN, we'd end up with an unusable result instead of a meaningful number.

PR #319, opened 2025-03-03
ITP-JAN-25 | LONDON | Bakytbek Sydykov | Module-Data-Groups | sprint 1

Bahasyd:
## Learners, PR Template

Self checklist

## Changelist

n/a

## Questions

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -4,8 +4,11 @@
 
 function calculateMedian(list) {
   const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+   if (list.length % 2 === 0) {
+    return (list[middleIndex - 1] + list[middleIndex]) / 2; 
+  } else {
+    return list[middleIndex]; 
+  }
 }
-
+console.log(calculateMedian([1,2,3,6,7,9]));
 module.exports = calculateMedian;
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -2,12 +2,24 @@
 // Start by running the tests for this function
 // If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
 
-function calculateMedian(list) {
+/*function calculateMedian(list) {
   const middleIndex = Math.floor(list.length / 2);
    if (list.length % 2 === 0) {
     return (list[middleIndex - 1] + list[middleIndex]) / 2; 
   } else {
     return list[middleIndex]; 
+  }
+}
+console.log(calculateMedian([1,2,3,6,7,9]));
+module.exports = calculateMedian;*/
+function calculateMedian(list) {
+  const sortedList = [...list].sort((a, b) => a - b);
+  const middleIndex = Math.floor(sortedList.length / 2);
+
+  if (sortedList.length % 2 === 0) {
+    return (sortedList[middleIndex - 1] + sortedList[middleIndex]) / 2;
+  } else {
+    return sortedList[middleIndex];
   }
 }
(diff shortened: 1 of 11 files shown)


Sprint-1/fix/median.js
```diff
@@ -4,8 +4,11 @@
 
 function calculateMedian(list) {
   const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+   if (list.length % 2 === 0) {
+    return (list[middleIndex - 1] + list[middleIndex]) / 2; 
+  } else {
+    return list[middleIndex]; 
+  }
```
cjyuan, line 11: The arrays specified in `median.test.js` happened to be sorted. (Those are poorly selected test values)

Also, because `.sort()` can modify the content in an array. So it is better to sort a cloned version of `list`.

Sprint-1/fix/median.test.js
```diff
@@ -23,4 +23,5 @@ describe("calculateMedian", () => {
 
     expect(list).toEqual([1, 2, 3]);
   });
+
```
cjyuan, line 26: Can you figure out why the "doesn't modify the input" test in `median.test.js` may fail to check what it describes, and improve the test code accordingly?

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,37 @@ const sum = require("./sum.js");
...
+test("given an array containing negative numbers", () => {
+expect(sum([-3,-2])).toEqual(-5);
+});
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
-
+test("given an array with decimal/float numbers", () => {
+    expect(sum([2.3,3.1,2.5,7.6])).toEqual(15.5)
+});
```
cjyuan, line 37: Decimal numbers in most programming languages (including JS) are internally represented in "floating point number" format. Floating point arithmetic is not exact. For example, the result of `46.5678 - 46 === 0.5678` is false because `46.5678 - 46` only yield a value that is very close to `0.5678`. Even changing the order in which the program add/subtract numbers can yield different values.

So the following could happen
```

```

cjyuan (commented):
Code looks good. Just a few suggestions.

Sprint-1/fix/median.test.js
Bahasyd: Shallow copy ([...list]): Ensures the original order and content are preserved.
toStrictEqual instead of toEqual: Ensures the entire structure and reference remain unchanged.
This will make sure that 'calculateMedian' does not modify the original input.

Sprint-1/fix/median.js
Bahasyd: Okey. I did it.

Sprint-1/implement/sum.test.js
Bahasyd: Okey. I understood. Our test using .toEqual() may fail due to floating-point precision issues. Instead, we should use .toBeCloseTo() in Jest, which allows for small differences due to floating-point arithmetic.
     Main point is:
      .toEqual() checks for exact equality, which can fail due to floating-point rounding errors.
      .toBeCloseTo(expected, precision) allows for a margin of error, making the test more reliable.

Sprint-1/fix/median.test.js
```diff
@@ -16,11 +16,24 @@ describe("calculateMedian", () => {
...
   });
+  
+});*/
+test("doesn't modify the input", () => {
+  const list = [1, 2, 3];
+  const originalList = [...list]; 
+
+  calculateMedian(list);
+
+  expect(list).toStrictEqual(originalList);
```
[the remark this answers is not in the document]
Bahasyd: Okey. I've got it.

PR #322, opened 2025-03-04
WM | ITP-JAN-25 | HATEF_EIDI | Module-Data-Groups | sprint 1

HatefEidi:
## Learners, PR Template

Self checklist

- [ *] I have committed my files one by one, on purpose, and for a reason
- [ *] I have titled my PR with COHORT_NAME | FIRST_NAME LAST_NAME | REPO_NAME | WEEK 
- [ *] I have tested my changes
- [ *] My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/contributing/)
- [ *] My changes meet the [requirements](./README.md) of this task

## Changelist

Calculating and testing mean and median of an array with odd or even number of elements, Prep exercises.
Diving deeper into JavaScript Node properties and functions. 
Deduplicating repeated elements of an array. 
Filtering number values of an array and storing in to a new array, expanding the array by spread operator(...) to be able to run Math.max
using reduce function on an array of number to add all the elements.
Refactoring and use of for ... of to iterate through all the elements of an array.
Solving stretch problem, using all the implemented code in sprint 1, by exporting and importing each file.

## Questions


Sprint-1/implement/dedupe.test.js
```diff
@@ -16,12 +16,28 @@ E.g. dedupe([1, 2, 1]) target output: [1, 2]
...
 
 // Given an array with strings or numbers
 // When passed to the dedupe function
 // Then it should remove the duplicate values, preserving the first occurence of each element
+test("Given an array with strings or numbers", () => {
+    const result = dedupe(['a', 'a', 'a', 'b', 'b', 'c']);
+    expect(result).toEqual(['a', 'b', 'c']);
+});
+
+test("Given an array with mixed types", () => {
```
suziebrown, line 40: Make sure the names of tests accurately describe what they are testing

Sprint-1/implement/max.test.js
```diff
@@ -16,28 +16,50 @@ const findMax = require("./max.js");
...
 // Then it should return the max and ignore non-numeric values
+test("Given an array with non-number values" ,()=>{
+    expect(findMax(['hey', 10, 'hi', 60, 10])).toBe(60);
+});
 
 // Given an array with only non-number values
 // When passed to the max function
 // Then it should return the least surprising value given how it behaves for all other inputs
+test("Given an array with only non-number values" ,()=>{
+    expect(findMax(['hey', 'hi', 'hello', 'world'])).toBe(undefined);
```
suziebrown, line 64: `undefined` is a very reasonable result in this case, but is there a different result that would be more consistent with how the function behaves in other cases?

Sprint-1/implement/max.test.js
```diff
@@ -16,28 +16,50 @@ const findMax = require("./max.js");
 // When passed to the max function
 // Then it should return -Infinity
 // Delete this test.todo and replace it with a test.
-test.todo("given an empty array, returns -Infinity");
+test("Given an empty array", () => {
```
suziebrown, line 19: The names of tests should usually include the test case *and* the expected result, for example "given an empty array, returns -Infinity"

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,43 @@ const sum = require("./sum.js");
...
+test("Given an array containing non-number values" ,()=>{
+    expect(sum(['hey', 10, 'hi', 60, 10])).toBe(80);
+});
+
 
 // Given an array with only non-number values
 // When passed to the sum function
 // Then it should return the least surprising value given how it behaves for all other inputs
+test("Given an array with only non-number values" ,()=>{
+    expect(sum(['hey', 'Hatef', 'hi', 'Isfahan', 'London', 'United Kingdom'])).toBe(undefined);
```
suziebrown, line 54: Also here, `undefined` is reasonable but is there a "less surprising" result?

suziebrown (approved):
Well done 👍
There are just a few improvements you could make to the names of tests - see comments

Sprint-1/implement/dedupe.test.js
HatefEidi: Correct, maybe Given an array with Number values was a better description. Updated

Sprint-1/implement/max.test.js
HatefEidi: I replaced undefined with "No Number was given"

Sprint-1/implement/max.test.js
HatefEidi: Thank you, updated

Sprint-1/implement/sum.test.js
HatefEidi: You mean more descriptive by less surprising? such as no number value was given?

suziebrown: As far as this function is concerned, an array with no numbers in it is equivalent to an empty array, because it will ignore all of the elements. So I think it would make more sense if the same result is produced for the empty array and the non-numeric array. Does that make sense?

HatefEidi: Ah yes totally, thanks for explanation, updated

PR #324, opened 2025-03-04
 LON | ITP JAN-25 | Amir Montazeri | Data groups | Sprint2 tasks

Amirmontazerii:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-2/debug/recipe.js
+++ b/Sprint-2/debug/recipe.js
@@ -10,6 +10,6 @@ const recipe = {
   ingredients: ["olive oil", "tomatoes", "salt", "pepper"],
 };
 
-console.log(`${recipe.title} serves ${recipe.serves}
-  ingredients:
-${recipe}`);
+console.log(`${recipe.title} serves ${recipe.serves}\ningredients:`);
+recipe.ingredients.forEach(ingredients => console.log(ingredients));
+  
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -20,16 +20,27 @@ as the object doesn't contains a key of 'c'
 // Given an empty object
 // When passed to contains
 // Then it should return false
-test.todo("contains on empty object returns false");
+test("Given an empty object, it should return false", () => {
+    expect(contains({}, "a")).toBe(false);
+});
 
 // Given an object with properties
 // When passed to contains with an existing property name
 // Then it should return true
+test("Given an object with properties,it should return true", () => {
+    expect(contains({a : 1 , b : 2} , "a")).toBe(true);
+});
 
 // Given an object with properties
 // When passed to contains with a non-existent property name
 // Then it should return false
+test("Given an object with properties,contains with a non-existent property name,it should return false", () => {
+    expect(contains({a : 1, b: 2}, "c")).toBe(false);
+});
 
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+test("Given invalid parameters like an array,it should return false or throw an error", () => {
+    expect(contains(["a","b","c"], "a")).toBe(false);
+});
--- a/Sprint-2/implement/querystring.js
+++ b/Sprint-2/implement/querystring.js
@@ -1,13 +1,24 @@
 function parseQueryString(queryString) {
   const queryParams = {};
-  if (queryString.length === 0) {
-    return queryParams;
-  }
+
+  if (!queryString) return queryParams;
+
   const keyValuePairs = queryString.split("&");
 
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
-    queryParams[key] = value;
+    const parts = pair.split("=");
+    const key = decodeURIComponent(parts[0]);
+    const value =
+      parts.length > 1 ? decodeURIComponent(parts.slice(1).join("=")) : "";
+
+    if (queryParams[key]) {
+      if (!Array.isArray(queryParams[key])) {
+        queryParams[key] = [queryParams[key]];
+      }
+      queryParams[key].push(value);
+    } else {
+      queryParams[key] = value;
+    }
   }
 
   return queryParams;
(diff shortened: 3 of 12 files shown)


Sprint-2/debug/recipe.js
```diff
@@ -10,6 +10,6 @@ const recipe = {
   ingredients: ["olive oil", "tomatoes", "salt", "pepper"],
 };
 
-console.log(`${recipe.title} serves ${recipe.serves}
-  ingredients:
-${recipe}`);
+console.log(`${recipe.title} serves ${recipe.serves}\ningredients:`);
+recipe.ingredients.forEach(ingredients => console.log(ingredients));
```
OracPrime, line 14: This works, but is confusing.  The arrow function acts on each ingredient in turn, not all of them at once, so it might be clearer to write
recipe.ingredients.forEach(ingredient => console.log(ingredient));

Variable naming is actually one of the harder parts of coding, but it's important because it really helps others understand your code

Sprint-2/implement/contains.test.js
```diff
@@ -20,16 +20,27 @@ as the object doesn't contains a key of 'c'
...
+test("Given an object with properties,contains with a non-existent property name,it should return false", () => {
+    expect(contains({a : 1, b: 2}, "c")).toBe(false);
+});
 
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+test("Given invalid parameters like an array,it should return false or throw an error", () => {
+    expect(contains(["a","b","c"], "a")).toBe(false);
+});
```
OracPrime, line 46: This tests one case.  However you might argue that if this code was a user of your function misunderstanding how to use it, then it is not helpful to just return false.  They need to know they've done something wrong.  So how would you make it throw an error in this case?

Sprint-2/implement/querystring.js
```diff
@@ -1,13 +1,24 @@
...
-    const [key, value] = pair.split("=");
-    queryParams[key] = value;
+    const parts = pair.split("=");
+    const key = decodeURIComponent(parts[0]);
+    const value =
+      parts.length > 1 ? decodeURIComponent(parts.slice(1).join("=")) : "";
+
+    if (queryParams[key]) {
+      if (!Array.isArray(queryParams[key])) {
+        queryParams[key] = [queryParams[key]];
```
OracPrime, line 16: This is an interesting approach, but it creates problems.  You need to think how a user of your function is going to be thinking, and what the constraints of the problem are.  If the query string contains the same key twice, is this valid, or should it be an error?  If it is valid, then we have the interesting case that queryParams[key] is either a string or an array of strings, which is legal in javascript, but can be awkward to deal with as a programmer (or at least inefficient if you are having to test of string or array before every access).  You could be consistent and return an array with a single string in it for non-duplicated keys.  But actually, I suspect that duplicate keys should be an error.

Sprint-2/implement/querystring.js
```diff
@@ -1,13 +1,24 @@
...
+  if (!queryString) return queryParams;
+
   const keyValuePairs = queryString.split("&");
 
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
-    queryParams[key] = value;
+    const parts = pair.split("=");
+    const key = decodeURIComponent(parts[0]);
+    const value =
```
OracPrime, line 11: Rather than splitting and rejoining, how might you do this with indexOf and slice?   Your method works, I just want you to consider alternatives.  In what circumstances is each approach better?  (you don't need to respond to that question here, just have a think about it - although feel free to do so if you want feedback on your answer)

Sprint-2/implement/querystring.test.js
```diff
@@ -3,10 +3,40 @@
...
+const parseQueryString = require("./querystring.js");
 
 test("parses querystring values containing =", () => {
   expect(parseQueryString("equation=x=y+1")).toEqual({
-    "equation": "x=y+1",
+    equation: "x=y+1",
   });
 });
+test("handles multiple key-value pairs", () => {
+  expect(parseQueryString("key1=value1&key2=value2")).toEqual({
```
OracPrime, line 14: You've correctly used toEqual here.  What would happen if you used toBe?  (again, if you understand your answer, no need to give it here)

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,13 @@
-function tally() {}
+function tally(items) {
+    if (!Array.isArray(items)) {
+      throw new Error("Input must be an array");
+    }
+  
+    return items.reduce((acc, item) => {
+      acc[item] = (acc[item] || 0) + 1;
```
OracPrime, line 7: Good use of || here, well done

Sprint-2/interpret/invert.js
```diff
@@ -6,24 +6,32 @@
...
 // e) Fix the implementation of invert (and write tests to prove it's fixed!)
+function invert(obj) {
+  const invertedObj = {};
+  for (const [key, value] of Object.entries(obj)) {
+    invertedObj[value] = key;
+  }
+  return invertedObj;
+}
+
+console.log(invert({ a: 1 }));
```
OracPrime, line 34: The task asked for tests.  console.log statements aren't tests!  I'm expecting some expect() code...

OracPrime: Adding invert.test.js for example...

OracPrime (approved):
Generally good work here.  A few minor suggestions, and a few questions for you to consider yourself.  But overall, good stuff - move on!

Sprint-2/debug/recipe.js
Amirmontazerii: @OracPrime Thank you for your suggestion! I'll definitely keep it in mind.

Sprint-2/implement/contains.test.js
Amirmontazerii: I updated the function and test so this way if someone tries to misuse the function by passing an array, a clear error message will be displayed, making it easier to debug

Sprint-2/implement/querystring.js
Amirmontazerii: @OracPrime You’re right that having the value sometimes as a string and sometimes as an array can be confusing. A simple fix would be to always store values as arrays, even if there’s only one value. That way, the function is consistent, and users won’t have to check the type before using it.

OracPrime:
A couple of good updates there - I think  you can move on now!  Good luck with the next task.

PR #327, opened 2025-03-05
ITP-JAN-LONDON |FOROGH AGHAEIYARIJANI |MODULE-DATA-GROUPS|SPRINT1|WEEK1|

Forogh-Aghaeiyarijani:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -3,9 +3,14 @@
 // If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  const sortedList = [...list].sort((a, b) => a - b);
+  const middleIndex = Math.floor(sortedList.length / 2);
+
+  if (sortedList.length % 2 === 0) {
+    return (sortedList[middleIndex - 1] + sortedList[middleIndex]) / 2;
+  } else {
+    return sortedList[middleIndex];
+  }
 }
 
 module.exports = calculateMedian;
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -3,6 +3,10 @@
 // If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
 
 function calculateMedian(list) {
+  if (list.length === 0) {
+    throw new Error("Cannot calculate median of an empty array");
+  }
+
   const sortedList = [...list].sort((a, b) => a - b);
   const middleIndex = Math.floor(sortedList.length / 2);
 
@@ -12,5 +16,4 @@ function calculateMedian(list) {
     return sortedList[middleIndex];
   }
 }
-
 module.exports = calculateMedian;
--- a/Sprint-1/implement/dedupe.test.js
+++ b/Sprint-1/implement/dedupe.test.js
@@ -1,4 +1,4 @@
-const dedupe = require("./dedupe.js");
+// const dedupe = require("./dedupe.js");
 /*
 Dedupe Array
 
@@ -16,7 +16,7 @@ E.g. dedupe([1, 2, 1]) target output: [1, 2]
 // Given an empty array
 // When passed to the dedupe function
 // Then it should return an empty array
-test.todo("given an empty array, it returns an empty array");
+// test.todo("given an empty array, it returns an empty array");
 
 // Given an array with no duplicates
 // When passed to the dedupe function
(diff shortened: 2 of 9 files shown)


Sprint-1/fix/median.js
```diff
@@ -3,9 +3,14 @@
 // If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  const sortedList = [...list].sort((a, b) => a - b);
```
illicitonion, line 6: This looks like it works, but just to check I understand why you've done it this way - what's the benefit of using array destructuring here? What would happen if we didn't use it? Is there a test we could write to show it's needed?

Sprint-1/fix/median.js
```diff
@@ -3,9 +3,14 @@
...
+  const middleIndex = Math.floor(sortedList.length / 2);
+
+  if (sortedList.length % 2 === 0) {
```
illicitonion, line 9: What would happen if we were passed an empty array? What answer would we return? What would we expect?

Sprint-1/implement/dedupe.test.js
```diff
@@ -25,3 +25,21 @@ test.todo("given an empty array, it returns an empty array");
 // Given an array with strings or numbers
 // When passed to the dedupe function
 // Then it should remove the duplicate values, preserving the first occurence of each element
+const dedupe = require("./dedupe.js");
```
illicitonion, line 28: It looks like you commented out an identical line on line 1 - why did you comment and re-add the same thing?

Sprint-1/implement/dedupe.test.js
```diff
@@ -25,3 +25,21 @@ test.todo("given an empty array, it returns an empty array");
...
+
+describe("dedupe", () => {
+  test("given an empty array, it returns an empty array", () => {
+    expect(dedupe([])).toEqual([]);
+  });
+
+  test("given an array with no duplicates, it returns a copy of the original array", () => {
```
illicitonion, line 35: You say in the test name "a copy", but I don't see the test testing that you actually copied the array, it could be from this test that the same array was returned as was passed in. How could you show this isn't happening in the test?

Sprint-1/implement/max.test.js
```diff
@@ -41,3 +41,34 @@ test.todo("given an empty array, returns -Infinity");
 // Given an array with only non-number values
 // When passed to the max function
 // Then it should return the least surprising value given how it behaves for all other inputs
+const findMax = require("./max.js");
+
+describe("findMax", () => {
```
illicitonion, line 46: This is a great set of tests - it seem to cover lots of edge-cases, good job!

illicitonion (commented):
This is generally looking really good - I've left a few questions about some generally quite small details :)

Sprint-1/fix/median.js
Forogh-Aghaeiyarijani: This creates a shallow copy of the input array using array destructuring ([...]). The benefit of doing this is that it prevents the original array from being modified when you call sort().
The sort() method would modify the original list in place because sort() mutates the array it’s called on. This could lead to unexpected side effects if other parts of the code rely on the original order of the input array.
this test:
  const list = [1, 2, 3];
  calculateMedian(list);

});
If you remove the array destructuring and let sort() mutate the original array, this test will fail because the original list will no longer be [1, 2, 3] after calling calculateMedian().

Sprint-1/fix/median.js
Forogh-Aghaeiyarijani: Return null or undefined to signal that the input is invalid.
Or throw an error (like new Error("Cannot calculate median of an empty array")).

Sprint-1/implement/dedupe.test.js
Forogh-Aghaeiyarijani: I temporarily commented it out while organizing the code to keep everything neat and aligned. Once I finished, I re-added it to make sure everything works correctly.

Sprint-1/implement/dedupe.test.js
Forogh-Aghaeiyarijani: To test that a new array (not the same reference) is returned, you can use toBe
I did it

illicitonion: Nice, this is really good!

Sprint-1/fix/median.js
illicitonion: Are you saying this is what you _expect_, or this is what _actually happens_?

Maybe you could write a test to make sure it's doing what you expect?

illicitonion (commented):
Good answers, thanks!

Sprint-1/fix/median.js
Forogh-Aghaeiyarijani: I did it.

illicitonion (commented):
Perfect! Thanks so much, and congratulations!

PR #328, opened 2025-03-05
ITP-JAN-25 | London | Saba Farjamfard | Module Data Groupe | Sprint 2 | Week 2

sabafarjamfard:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -33,34 +33,28 @@ test.todo("contains on empty object returns false");
 // Then it should return false or throw an error
 const contains = require("./contains.js");
 
-describe("contains function", () => {
-  test("returns false for an empty object", () => {
+describe("contains", () => {
+  test("contains on empty object returns false", () => {
     expect(contains({}, "a")).toBe(false);
   });
 
-  test("returns true if the object contains the key", () => {
-    const obj = { a: 1, b: 2 };
-    expect(contains(obj, "a")).toBe(true);
-    expect(contains(obj, "b")).toBe(true);
+  test("contains with existing property returns true", () => {
+    expect(contains({ a: 1, b: 2 }, "a")).toBe(true);
   });
 
-  test("returns false if the object does not contain the key", () => {
-    const obj = { a: 1, b: 2 };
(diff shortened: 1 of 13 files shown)


Sprint-2/implement/contains.test.js
```diff
@@ -33,3 +31,36 @@ test.todo("contains on empty object returns false");
...
+    const obj = { a: 1, b: 2 };
+    expect(contains(obj, "c")).toBe(false);
+  });
+
+  test("returns false for invalid inputs", () => {
+    expect(contains(null, "a")).toBe(false);
+    expect(contains(undefined, "a")).toBe(false);
+    expect(contains(42, "a")).toBe(false);
+    expect(contains("string", "a")).toBe(false);
+    expect(contains([], "a")).toBe(false);
```
cjyuan, line 57: > `expect(contains([], "a")).toBe(false);`

The function can also return `false` because `[]` does not contain the key "a"; such test may not properly check if the function can correctly recognise an array as an invalid parameter.

Can you think of better test values such that without this check, `Array.isArray(obj)`, the function would have returned `true`?

Sprint-2/implement/querystring.js
```diff
@@ -1,12 +1,15 @@
...
   if (queryString.length === 0) {
     return queryParams;
   }
+
   const keyValuePairs = queryString.split("&");
 
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
+    const [key, ...valueParts] = pair.split("=");
+    const value = valueParts.join("=");
```
cjyuan, line 12: For each of the following function calls, does your function return the value you expect?
```
parseQueryString("a=b&=&c=d")
parseQueryString("a=")
parseQueryString("=b")
parseQueryString("a=b&&c=d")
```

cjyuan (commented):
Code looks great. Just have two suggestions for you to consider.

Sprint-2/implement/contains.test.js
sabafarjamfard: Thanks I fixed

PR #329, opened 2025-03-05
Glasgow ITP-JAN-25 | Aiah Ibrahim | Module-Data-Groups | Week 9

aiahibrahim:
Self checklist

### DIFF
--- a/Sprint-2/debug/author.js
+++ b/Sprint-2/debug/author.js
@@ -1,5 +1,14 @@
 // Predict and explain first...
+/*
+some kind of error will occur because of the (for .. of) loop, it works in iterables such as array and strings not objects 
+since author is an object the code will not work as intended 
+when running the code type error occur
+for (const value of author) {
+                    ^
 
+TypeError: author is not iterable
+    at Object.<anonymous> 
+*/
 // This program attempts to log out all the property values in the object.
 // But it isn't working. Explain why first and then fix the problem
 
@@ -10,7 +19,12 @@ const author = {
   age: 40,
   alive: true,
 };
-
+/*
 for (const value of author) {
   console.log(value);
+}
+*/
+// using (for .. in) loop to iterate over the keys of the object 
+for (const value in author){
+  console.log(author[value]);
 }
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,3 +1,8 @@
-function contains() {}
+function contains(obj, property) {
+    if (typeof obj !== 'object' || obj === null || typeof property !== 'string'){
+        return false;
+    }
+    return property in obj;
+}
 
 module.exports = contains;
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -20,16 +20,28 @@ as the object doesn't contains a key of 'c'
 // Given an empty object
 // When passed to contains
 // Then it should return false
-test.todo("contains on empty object returns false");
+ test("given an empty object, it returns false", () => {
+    expect(contains({}, 'a')).toEqual(false);
+  });
 
 // Given an object with properties
 // When passed to contains with an existing property name
 // Then it should return true
+ // Test 1: Given an empty array, it returns an empty array
+ test("contains with an existing property return true", () => {
+    expect(contains({a: 1, b: 2},'a')).toEqual(true);
+  });
 
 // Given an object with properties
 // When passed to contains with a non-existent property name
 // Then it should return false
-
+test("contains with non-existent property return false", () => {
+    expect(contains({a: 1, b:2}, 'c')).toEqual(false);
+});
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+test("contains with invalid parameters (array) return false", () => {
+    expect(contains([] , 'a')).toEqual(false);
+});
+
--- a/Sprint-2/implement/querystring.js
+++ b/Sprint-2/implement/querystring.js
@@ -1,12 +1,21 @@
 function parseQueryString(queryString) {
   const queryParams = {};
+
+  // If the query string is empty, return an empty object
   if (queryString.length === 0) {
     return queryParams;
   }
+
+  // Split the query string into key-value pairs based on "&"
   const keyValuePairs = queryString.split("&");
 
+  // Loop through each key-value pair
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
+    // Split only at the first "=" to separate key and value, keeping the rest of the value intact
+    const [key, ...valueParts] = pair.split("=");
+    const value = valueParts.join("="); // Join the rest of the parts together to preserve the value
+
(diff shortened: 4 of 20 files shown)


cjyuan (commented):
Can you rebase the current branch onto `main`? (Currently it was created from your Sprint-1 branch). I will review your Sprint-2 exercise in a separate comment.

Here are the steps, and you may want to **backup all your files in the Sprint-2 folder** before carrying out these commands.

These instructions assume you had created a branch named `B2` from a branch named `B1` instead of from `main`, and you wanted to rebase `B2` from `B1` to `main`.
In your case, `B2` is the name of your Sprint-2 branch and `B1` is the name of your Sprint-1 branch.

#### 1. Open a terminal in VSCode

#### 2. Switch to the branch you want to rebase (`B2`)
`git switch B2`

Note: You can check which branch is the current branch via the command `git branch` (to list all branches with current branch highlighted)

#### 3. Rebase `B2` from `B1` onto `main`
`git rebase --onto main B1 B2`

For more details about this command, see
https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase#:~:text=interactive%20rebase%20display-,Advanced%20rebase%20application,-The%20command%20line

#### 4. Update (and Overwrite) your files in branch B2 at Github
While you are in branch B2 and you have verified that it has been successfully rebased, execute the following command:
`git push --force origin`

Sprint-2/debug/author.js
```diff
@@ -10,7 +19,12 @@ const author = {
...
-
+/*
 for (const value of author) {
   console.log(value);
 }
+*/
+// using (for .. in) loop to iterate over the keys of the object 
+for (const value in author){
+  console.log(author[value]);
+}
```
cjyuan, line 30: With this approach, it would be better to rename the variable `key`.

Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,8 @@
-function contains() {}
+function contains(obj, property) {
+    if (typeof obj !== 'object' || obj === null || typeof property !== 'string'){
```
cjyuan, line 2: Array is a kind of object in JS. Can you lookup how to check if an object is an array in JS?

Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,8 @@
...
+        return false;
+    }
+    return property in obj;
```
cjyuan, line 5: Consider the following two approaches for determining if an object contains a property:
```
 
```

Sprint-2/implement/contains.test.js
```diff
@@ -20,16 +20,28 @@ as the object doesn't contains a key of 'c'
...
 // Then it should return false
-
+test("contains with non-existent property return false", () => {
+    expect(contains({a: 1, b:2}, 'c')).toEqual(false);
+});
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+test("contains with invalid parameters (array) return false", () => {
+    expect(contains([] , 'a')).toEqual(false);
```
cjyuan, line 45: Array is a kind of object in JS, and "0", "1", "2" are keys of `[1, 2, 3]`. So it is better to express the test as `expect(contains([1, 2, 3], "1")).toBe(false);` to ensure the function can truly reject array.

You can also test other types of invalid parameters (e. g., number, undefined, null, string, boolean).

Sprint-2/implement/querystring.js
```diff
@@ -1,12 +1,21 @@
...
   const keyValuePairs = queryString.split("&");
 
+  // Loop through each key-value pair
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
+    // Split only at the first "=" to separate key and value, keeping the rest of the value intact
+    const [key, ...valueParts] = pair.split("=");
+    const value = valueParts.join("="); // Join the rest of the parts together to preserve the value
+
+    // Assign the key-value pair to the queryParams object
```
cjyuan, line 18: For each of the following function calls, does your function return the value you expect?
```
parseQueryString("a=b&=&c=d")
parseQueryString("a=")
parseQueryString("=b")
parseQueryString("a=b&&c=d")
```

Sprint-2/implement/tally.test.js
```diff
@@ -23,12 +23,29 @@ const tally = require("./tally.js");
...
+test("tally throws an error when input is not an array", () => {
+    expect(() => tally("string")).toThrow("Input must be an array");
+    expect(() => tally(123)).toThrow("Input must be an array");
+    expect(() => tally({})).toThrow("Input must be an array");
+});
+
+
+
+test("tally with mixed data type", () => {
+    expect(tally(['a','a',1,2,1,1])).toEqual({a: 2, 1: 3, 2:1});
```
cjyuan, line 50: What do you expect from the function call `tally(['1', 1, 1.0, "1.0"])`?

cjyuan (commented):
Code looks good. I just have a few suggestions.

Sprint-2/debug/author.js
aiahibrahim: I've changed it, thank you so much for your review and showing me how to rebase a branch as i always make this mistake and have to start over

Sprint-2/implement/contains.js
aiahibrahim: The best way to check if an object is an array is: 
Array.isArray(obj);
I've updated my function accordingly, thanks

Sprint-2/implement/contains.js
aiahibrahim: 1. The in operator checks for the property in the object and its prototype chain.
2. Object.hasOwn(obj, propertyName) only checks if the property exists directly on the object, not its prototype.
based on that Object.hasOwn() is the better approach for the contains function because it ensures that only direct properties are checked, preventing false positives from prototype properties like toString.

Sprint-2/implement/querystring.js
aiahibrahim: updated my function and test cases so it handles various edge cases correctly, including handling empty keys, missing values, and percent-encoded characters

Sprint-2/implement/tally.test.js
aiahibrahim: in JavaScript 1.0 is treated as the same value as 1, since 1.0 is a floating-point representation of the same number. But "1" (string) and 1 (number) are not equal because of their different types.
so the expected out come would be 
{
  '1': 1,    // The string "1" occurs once
  1: 2,      // The number 1 occurs three times (once as 1, once as 1.0)
  '1.0': 1   // The string "1.0" occurs once (if treated as a string)
}

aiahibrahim: can i change the label to complete ??

cjyuan: Did you check if your function returns what you expect? It can help you understand the characteristics of object's keys in JS.

cjyuan:
Changes look goos. Feel free to mark this as "Complete".

PR #331, opened 2025-03-06
Westmidlands/Millena-Mesfin/Module-Data-Groups/Sprint-1/Week#7

Millena28:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -3,9 +3,16 @@
 // If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
-}
+ 
+  const sortedList = [...list].sort((a,b) => a - b);
+  const middleIndex = Math.floor(sortedList.length / 2);
+  if (sortedList.length % 2 === 0){
+    return (sortedList[middleIndex - 1] + sortedList[middleIndex]) / 2;
+  }
+  else{
+    return sortedList[middleIndex];
+  }
+
+};
 
 module.exports = calculateMedian;
--- a/Sprint-1/implement/dedupe.test.js
+++ b/Sprint-1/implement/dedupe.test.js
@@ -16,12 +16,32 @@ E.g. dedupe([1, 2, 1]) target output: [1, 2]
 // Given an empty array
 // When passed to the dedupe function
 // Then it should return an empty array
-test.todo("given an empty array, it returns an empty array");
+//test.todo("given an empty array, it returns an empty array");
+
+test("given an empty array should return an empty array", () => {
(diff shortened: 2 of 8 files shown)


Sprint-1/fix/median.js
```diff
@@ -3,9 +3,16 @@
...
+  const sortedList = [...list].sort((a,b) => a - b);
+  const middleIndex = Math.floor(sortedList.length / 2);
+  if (sortedList.length % 2 === 0){
+    return (sortedList[middleIndex - 1] + sortedList[middleIndex]) / 2;
+  }
+  else{
+    return sortedList[middleIndex];
+  }
+
+};
```
cjyuan, line 16: Can you also improve the test code in `median.test.js` so that the code can properly test if the function can deal with arrays with unordered data?

Sprint-1/implement/dedupe.test.js
```diff
@@ -16,12 +16,32 @@ E.g. dedupe([1, 2, 1]) target output: [1, 2]
...
 // When passed to the dedupe function
 // Then it should return a copy of the original array
 
+test("An array with no duplicates should return a copy of the original" , () => {
+   
+    expect(dedupe(["H","E","Y"]).toEqual(["H","E","Y"]));
+    expect(dedupe(["r","a","e"]).toEqual(["r","a","e"]));
+
+});
+
```
cjyuan, line 36: The current test code cannot check if the returned array is a **copy** of the original array because `toEqual()` compares objects (including arrays) by value. To illustrate,

```
  const A = [2, 3, 1];
    
```

In order to check if the returned array is a **copy** of the original array, we would need **additional** checks. 
Can you find out what code you need to add in order to ensure the returned value is not the original array?

Sprint-1/implement/max.test.js
cjyuan, line 1: Have you tried running this test file? I don't think your function can pass the tests you specified.

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,44 @@ const sum = require("./sum.js");
...
+    expect (sum([-1,-2,-3,-4,-5])).toEqual(-15);
+});
 
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
+test("given an array with decimal numbers returns the total sum number" , () => {
+    expect (sum([0.5,2.5,3.1,6.6])).toEqual(12.7);
+});
+
```
cjyuan, line 41: Decimal numbers in most programming languages (including JS) are internally represented in "floating point number" format. Floating point arithmetic is not exact. For example, the result of `46.5678 - 46 === 0.5678` is false because `46.5678 - 46` only yield a value that is very close to `0.5678`. Even changing the order in which the program add/subtract numbers can yield different values.

So the following could happen
```

```

cjyuan (commented):
Code is generally good. `findMax()` could use some improvement.

PR #333, opened 2025-03-06
ITP-JAN-25 | London | Saba Farjamfard | Module-Data-Groups | Quote-generator| Sprint-3 | Week-3 

sabafarjamfard:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-3/quote-generator/quotes.js
+++ b/Sprint-3/quote-generator/quotes.js
@@ -491,3 +491,14 @@ const quotes = [
 ];
 
 // call pickFromArray with the quotes array to check you get a random quote
+function displayRandomQuote() {
+  const randomQuote = pickFromArray(quotes); // Select a random quote
+  document.getElementById("quote").innerText = `"${randomQuote.quote}"`; // Display the quote text
+  document.getElementById("author").innerText = `— ${randomQuote.author}`; // Display the author name
+}
+
+document
+  .getElementById("new-quote")
+  .addEventListener("click", displayRandomQuote);
+
+window.onload = displayRandomQuote;
--- a/Sprint-3/quote-generator/style.css
+++ b/Sprint-3/quote-generator/style.css
@@ -1,1 +1,56 @@
 /** Write your CSS in here **/
+body {
+  font-family: Arial, sans-serif;
+  background-color: #f5f5f5;
+  color: #333;
+  text-align: center;
+  padding: 20px;
+}
+
+.container {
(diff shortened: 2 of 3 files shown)


Sprint-3/quote-generator/quotes.js
```diff
@@ -491,3 +491,14 @@ const quotes = [
...
+  const randomQuote = pickFromArray(quotes); // Select a random quote
+  document.getElementById("quote").innerText = `"${randomQuote.quote}"`; // Display the quote text
+  document.getElementById("author").innerText = `— ${randomQuote.author}`; // Display the author name
+}
+
+document
+  .getElementById("new-quote")
+  .addEventListener("click", displayRandomQuote);
+
+window.onload = displayRandomQuote;
```
halilibrahimcelik, line 504: As a rule of thumb, always define your dom elements in the beginning of your code block
and please define them  something similar to this. 
`const quoateText= document.getElementById("quote")`
and then use these variables in your functions.
This is essential  to  write more clean codes.

Sprint-3/quote-generator/style.css
```diff
@@ -1 +1,56 @@
...
+  padding: 10px 20px;
+  font-size: 1rem;
+  cursor: pointer;
+  border-radius: 5px;
+  transition: background-color 0.3s ease;
+}
+
+button:hover {
+  background-color: #0056b3;
+}
```
halilibrahimcelik, line 56: It is nice to see that you gave importance on styling. You can also use nesting operator while using hover 
https://developer.mozilla.org/en-US/docs/Web/CSS/Nesting_selector you can check this page for more information

Sprint-3/quote-generator/quotes.js
halilibrahimcelik, line 504: Why don't we call displayRandomQuote() like this but use window.load=displayRandomQuote, can you tell me  what is benefit of using window.load ?

halilibrahimcelik (changes requested):
Good work, If you have time it would be really nice to see  stretch cases as well.

Sprint-3/quote-generator/quotes.js
sabafarjamfard: Hello, thank you for the review of my code 

window.onload = displayRandomQuote(); the function would run immediately instead of waiting for the page to load, which is not what we want.

PR #335, opened 2025-03-06
ITP-JAN-25 | LONDON | Bakytbek Sydykov | Module-Data-Groups | sprint 2

Bahasyd:
## Learners, PR Template

Self checklist

## Changelist

n|a

## Questions

### DIFF
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,3 +1,13 @@
-function contains() {}
+function contains(obj, key) {
+        if (typeof obj !== 'object' || obj === null) {
+    return false; 
+}
+return Object.prototype.hasOwnProperty.call(obj, key);
+    }
+    
+    console.log(contains({a: 1, b: 2}, 'a')); 
+    console.log(contains({a: 1, b: 2}, 'c')); 
+    console.log(contains({}, 'a')); 
+
 
 module.exports = contains;
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -16,20 +16,34 @@ as the object doesn't contains a key of 'c'
 // Given a contains function
 // When passed an object and a property name
 // Then it should return true if the object contains the property, false otherwise
-
+test("a contains function should return true if the object contains the property, false otherwise", () => {
+  expect(contains({a: 1},'a')).toEqual(true);
+  expect(contains({a: 1}, 'b')).toEqual(false);
+});
 // Given an empty object
 // When passed to contains
 // Then it should return false
-test.todo("contains on empty object returns false");
+test("contains on empty object returns false", () => {
+    expect(contains({}, 'a')).toEqual(false);
+});
 
 // Given an object with properties
 // When passed to contains with an existing property name
 // Then it should return true
-
+test("given an object with properties, contains should return true for an existing property", () => {
+    expect(contains({ a: 1 }, 'a')).toEqual(true);  
+});
 // Given an object with properties
 // When passed to contains with a non-existent property name
 // Then it should return false
-
+test("contains should return false for a non-existing property", () => {
+    expect(contains({ a: 1 }, 'b')).toEqual(false);
+});
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+test("contains should return false for invalid parameters", () => {
+    expect(contains([], 'a')).toEqual(false);     
+    expect(contains(null, 'a')).toEqual(false);   
+    expect(contains(undefined, 'a')).toEqual(false); 
+});
--- a/Sprint-2/implement/lookup.js
+++ b/Sprint-2/implement/lookup.js
@@ -1,5 +1,9 @@
-function createLookup() {
-  // implementation here
-}
-
+function createLookup(countryCurrencyPairs) {
+    return countryCurrencyPairs.reduce((lookup, [country, currency]) => {
+      lookup[country] = currency;
+      return lookup;
+    }, {});
+  }
+const countryCurrencyPairs = [['US', 'USD'], ['CA', 'CAD']];
+console.log(createLookup(countryCurrencyPairs));
 module.exports = createLookup;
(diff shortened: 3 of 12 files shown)


Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,13 @@
-function contains() {}
+function contains(obj, key) {
+        if (typeof obj !== 'object' || obj === null) {
```
cjyuan, line 2: In JS, an array is also an object. If we want to ensure `obj` is not an array, we would need an additional condition at line 2.

Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,13 @@
...
+    return false; 
+}
+return Object.prototype.hasOwnProperty.call(obj, key);
```
cjyuan, line 5: There is a shorter syntax to call the method `hasOwnProperty()` directly through the object `obj`. 
(Can  you figure out this syntax?)

We can also use `Object.hasOwn()` to accomplish the same objective.

Sprint-2/implement/contains.test.js
```diff
@@ -16,20 +16,34 @@ as the object doesn't contains a key of 'c'
...
 // Then it should return false
-
+test("contains should return false for a non-existing property", () => {
+    expect(contains({ a: 1 }, 'b')).toEqual(false);
+});
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+test("contains should return false for invalid parameters", () => {
+    expect(contains([], 'a')).toEqual(false);
```
cjyuan, line 46: If we don't reject array properly in the function `contains()`, `contains(['A', 'B', 'C'], '0')` would evaluate to `true`.
So that means the following test would fail.
```
  expect(contains(['A', 'B', 'C'], '0')).toEqual(false); 
```

Sprint-2/implement/lookup.test.js
```diff
@@ -1,7 +1,10 @@
 const createLookup = require("./lookup.js");
 
-test.todo("creates a country currency code lookup for multiple codes");
-
+test("creates a country currency code lookup for multiple codes", () => {
+const testPairs = [['US', 'USD'], ['CA', 'CAD']];
+console.log(createLookup(testPairs));
```
cjyuan, line 5: Seems like Jest test code is missing here.

Sprint-2/implement/querystring.js
```diff
@@ -1,16 +1,35 @@
...
+    let value = values.length > 0 ? values.join("=") : null;
+  
+    if (value === "") {
+      value = null;
+    } else if (value !== null) {
+      value = decodeURIComponent(value);
+    }
+
+    if (acc[key] !== undefined) {
+      acc[key] = [].concat(acc[key], value);
```
cjyuan, line 20: Nice touch!

Sprint-2/implement/querystring.js
```diff
@@ -1,16 +1,35 @@
...
   }
 
-  return queryParams;
+  return query.split("&").reduce((acc, pair) => {
+    let [key, ...values] = pair.split("=");
+    key = decodeURIComponent(key);
+    let value = values.length > 0 ? values.join("=") : null;
+  
+    if (value === "") {
+      value = null;
```
cjyuan, line 14: What if the value is an empty string?
Do you plan to distinguish the following two kinds of parameter?
`"A="` and `"A"`

Sprint-2/interpret/invert.js
```diff
@@ -15,15 +15,32 @@ function invert(obj) {
...
   return invertedObj;
 }
-
+*/
 // a) What is the current return value when invert is called with { a : 1 }
-
+//The function currently does not invert keys and values—it simply copies the input object.;
 // b) What is the current return value when invert is called with { a: 1, b: 2 }
-
+//invert({ a: 1, b: 2 }) the function does not swap keys and values; it just returns the original object.;
```
cjyuan, line 22: It is true that the function does not swap key with value properly, but the object it returns is not quite the same as the original object. Can you find out what it returns?

cjyuan (commented):
Code looks good. I only have a few suggestions.

Sprint-2/implement/lookup.test.js
Bahasyd: Ok. Yes, I missed. I fixed it.

Sprint-2/implement/contains.test.js
Bahasyd: I understood. Thank you.

Sprint-2/implement/contains.js
Bahasyd: I figured out since ES2022, Object.hasOwn(obj, key) is a better alternative.
   Thank you.

Sprint-2/implement/contains.js
Bahasyd: I added Array.isArray(obj).

Sprint-2/interpret/invert.js
Bahasyd: Okey.
The function does not return the original object { a: 1, b: 2 }, but rather a new object where the keys and values have been swapped as strings.
The keys in JavaScript objects are always strings, so 1 and 2 get converted to "1" and "2".

Sprint-2/implement/querystring.js
Bahasyd: "A=" is now { A: "" }, while "A" is { A: null }

Sprint-2/interpret/invert.js
cjyuan: The original code
```
    invertedObj.key = value;
  }
```
produces something else.

You can call the function with the original code to find out what it returns.

Sprint-2/implement/lookup.js
```diff
@@ -1,9 +1,14 @@
...
-      return lookup;
-    }, {});
-  }
 const countryCurrencyPairs = [['US', 'USD'], ['CA', 'CAD']];
+function createLookup(country) {
+  let object ={};
+  country.forEach(element => {
+    object[element[0]]= element[1];
+    
+  });
```
cjyuan, line 7: Why not try using **array destructuring** to improve the readability of the code?

cjyuan (commented):
Changes look good.

Feel free to mark this as "Completed" anytime.

PR #337, opened 2025-03-07
LONDON-JAN-25 | ANDREI FILIPPOV | Module-Data groups | WEEK 2

Droid-An:
## Learners, PR Template

Self checklist

## Changelist

done sprint 2 exercises

## Questions

no questions

### DIFF
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,3 +1,5 @@
-function contains() {}
+function contains(object, property) {
+  return Object.keys(object).includes(property);
+}
 
 module.exports = contains;
--- a/Sprint-2/implement/querystring.js
+++ b/Sprint-2/implement/querystring.js
@@ -6,7 +6,7 @@ function parseQueryString(queryString) {
   const keyValuePairs = queryString.split("&");
 
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
+    const [key, value] = pair.split(/=(.*)/s);
     queryParams[key] = value;
   }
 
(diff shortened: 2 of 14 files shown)


Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,5 @@
-function contains() {}
+function contains(object, property) {
+  return Object.keys(object).includes(property);
```
day-lee, line 2: This looks good! As a challenge, can you think of an Object method that checks if it contains the specified property?

Sprint-2/implement/querystring.js
day-lee, line 1: Using a regex as a split() argument was a good idea! 

In the real world, URLs are usually encoded with percent encoding to represent special characters. In URL, `&` can be used as a separator (key=value&key=value...) or special character within a key or value. if we were to use `&` as a special character, we need to encode it.

Can you refactor the code so that `parseQueryString("a%25b=c%26d")` results in `{ "a%b": "c&d" }`? 
FYI, `%25` is same as `%`, `%26` is same as `&`

Sprint-2/implement/contains.js
Droid-An: Sure! We can use the `Object.values` for that.

Sprint-2/implement/querystring.js
Droid-An: Thanks for the real-world example - I hadn’t heard about it before. I added the `decodeURIComponent()` method because it seems to be specifically used for decoding.

Sprint-2/implement/contains.js
day-lee: object.hasOwnProperty() can be used to check whether an object has a specific property! Your current code works well, and I wanted to suggest another way of checking. :)

day-lee:
Good job! 💯

PR #338, opened 2025-03-07
ITP_Jan-25| Sheetal_Kharab| Module_Data_group| Week_9

sheetalkharab:
## Learners, PR Template

Self checklist

## Changelist
Write test case and update function

## Questions

### DIFF
--- a/Sprint-2/debug/author.js
+++ b/Sprint-2/debug/author.js
@@ -1,5 +1,5 @@
 // Predict and explain first...
-
+// it will give error as it is not iterable object(author)
 // This program attempts to log out all the property values in the object.
 // But it isn't working. Explain why first and then fix the problem
 
@@ -11,6 +11,6 @@ const author = {
   alive: true,
 };
 
-for (const value of author) {
-  console.log(value);
+for (const value in author) {// use for ...in to iterate keys and access values
+  console.log(author[value]);
 }
--- a/Sprint-2/debug/author.js
+++ b/Sprint-2/debug/author.js
@@ -3,6 +3,9 @@
 // This program attempts to log out all the property values in the object.
 // But it isn't working. Explain why first and then fix the problem
 
+/* it will not work as objects in javascript are not iterable using a for...of loop.This loop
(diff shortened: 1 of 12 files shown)


lawkunchi (approved):

lawkunchi (changes requested):
Add a comment to explain why the previous solution was not working and explain your solution after rather

Sprint-2/debug/author.js
lawkunchi, line 1: Add a comment to explain why the previous solution was not working and explain your solution after rather

Sprint-2/implement/querystring.test.js
lawkunchi, line 1: An example test case as below will currently faile, check above comment.

test("parses querystring with encoded values", () => {
  expect(parseQueryString("name=John%20Doe&message=hello%3Dworld")).toEqual({
    name: "John Doe",
    message: "hello=world",
  });
});

lawkunchi (changes requested):
@sheetalkharab Added a few comments for improvement on a couple of tasks for you to check

Sprint-2/debug/author.js
sheetalkharab: Thanks for review, I just explained it.

Sprint-2/implement/querystring.js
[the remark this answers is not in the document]
sheetalkharab: What happens if a query string contains URL-encoded characters? eg. %20 for spaces.- will not decode and pass it John%20Doe will get.
How does the function behave when the key is missing (e.g., =value)? it makes empty key as a key.
Does the function handle empty values correctly (e.g., key1=&key2=value)? may be not store empty values correctly. Thank you.

Sprint-2/implement/querystring.test.js
sheetalkharab: Thanks ,Yes ,you are right. it's not checking it .I just implement logic for this using decodeUriComponent which will handle url encoded character to decode key and value. I used this test case and it passed.

sheetalkharab:
@lawkunchi, I commit the changes. please review them. Thanks

lawkunchi (approved):

PR #341, opened 2025-03-07
ITP-25-LONDON | MIKIYAS GEBREMICHAEL | DATA-GROUPS | alarmclockapp | SPRINT-3 | WEEK-3

Mikiyas-STP:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -1,106 +1,62 @@
+var audio = new Audio("alarmsound.mp3");
+var countDown;
+let minutesVal;
+let secVal;
 function setAlarm() {
+  if (countDown) {
+    clearInterval(countDown);
+  }
   var textareaVal = document.getElementById("alarmSet").value;
-  let separatedValues =  textareaVal.split(':');
-  let minutesVal = parseInt(separatedValues[0]);
-  let secVal = parseInt(separatedValues[1]);
-  //document.getElementById("timeRemaining").innerHTML = `Time Remaining: ${minutesVal} : ${secVal}`;
-  if (isNaN(minutesVal) || isNaN(secVal) || minutesVal > 59 || minutesVal < 0 || secVal > 59 || secVal < 0 ) {
-    document.getElementById("timeRemaining").innerHTML ="Input Invalid: Provide a string in MM:SS where MM and SS are numbers";
+  if(textareaVal>0 && textareaVal<=60){
+    minutesVal = 0;
+    secVal = parseInt(textareaVal);
+  }
+  else if(textareaVal>60){
+    minutesVal = Math.floor(textareaVal/60);
+    secVal = parseInt(textareaVal%60);
+
+  }
+  else if (
+    isNaN(minutesVal) ||
+    isNaN(secVal) ||
+    minutesVal > 59 ||
+    minutesVal < 0 ||
+    secVal > 59 ||
+    secVal < 0
+  ) {
+    //document.getElementById("timeRemaining").innerHTML ="Input Invalid: Provide a string in MM:SS where MM and SS are numbers";
+    alert("Input Invalid: Provide a Whole Number OR Provide a string in MM:SS where MM and SS are numbers");
     return;
   }
-  var countDown = setInterval(function () {
-      if (minutesVal !== 0) {
-        if (secVal !== 0) {
-          secVal--;
-          if (secVal === 0) {
-            minutesVal = minutesVal - 1;
-            secVal = 59;
-          }
-        }
-      } else if (minutesVal === 0) {
-        if (secVal !== 0) {
-          secVal--;
-        } else if (secVal === 0) {
-          clearInterval(countDown);
-          document.body.style.backgroundColor = "red";
-          playAlarm();
-        }
-      }
-      document.getElementById("timeRemaining").innerHTML = `Time Remaining: ${String(minutesVal).padStart(2,"0")} : ${String(secVal).padStart(2,"0")}`;
-    }, 1000);
+  else  {
+    let separatedValues = textareaVal.split(":");
+    minutesVal = parseInt(separatedValues[0]);
+    secVal = parseInt(separatedValues[1]);
   }
   
-
-/*if (secVal !== 0) {
-      if (minutesVal !== 0) {
+  countDown = setInterval(function () {
+    if (minutesVal !== 0) {
+      if (secVal !== 0) {
         secVal--;
         if (secVal === 0) {
           minutesVal = minutesVal - 1;
-          secVal = secVal + 60;
-          secVal--;
-        }
-        if (minutesVal === 0 && secVal === 0) {
-          playAlarm();
+          secVal = 59;
         }
       }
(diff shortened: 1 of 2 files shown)


cjyuan (commented):
Can you fix the issue?

Also, You missed updating `index.html` according to an instruction in `readme.md`.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,5 +1,113 @@
-function setAlarm() {}
+function setAlarm() {
+  var textareaVal = document.getElementById("alarmSet").value;
+  let separatedValues =  textareaVal.split(':');
+  let minutesVal = parseInt(separatedValues[0]);
+  let secVal = parseInt(separatedValues[1]);
+  //document.getElementById("timeRemaining").innerHTML = `Time Remaining: ${minutesVal} : ${secVal}`;
+  if (isNaN(minutesVal) || isNaN(secVal) || minutesVal > 59 || minutesVal < 0 || secVal > 59 || secVal < 0 ) {
+    document.getElementById("timeRemaining").innerHTML ="Input Invalid: Provide a string in MM:SS where MM and SS are numbers";
```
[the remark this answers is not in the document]
Mikiyas-STP: i have implemented another technique now. thanks

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,5 +1,113 @@
-function setAlarm() {}
+function setAlarm() {
+  var textareaVal = document.getElementById("alarmSet").value;
+  let separatedValues =  textareaVal.split(':');
+  let minutesVal = parseInt(separatedValues[0]);
+  let secVal = parseInt(separatedValues[1]);
```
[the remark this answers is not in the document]
Mikiyas-STP: yes your point does make sense thanks. i have also written code for single number it is very easy to remove the whole code i worked on and just use a single number input the reason i keep it is since i have learnt a lot when doing this project i want it to stay there to refer it for future. but of course i will refactor it in the future.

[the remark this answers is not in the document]
Mikiyas-STP: the reason i used mm:ss was because it was displayed in that format. it is not completely wrong way of doing it but it makes sense using a single number as well

Mikiyas-STP:
> Code works fine if a user only clicks the "Set Alarm" button once. However, if the user enters a time and then clicks the "Set Alarm" button multiple times, the countdown clock will not display properly. Can you fix the issue?
> 
> Also, You missed updating `index.html` according to an instruction in `readme.md`.

I have fixed this Issues Now if button is clicked before the alarm count finished it will restart from our new input like a new alarm. also html file is edited according to the readme file. thankyou for the comments

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,5 +1,113 @@
...
+      document.getElementById(
+        "timeRemaining"
+      ).innerHTML = `Time Remaining: ${minutesVal} : ${secVal}`;
+      
+    }
+
+  }
+
+
+*/
```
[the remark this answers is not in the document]
Mikiyas-STP: was for reference, now i deleted it.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,106 +1,62 @@
+var audio = new Audio("alarmsound.mp3");
+var countDown;
+let minutesVal;
+let secVal;
```
cjyuan, line 4: Why some variables are declared using `var` and some are declared using `let`?
`audio` can be declared using `const` if its value will not be reassigned.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,106 +1,62 @@
...
-    document.getElementById("timeRemaining").innerHTML ="Input Invalid: Provide a string in MM:SS where MM and SS are numbers";
+  if(textareaVal>0 && textareaVal<=60){
+    minutesVal = 0;
+    secVal = parseInt(textareaVal);
+  }
+  else if(textareaVal>60){
+    minutesVal = Math.floor(textareaVal/60);
+    secVal = parseInt(textareaVal%60);
+
+  }
```
cjyuan, line 18: Why treat 60 seconds as 0 minute 60 seconds but 120 seconds as 2 minutes 0 seconds?

I think you can possibly simplify these code fragment.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,106 +1,62 @@
...
+  else if (
+    isNaN(minutesVal) ||
+    isNaN(secVal) ||
+    minutesVal > 59 ||
+    minutesVal < 0 ||
+    secVal > 59 ||
+    secVal < 0
+  ) {
```
cjyuan, line 26: When the execution reaches this if-statement, `minutesVal` and `secVal` were not assigned any value in the current function call.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,106 +1,62 @@
...
-    else{
-      minutesVal--;
-      document.getElementById(
-        "timeRemaining"
-      ).innerHTML = `Time Remaining: ${minutesVal} : ${secVal}`;
-      
-    }
-
-  }
+    document.getElementById("timeRemaining").innerHTML = `Time Remaining: ${String(minutesVal).padStart(2,"0")} : ${String(secVal).padStart(2, "0")}`;
```
cjyuan, line 55: There is a 1 second delay before this statement get executed. So if a user set the countdown value to 30, the app would show 00:29 one second after the user clicks the "Set alarm" button (instead of showing 00:30 immediately).

cjyuan (commented):
Good job in keeping only one instance of alarm clock running at any given time!

I have few more suggestions for you to consider.

Sprint-3/alarmclock/alarmclock.js
Mikiyas-STP: Thanks, Just updated as you suggested. yes audio is a const element and using let is better than var since it is more updated version of js replacing var during variable declaration.

Sprint-3/alarmclock/alarmclock.js
Mikiyas-STP: Thanks I have simplified this condition.

Sprint-3/alarmclock/alarmclock.js
Mikiyas-STP: Thank You i fixed it it will only work if the user give the input in mm:ss version, keeping the basic functionality for a single number too.

Sprint-3/alarmclock/alarmclock.js
Mikiyas-STP: Noted !

Mikiyas-STP:
I have done the correction and commit my work here,
Thanks @cjyuan ,Please let me know if i have to close this repo

commit bbcc3ec07e86256efa94871e2f60c78e6be539e1

PR #347, opened 2025-03-08
ITPJ25LDN | Eyuel Abraham | Data Groups | Alarm-clock project sprint 3

eyuell21:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions


fernandamelov (approved):
Ok!

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,25 +1,93 @@
...
+// collecting values from the user
+const headingElement = document.querySelector("#timeRemaining")
+const stopBtn = document.querySelector("#stop")
+const setBtn = document.querySelector("#set")
+
+let alarmTimer = null
+let countdownInterval = null
+
+function setAlarm() {
+  let min = document.querySelector("#alarmSet").value
```
MorganDavid, line 10: It would be cleaner if we moved the querySelector to the top of the file with headingElement and stopBotton and then just get the value here.

Also, it's good practice to avoid abbreviations and use full words where possible. I'm not sure what `min` means here.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,25 +1,93 @@
...
+
+  if (min.trim() === "" || min <= 0) {
+    alert('Please set a valid time')
+    return;
+  }
+
+  min = Math.floor(parseInt(min))
```
MorganDavid, line 17: I don't think Math.floor does anything here since parseInt will always round to an integer already? 

You could do `Math.floor(Number(min))` instead if you want to make sure it's floored instead of rounded to nearest integer.

Sprint-3/alarmclock/alarmclock.test.js
```diff
@@ -1,104 +1,104 @@
...
-  input.value = "10";
-  button.click();
-
-  expect(mockPlayAlarm).toHaveBeenCalledTimes(0);
-
-  jest.runAllTimers();
-
-  expect(mockPlayAlarm).toHaveBeenCalledTimes(1);
-});
+/* ======= TESTS - DO NOT MODIFY =====
```
MorganDavid, line 1: I'm not sure why this file has ended up in the diff when there's no changes, that's a bit weird. but fine for now I guess.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,25 +1,93 @@
...
+    playAlarm()
+    changeBgColor()
+    setBtn.disabled = false
+  }, difference)
+}
+////////////////////////////////////////////////////////////////////////////////////////////////////////////
+function updateDisplay(seconds) {
+  let min = Math.floor(seconds / 60)
+  let remainingSeconds = seconds % 60
+  document.querySelector("#timeRemaining").innerText = `${min}:${remainingSeconds < 10 ? '0' + remainingSeconds : remainingSeconds}`
```
MorganDavid, line 36:
```suggestion
  document.querySelector("#timeRemaining").innerText = `${min}:${remainingSeconds.padStart(2, "0"}`

```

Javascript has padStart built in for doing this https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,25 +1,93 @@
...
+
+  updateCountdown(sec) // Start countdown
+
+  setBtn.disabled = true
+
+  alarmTimer = setTimeout(() => {
+    playAlarm()
+    changeBgColor()
+    setBtn.disabled = false
+  }, difference)
```
MorganDavid, line 30: Why do we need this if we have the if(seconds<=0) check in the setInterval? 

I suspect this can be done with only the setInterval?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,25 +1,93 @@
...
+    }
+  }, 1000)
+}
+//////////////////////////////////////////////////////////////////////////////////////////////////////
+function changeBgColor() {
+  let timeRemainingText = document.getElementById("timeRemaining")
+  timeRemainingText.style.backgroundColor = "#FF0000" // Change background color
+  const body = document.querySelector("body");
+  body.style.color = "white"
+}
```
MorganDavid, line 60: So, this changes the bg colour of the timeRemainingText to red, but then what changes it's text back to the original colour when the timer resets?

MorganDavid (commented):
Looks good! I left some small suggestions, I haven't tested this but I think there might be a bug where the background colour doesn't reset? (but I could be wrong)

Sprint-3/alarmclock/alarmclock.js
eyuell21: Yes that's my mistake, I have tried to correct it by adding a code that resets the text back to it's original colour on my recent commit. Thank You!

Sprint-3/alarmclock/alarmclock.js
eyuell21: Noted, Thankyou.

Sprint-3/alarmclock/alarmclock.js
eyuell21: I didn't know parseInt rounds to an integer. I have used Math.floor(Number(min)) on my recent commit. Thankyou!

Sprint-3/alarmclock/alarmclock.js
eyuell21: Noted, Thank You.

PR #349, opened 2025-03-08
London_ITP_Jan | Aung_Ye_Kyaw |Data-Group | Sprint1

sarawone:
>> Hi, I hope you are doing well!
>> First I would like to thank you for checking my PR and provide feedback on it.
>> I have implemented all the functions in Sprint 1 . 
>> Kindly check and provide feedback on it ....

### DIFF
--- a/Sprint-1/implement/dedupe.js
+++ b/Sprint-1/implement/dedupe.js
@@ -1,1 +1,26 @@
-function dedupe() {}
+function dedupe(List) {
+
+//if the array is empty return array
+if (List.length === 0)
+{
+    return List;
+}
+else
+{
+setArray = new Set(List);
+let List2 = [...setArray];
+return List2;
+}
+
+}
+
+test1 = dedupe([1,2,3,3,4,4,5]);
+console.log(test1);
+
+test2 = dedupe([]);
+console.log(test2);
+
+test3 = [1,2,3];
+console.log(test3);
+
+module.exports = dedupe;
--- a/Sprint-1/implement/dedupe.test.js
+++ b/Sprint-1/implement/dedupe.test.js
@@ -16,11 +16,24 @@ E.g. dedupe([1, 2, 1]) target output: [1, 2]
 // Given an empty array
 // When passed to the dedupe function
 // Then it should return an empty array
-test.todo("given an empty array, it returns an empty array");
+
+test ("get an empty array,return an empty array", () => {
+    empArr = dedupe([]);
+    expect(empArr.length).toEqual(0);
+
+})
+
 
 // Given an array with no duplicates
 // When passed to the dedupe function
 // Then it should return a copy of the original array
+
+/*test ("Given array with no duplicate, return copy of original array", () =>
+    {
+    expect(dedupe([1,2,3])).toEqual([1,2,3]);
+    });
+   */ 
+
 
 // Given an array with strings or numbers
 // When passed to the dedupe function
--- a/Sprint-1/implement/max.js
+++ b/Sprint-1/implement/max.js
@@ -1,33 +1,60 @@
 function findMax(elements) {
-// Given an empty array
-// When passed to the max function
-// Then it should return -Infinity
 
+// if the array length is 0 return infinity
 if (elements.length ===0)
 {
     return Infinity;
 }
-
-// Given an array with one number
-// When passed to the max function
-// Then it should return that number
-else 
+// if the array not include non-numeric number 
+else if (elements.every(item => isNaN(item)=== false))
 {
     maxNum = Math.max(...elements);
     return maxNum;
 }
+// if all the values in array have string 
+else if (elements.every(item => isNaN(item)=== true))
+{
+    return "This array have no numbers";
+}
+// if the array include non-numeric number
+else 
+{
+    numArr = elements.filter(item => typeof item === 'number');
+    return Math.max(...numArr);
+    
+}
 }
 
 module.exports = findMax;
-
+//empty array
 test1 = findMax([]);
 console.log(test1);
 
+//one numeric number array
 test2 = findMax([7]);
 console.log(test2);
 
+//array with numbers 
 test3 = findMax([2,3,7,6,8,9,]);
 console.log(test3);
 
+//array with mixing numeric and non-numeric
 test4 = findMax([1,"hello",3,4,"CYF"]);
 console.log(test4);
+
+//array with negative number 
+test5 = findMax([-1,-7,-8,-6]);
+console.log(test5);
+
+//array with mix negative number & positive number
+test6 = findMax([1,-7,-8,6]);
+console.log(test5);
+
+//array with decimal numbers
+test7 = findMax([0.1,3.5,7.7]);
+console.log(test7);
+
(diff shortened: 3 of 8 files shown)


Sprint-1/implement/dedupe.js
```diff
@@ -1 +1,26 @@
-function dedupe() {}
+function dedupe(List) {
```
MorganDavid, line 1: This is a nitpick, but usually variables are lowercase `list` in javascript

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,60 @@
...
+// if the array not include non-numeric number 
+else if (elements.every(item => isNaN(item)=== false))
+{
+    maxNum = Math.max(...elements);
+    return maxNum;
+}
+// if all the values in array have string 
+else if (elements.every(item => isNaN(item)=== true))
+{
+    return "This array have no numbers";
```
MorganDavid, line 17: No need to change this but just to let you know that you would usually throw an error here like `throw new Error("This array has no numbers` which would cause the program to crash.

Sprint-1/implement/max.test.js
```diff
@@ -16,28 +16,64 @@ const findMax = require("./max.js");
...
+    expect(DecArr).toEqual(2.7);
+
+})
+
+// Given an array with non-number values and numeric values
 // When passed to the max function
 // Then it should return the max and ignore non-numeric values
 
+test ("get an array with both non-num & num, get the max from num & ignore non-num", () =>{
+    MixArr = findMax([7,6,5,4,"CYF"]);
```
MorganDavid, line 68: again, variables should be defined with `const` and not start with a capital letter, but be in camelCase https://developer.mozilla.org/en-US/docs/Glossary/Camel_case

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,70 @@
...
+    else if (elements.every(items => isNaN(items)) === true)
+    {
+        return "There is no numbers in this array"
+    }
+
+    // if the array has mix with non-numeric & numeric value 
+    else 
+    {
+       // filter the string from the array 
+        Filter_arr = elements.filter(s => typeof s === 'number');
```
MorganDavid, line 29: a `NaN` is still a number in javascript (despite standing for "Not a Number") so you need to also check it's not NaN with `&& !isNaN(s)`

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,70 @@
 function sum(elements) {
```
MorganDavid, line 1: This function is good and it's interesting that you've used recursion to solve the last case. I think there is an easier solution, but that's a matter of opinion. No need to change but:

```ts
const arrWithOnlyNumbers = elements.filter(s => !isNaN(s));
// if arrWithOnlyNumbers.length === 0, return error message.
// put for loop you already have
```

Sprint-1/implement/dedupe.test.js
```diff
@@ -16,12 +16,25 @@ E.g. dedupe([1, 2, 1]) target output: [1, 2]
...
+/*test ("Given array with no duplicate, return copy of original array", () =>
+    {
+    expect(dedupe([1,2,3])).toEqual([1,2,3]);
+    });
+   */ 
+
+
 // Given an array with strings or numbers
 // When passed to the dedupe function
 // Then it should remove the duplicate values, preserving the first occurence of each element
```
MorganDavid, line 40: I think there's some missing test cases here?

MorganDavid (changes requested):
Looks good, there's some missing test cases which is the main thing that needs fixing.

Sprint-1/implement/max.test.js
MorganDavid: If you're using VSCode you can press F2 while selecting a variable to rename all occurances of it

MorganDavid:
```suggestion
    const mixedArray = findMax([7,6,5,4,"CYF"]);
```

Sprint-1/implement/dedupe.js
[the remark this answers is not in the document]
sarawone: change to lowercase.

Sprint-1/implement/dedupe.js
```diff
@@ -1 +1,26 @@
...
+//if the array is empty return array
+if (List.length === 0)
+{
+    return List;
+}
+else
+{
+setArray = new Set(List);
+let List2 = [...setArray];
+return List2;
```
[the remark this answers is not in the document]
sarawone: set is use to remove duplicate value . I put the array into new set so then it remove duplicate value .
I assign new array and change the set into array by using spread operator . 
I find out this method by surfing the web and get this method . Thank you.

Sprint-1/implement/max.test.js
sarawone: I changed all the variable to camelCase . Thank you .

Sprint-1/implement/dedupe.test.js
sarawone: I added the test case .. Kindly check .. Thank you so much .

Sprint-1/implement/max.js
sarawone: got it thank you so much

Sprint-1/implement/sum.js
sarawone: Thanks for the explanation .

Sprint-1/implement/dedupe.js
```diff
@@ -1 +1,26 @@
...
+function dedupe(List) {
+
+//if the array is empty return array
+if (List.length === 0)
+{
+    return List;
+}
+else
+{
+setArray = new Set(List);
```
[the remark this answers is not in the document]
sarawone: Added let .

Sprint-1/implement/dedupe.test.js
```diff
@@ -16,12 +16,29 @@ E.g. dedupe([1, 2, 1]) target output: [1, 2]
...
+})
+
 
 // Given an array with no duplicates
 // When passed to the dedupe function
 // Then it should return a copy of the original array
 
+test ("Given array with no duplicate, return copy of original array", () =>
+    {
+    expect(dedupe([1,2,3])).toEqual([1,2,3]);
```
illicitonion, line 33: This doesn't test that it returns a copy, just that the returned array contains the same values.

How could you test that it returns a _copy_?

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,49 @@ const sum = require("./sum.js");
 // Given an empty array
 // When passed to the sum function
 // Then it should return 0
-test.todo("given an empty array, returns 0")
+test ("Get an empty array, return 0", () =>{
+    EmptArr = sum([]);
```
illicitonion, line 17: What is the scope of this variable (and the other ones you make in this test file)? Is it as narrow a scope as it could be?

illicitonion (commented):
This is all looking really good - I left a couple of comments to think about :)

PR #351, opened 2025-03-08
LON | ITP JAN-25 | Amir Montazeri | Data groups | Sprint3-tasks | Alarmclock

Amirmontazerii:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -1,4 +1,29 @@
-function setAlarm() {}
+let countdown;
+
+function setAlarm() {
+  clearInterval(countdown);
+
+  let timeRemaining = document.getElementById("alarmSet").value;
+  timeRemaining = parseInt(timeRemaining);
+
+  if (isNaN(timeRemaining) || timeRemaining <= 0) return;
+
+  countdown = setInterval(function () {
+    document.getElementById("timeRemaining").innerText = `Time Remaining: 00:${
+      timeRemaining < 10 ? "0" + timeRemaining : timeRemaining
+    }`;
+
+    if (timeRemaining === 0) {
+      clearInterval(countdown);
+      playAlarm();
+    }
+    timeRemaining--;
+  }, 1000);
+}
+document.getElementById("stop").addEventListener("click", function () {
+  clearInterval(countdown);
+  pauseAlarm();
+});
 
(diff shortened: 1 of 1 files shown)


Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,29 @@
...
+  clearInterval(countdown);
+
+  let timeRemaining = document.getElementById("alarmSet").value;
+  timeRemaining = parseInt(timeRemaining);
+
+  if (isNaN(timeRemaining) || timeRemaining <= 0) return;
+
+  countdown = setInterval(function () {
+    document.getElementById("timeRemaining").innerText = `Time Remaining: 00:${
+      timeRemaining < 10 ? "0" + timeRemaining : timeRemaining
```
MorganDavid, line 13: What if I enter 120 in the time field? It will show `00:120` whereas we'd expect `2:00`. 

You'll need to look into the modulus operator to fix that.

MorganDavid: Can you check if all the tests pass? I think they should check for that.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,29 @@
...
+    }`;
+
+    if (timeRemaining === 0) {
+      clearInterval(countdown);
+      playAlarm();
+    }
```
MorganDavid, line 19: it can be better to do `timeRemaining <= 0`. That way, just in case you add a bug or something goes wrong in the evaluation of the setTimeout then the timer won't get stuck going into negative time. 

That's not a big deal though.

MorganDavid (changes requested):
I don't think this will work work for times more than 60 seconds. I'm not sure the tests pass in it's current state?

Sprint-3/alarmclock/alarmclock.js
Amirmontazerii: Thank you for review @MorganDavid , I've updated the code to correctly display the time in mm:ss format using Math.floor() for minutes and the modulus operator (%) for seconds.

Amirmontazerii: @MorganDavid thank you for review, you are right, I fixed it.

PR #361, opened 2025-03-10
LONDON-JAN-25 | ANDREI FILIPPOV | Module-Data groups | Alarm Clock

Droid-An:
## Learners, PR Template

Self checklist

## Changelist

done alarm clock app

## Questions

no questions

### DIFF
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -1,27 +1,29 @@
+const input = document.querySelector("input");
+let time = 0;
+let timerId;
+
 function setAlarm() {
-  const input = document.querySelector("input");
   time = input.valueAsNumber;
-  function changeDisplayTime(time) {
-    minutesRemaining = Math.floor(time / 60).toString();
-    paddedMinutesRemaining = minutesRemaining.padStart(2, "0");
-    secondsRemaining = (time % 60).toString();
-    paddedSecondsRemaining = secondsRemaining.padStart(2, "0");
-    const timeRemaining = `${paddedMinutesRemaining}:${paddedSecondsRemaining}`;
-    const charactersLeftP = document.querySelector("#timeRemaining");
-    charactersLeftP.innerText = `Time Remaining: ${timeRemaining}`;
-    
-  }
-  function deduction() {
-    changeDisplayTime(time)
-    if (time == 0) {
-      clearInterval(timerId)
-    }
-    --time
-  }
-  let timerId = setInterval(deduction, 1000)
+  timerId = setInterval(deduction, 1000);
 }
-function pauseAlarm() {
-  clearInterval(timerId)
+
+function deduction() {
+  changeDisplayTime(time);
+  if (time == 0) {
+    clearInterval(timerId);
+    playAlarm()
+  }
+  --time;
+}
+
+function changeDisplayTime(time) {
+  minutesRemaining = Math.floor(time / 60).toString();
+  paddedMinutesRemaining = minutesRemaining.padStart(2, "0");
+  secondsRemaining = (time % 60).toString();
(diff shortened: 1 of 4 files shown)


Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,32 @@
-function setAlarm() {}
+const input = document.querySelector("input");
+let time = 0;
```
illicitonion, line 2: I'm not sure from reading this what are the units of "time"?

Is it the time that the timer was started at? Or finishes at? Or the time remaining? It it seconds? Milliseconds? Days? Could you think of a variable name that could make this more clear?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,32 @@
...
+  changeDisplayTime(time);
+  if (time == 0) {
+    clearInterval(timerId);
+    playAlarm();
+    document.body.style.background = "#ff4d4d";
+  }
+}
+
+function changeDisplayTime(time) {
+  minutesRemaining = Math.floor(time / 60).toString();
```
illicitonion, line 22: What's the scope of this variable? Is the scope you've chosen appropriate? Is there a better scope you could choose?

Same question for the variables below in this function

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,32 @@
...
+  --time;
+  changeDisplayTime(time);
+  if (time == 0) {
+    clearInterval(timerId);
+    playAlarm();
+    document.body.style.background = "#ff4d4d";
+  }
+}
+
+function changeDisplayTime(time) {
```
illicitonion, line 21: Good job extracting a function here to reduce duplication between the initial "set alarm" and the updates every second!

illicitonion (commented):
This is generally looking good!

One thing I noticed is when I used the app, if I started the timer (say for 70 seconds), and then pressed "Stop alarm", it kept counting down. I would've expected pressing that button to cancel the timer completely. What do you think?

Sprint-3/alarmclock/alarmclock.js
Droid-An: The scope of this variable is limited to the changeDisplayTime function. I think this scope appropriate for it's purpose. I can’t think of a better scope for it.

Sprint-3/alarmclock/alarmclock.js
Droid-An: Thank you for the remark! I’ve changed the name to `secondsLeft`.

Sprint-3/alarmclock/alarmclock.js
illicitonion: Are you sure about that? What defines a variable's scope? How could you test that this variable doesn't have global scope?

Droid-An: A variable's scope is defined by where it is declared in the code. Do you mean that my problem is that I didn't type `let` or `const` before the variable name?

I forgot to declare the variable, but my code still worked, so I assumed everything was fine. However, after adding "use strict" at the top of the file, my code failed to run, and the console threw an "Uncaught ReferenceError."

I fixed it by adding `const` before every variable that didn’t have a clear declaration.

illicitonion: Right - in non-strict-mode in JS if you write this code:
```js
function f() {
  x = 1;
}

f();
console.log(x);
```

it will log `1`, because `x` ends up as a global variable instead of being scoped to the function. As you noted, `let` or `const` (or even `var`) will scope the variable down to the function.

day-lee:
FYI, This website is good for learning git - https://learngitbranching.js.org/

PR #370, opened 2025-03-12
ITP-Jan_25| Sheetal_Kharab| Module_Data_Group| Week_10

sheetalkharab:
## Learners, PR Template

Self checklist

## Changelist

Write logic for function and add tittle in Index file

## Questions

### DIFF
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -1,4 +1,35 @@
-function setAlarm() {}
+let countDown;
+let timeLeft;
+
+function setAlarm() {
+  let input= document.getElementById("alarmSet").value;
+  let timeRemaining = parseInt(input);
+  if(isNaN(timeRemaining)|| timeRemaining<0){
+    alert ("Please enter a positive number");
+    return;
+  }
+  timeLeft= timeRemaining; // to set left time 
+  updateTimeDisplay(timeLeft);// update time display
+
+  countDown = setInterval(()=>{
+    if(timeLeft<=0){
+      clearInterval(countDown);//when time is up it will clear countDown
+      playAlarm();
+    }else{
+      timeLeft--;
+      updateTimeDisplay(timeLeft);
+    }
+  }, 1000);// run countdown every 1 second
+}
+
+  function updateTimeDisplay(seconds){
+    let minute = Math.floor(seconds/60);
+    let remainingSeconds= seconds % 60;
+
+     document.getElementById("timeRemaining").innerText ="Time Remaining: " + String(minute).padStart(2,"0")
+     + ":" + String(remainingSeconds).padStart(2,"0");
+  
+ }
 
 // DO NOT EDIT BELOW HERE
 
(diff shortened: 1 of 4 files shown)


cjyuan (commented):

Also, the alarm does not sound as soon as the count down reaches zero; there is a delay of one second. Is it by design?

sheetalkharab:
@cjyuan ,Thankyou. I tried to resolve it .Now if we click multiple time on set alarm .it properly display time. I tried to remove one second delay as well. Please review it. Thankyou once again.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,39 @@
-function setAlarm() {}
+let countDown;
```
cjyuan, line 1: The name `countDown` might suggest that the variable stores some integer and that integer value is going to decrease.

Can you think of a more appropriate name for this variable so that when we see this name, we would have a good idea that it is going to be used to store the return value of `setInterval()`? (What value does this function return?)

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,39 @@
...
+    if(timeLeft<=0){
+      clearInterval(countDown);//when time is up it will clear countDown
+      playAlarm();
+      return;
+    }
+      
+      updateTimeDisplay(--timeLeft);
+    
+  }, 1000);// run countdown every 1 second
+}
```
cjyuan, line 27: With the current implementation, after the statement at line 24 displays 00:00, it will take one more second for the code at lines 18-21 to get executed. There is still a one second delay.

cjyuan (commented):
Your app now works properly even when "Set Alarm" button is clicked multiple times. Well done!

The alarm still only sounds 1 second after clock hits 00:00.

Sprint-3/alarmclock/alarmclock.js
sheetalkharab: @cjyuan , Thank you. I follow your suggestion and  change the name as you suggested. It return a unique Id  (+ve integer) for interval.

Sprint-3/alarmclock/alarmclock.js
sheetalkharab: @cjyuan , Thank you so much. I tried once again by changing logic little bit. Hopefully it will work now. Please review it once again. Thank you.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,41 @@
...
+
+  timerId = setInterval(() => {
+    if (timeLeft > 1) {
+      timeLeft--;
+      updateTimeDisplay(timeLeft);
+    } else {
+      clearInterval(timerId);
+      updateTimeDisplay(0); // ensure 00:00 is displayed
+      playAlarm();
+    }
```
cjyuan, line 27: This works. Another way of express this could be
```
timeLeft--; 
// update time display

if (timeLeft <= 0) {  // if timer reaches 0
  ...
}
```

cjyuan (commented):
Changes look good. You can mark this as "Complete".

PR #372, opened 2025-03-12
ITP_LONDON_JAN_2025| CRISTINA_MARIA_LEASCU | DATA_GROUPS | Sprint_2

CristinaLeascu:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-2/debug/recipe.js
+++ b/Sprint-2/debug/recipe.js
@@ -1,5 +1,9 @@
 // Predict and explain first...
 
+
+// Answer | It would not log out the ingredients, as the key and indexes for each ingredient were not called. 
+
+
 // This program should log out the title, how many it serves and the ingredients.
 // Each ingredient should be logged on a new line
 // How can you fix it?
@@ -12,4 +16,7 @@ const recipe = {
 
 console.log(`${recipe.title} serves ${recipe.serves}
   ingredients:
-${recipe}`);
+${recipe.ingredients[0]}
+${recipe.ingredients[1]}
+${recipe.ingredients[2]}
+${recipe.ingredients[3]}`);
--- a/Sprint-2/implement/querystring.js
+++ b/Sprint-2/implement/querystring.js
@@ -6,8 +6,9 @@ function parseQueryString(queryString) {
   const keyValuePairs = queryString.split("&");
 
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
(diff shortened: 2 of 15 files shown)


Sprint-2/debug/recipe.js
```diff
@@ -12,4 +16,7 @@ const recipe = {
 
 console.log(`${recipe.title} serves ${recipe.serves}
   ingredients:
-${recipe}`);
+${recipe.ingredients[0]}
```
day-lee, line 19: While this code is valid, when there are many more ingredients in an array, it can become cumbersome because it is hard-coded. Since arrays can be iterated, can you think of a way to make it easier?

Sprint-2/implement/querystring.js
```diff
@@ -6,8 +6,9 @@ function parseQueryString(queryString) {
   const keyValuePairs = queryString.split("&");
 
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
-    queryParams[key] = value;
+    const [key, ...valueParts] = pair.split("=");
+    const value = valueParts.join("="); // Rejoin the value in case it contains '='
+    queryParams[key] = decodeURIComponent(value || ""); // Handle undefined values and decode URI components
```
day-lee, line 11: What if the key also contains special characters, like in parseQueryString("a%25b=c%26d")? Can you refactor the code so it can handle special characters in the key as well? expeced output is `{ "a%b" : "c&d" }`

Sprint-2/debug/recipe.js
CristinaLeascu: Thnk you fopr your feedback, updated my code.

CristinaLeascu:
@day-lee 
I have implemented the required changes, looking forward to hearing from you and to mark this PR as completed to be able to submit to Module 3. Thank you.

Sprint-2/debug/recipe.js
```diff
@@ -10,6 +12,12 @@ const recipe = {
   ingredients: ["olive oil", "tomatoes", "salt", "pepper"],
 };
 
-console.log(`${recipe.title} serves ${recipe.serves}
-  ingredients:
-${recipe}`);
+for (const key in recipe) {
```
day-lee, line 15: This works, but you can directly access `recipe.ingredients`, instead of using nested for loop.

PR #375, opened 2025-03-13
ITPJ25LDN | Eyuel Abraham | Data Groups | Week 3 Sprint 3 To-do list

eyuell21:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

created a simple to-do list web app using javaScript.

## Questions

### DIFF
--- a/Sprint-3/todo-list/script.js
+++ b/Sprint-3/todo-list/script.js
@@ -1,25 +1,63 @@
-function populateTodoList(todos) {
-  let list = document.getElementById("todo-list");
-  // Write your code to create todo list elements with completed and delete buttons here, all todos should display inside the "todo-list" element.
+
+const userInput = document.querySelector('.toDoInput')
+const addToDobtn = document.querySelector('.addTask')
+const todoList = document.querySelector('.todoList')
+const isDone = document.getElementById('isDone')
+
+
+let todo_collection = []  //empty array to append our todo lists as an object
+
+function getTodo() {
+    
+    if (userInput.value && isDone) {
+        let taskLi = userInput.value
+        const taskObject = {  // creating an object of todo lists
+            nameOfTask: taskLi,
+            isCompleted: isDone.checked
+        }
+        todo_collection.push(taskObject)
+        isDone.checked = false
+
+        displayTask()
+    }
+    
+    else {
+        alert('Please enter both task and status.');
+    }
+
 }
 
-// These are the same todos that currently display in the HTML
-// You will want to remove the ones in the current HTML after you have created them using JavaScript
-let todos = [
-  { task: "Wash the dishes", completed: false },
-  { task: "Do the shopping", completed: false },
-];
+function displayTask() {
 
-populateTodoList(todos);
+    todoList.innerHTML = '';
 
-// This function will take the value of the input field and add it as a new todo to the bottom of the todo list. These new todos will need the completed and delete buttons adding like normal.
-function addNewTodo(event) {
-  // The code below prevents the page from refreshing when we click the 'Add Todo' button.
-  event.preventDefault();
-  // Write your code here... and remember to reset the input field to be blank after creating a todo!
+    todo_collection.forEach(task =>{
+        let taskLi = document.createElement("li") // creating a li element
+        let completedBtn = document.createElement('button')
+        completedBtn.textContent = 'completed'
+
+        let eraseBtn = document.createElement('button')
+        eraseBtn.textContent = "erase"
+
+        taskLi.textContent = `${task.nameOfTask} - ${task.isCompleted ? 'Completed' : 'Pending'}`;
+        todoList.appendChild(taskLi)
+
+        taskLi.appendChild(completedBtn)
+        taskLi.appendChild(eraseBtn)
+
+        completedBtn.addEventListener('click', function(){
+            taskLi.style.textDecoration = "line-through";
+        })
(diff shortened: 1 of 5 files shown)


Sprint-3/todo-list/script.js
```diff
@@ -1,25 +1,63 @@
-function populateTodoList(todos) {
-  let list = document.getElementById("todo-list");
-  // Write your code to create todo list elements with completed and delete buttons here, all todos should display inside the "todo-list" element.
+
+const userInput = document.querySelector('.toDoInput')
+const addToDobtn = document.querySelector('.addTask')
+const todoList = document.querySelector('.todoList')
+const isDone = document.getElementById('isDone')
```
cjyuan, line 5: The name `isDone` would suggest the variable stores a Boolean value. However, it is assigned a DOM element.

Sprint-3/todo-list/script.js
```diff
@@ -1,25 +1,63 @@
...
+
+
+let todo_collection = []  //empty array to append our todo lists as an object
+
+function getTodo() {
+    
+    if (userInput.value && isDone) {
```
cjyuan, line 12: Unless the DOM element that assigned to `isDone` does not exist, otherwise it will always has a truthy value. What's the meaning of including it in this condition?

Sprint-3/todo-list/script.js
```diff
@@ -1,25 +1,63 @@
...
+            isCompleted: isDone.checked
+        }
+        todo_collection.push(taskObject)
+        isDone.checked = false
+
+        displayTask()
+    }
+    
+    else {
+        alert('Please enter both task and status.');
```
cjyuan, line 25: This message is a bit confusing to me because it tells me I have to enter two input values.
However, the app would still work if I only entered just the task.

Sprint-3/todo-list/script.js
```diff
@@ -1,25 +1,63 @@
...
+
+        taskLi.textContent = `${task.nameOfTask} - ${task.isCompleted ? 'Completed' : 'Pending'}`;
+        todoList.appendChild(taskLi)
+
+        taskLi.appendChild(completedBtn)
+        taskLi.appendChild(eraseBtn)
+
+        completedBtn.addEventListener('click', function(){
+            taskLi.style.textDecoration = "line-through";
+        })
```
cjyuan, line 50: When the user clicks this button, the event listener (the function) should just update the state of the current task instead of updating its appearance. That is,
- The event listener should just toggle the task's `completed` property, and then
- Call `displayTask()` to update the view. The code to apply `line-through` based on the task's `completed` property should be written in `displayTask()`.

Sprint-3/todo-list/script.js
```diff
@@ -1,25 +1,63 @@
...
+        eraseBtn.addEventListener('click', function(){
+            todoList.removeChild(taskLi);
+        })
```
cjyuan, line 53: Similarly, this event listener should just update the application state by removing the task (an array element) from `todo_collection`, and the call `displayTask()` to update the view. 

With the current implementation, the task is not actually deleted. If we later add a new task, the supposedly "deleted task" will reappear in the todo list.

Sprint-3/todo-list/script.js
```diff
@@ -1,25 +1,63 @@
...
+function getTodo() {
+    
+    if (userInput.value && isDone) {
+        let taskLi = userInput.value
+        const taskObject = {  // creating an object of todo lists
+            nameOfTask: taskLi,
+            isCompleted: isDone.checked
+        }
+        todo_collection.push(taskObject)
+        isDone.checked = false
```
cjyuan, line 19: Should this statement be kept together with `userInput.value = "";` (line 56). It looks like their purpose is to "reset" the input fields.

Sprint-3/todo-list/script.js
```diff
@@ -1,25 +1,63 @@
...
+    userInput.value = "";
 }
 
-// Advanced challenge: Write a fucntion that checks the todos in the todo list and deletes the completed ones (we can check which ones are completed by seeing if they have the line-through styling applied or not).
-function deleteAllCompletedTodos() {
-  // Write your code here...
+
+function eraseTask() {
+    todo_collection = []
+    displayTask()
```
cjyuan, line 62: The challenge suggests deletion of all completed tasks (instead of all tasks).
Can you figure out how to do it?

Sprint-3/todo-list/script.js
```diff
@@ -1,25 +1,63 @@
...
+
+        let eraseBtn = document.createElement('button')
+        eraseBtn.textContent = "erase"
+
+        taskLi.textContent = `${task.nameOfTask} - ${task.isCompleted ? 'Completed' : 'Pending'}`;
+        todoList.appendChild(taskLi)
+
+        taskLi.appendChild(completedBtn)
+        taskLi.appendChild(eraseBtn)
+
```
cjyuan, line 47: The spec in `readm.md` mentions 
```
Each todo should have this HTML inside it:

<span class="badge bg-primary rounded-pill">
  <i class="fa fa-check" aria-hidden="true"></i>
  <i class="fa fa-trash" aria-hidden="true"></i>
</span>
```

Can you figure out how to construct and insert this HTML content instead of using `<button>`?

cjyuan (commented):
This branch contains also some files in the "slideshow" folder, but I only give reviews to the files in "todo-list".

PR #376, opened 2025-03-13
LONDON_JAN25 | KHALIL ALI | DATA_GROUP | SPRINT 2 

khalil-ali1:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-2/debug/author.js
+++ b/Sprint-2/debug/author.js
@@ -10,7 +10,7 @@ const author = {
   age: 40,
   alive: true,
 };
-
-for (const value of author) {
+//you have to add the object values to an array first, then log out the array
+for (const value of Object.values(author)) {
   console.log(value);
 }
--- a/Sprint-2/implement/querystring.test.js
+++ b/Sprint-2/implement/querystring.test.js
@@ -6,7 +6,12 @@
 const parseQueryString = require("./querystring.js")
 
 test("parses querystring values containing =", () => {
-  expect(parseQueryString("equation=x=y+1")).toEqual({
-    "equation": "x=y+1",
+  expect(parseQueryString("equation=x&x=y+1")).toEqual({
+    equation: "x",
+    x: "y+1"
   });
+});
+
+test("parses empty querystring =", () => {
+  expect(parseQueryString("")).toEqual({  });
 });
--- a/Sprint-2/implement/tally.js
+++ b/Sprint-2/implement/tally.js
@@ -1,3 +1,19 @@
-function tally() {}
+function tally(array) {
+if (Array.isArray(array) === false) {
+throw new Error("Invalid input");   
+}
+return array.reduce((acc, item) => {
+acc[item] = (acc[item] || 0) + 1;
+return acc;}, {});
 
+}
+/*
+{} is the initial value of the accumulator(acc).
+accumulator(acc) is an empty object. This object will store the item counts.
+item is the current element being processed from the array.
+acc[item] = (acc[item] || 0) + 1 :
+(acc[item] || 0) checks if acc[item] already exists. If it does, it uses its current value; otherwise, it defaults to 0.
++ 1 increments the count for the current item.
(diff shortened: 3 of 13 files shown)


day-lee:
You can get a quicker review if you put a label 'needs review' on your PR, as reviewers tend to check available PRs from the dashboard. The dashboard can only track PRs that have a label.

Sprint-2/implement/querystring.test.js
day-lee, line 1: querystring.js file is missing. You will need to fix the function querystring() and write a corresponding test. 
The query string format is `key=value`, but the current function doesn't work because the value contains extra = equal sign. Please refactor the function querystring() so `equation=x=y+1` parses to `"equation": "x=y+1"`

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,19 @@
-function tally() {}
+function tally(array) {
+if (Array.isArray(array) === false) {
```
day-lee, line 2: Can you add indentation inside the function to make it easier to read?

Sprint-2/interpret/invert.js
```diff
@@ -10,20 +10,21 @@ function invert(obj) {
...
+//the current return value is { key: 2 }
 // c) What is the target return value when invert is called with {a : 1, b: 2}
-
+//the target return value is { '1': 'a', '2': 'b' }
 // c) What does Object.entries return? Why is it needed in this program?
-
+//Returns an array [key, value].we used it to iterate over the object and get the key and value of the object
 // d) Explain why the current return value is different from the target output
-
+//The current return value is different from the target output because the key and value are not swapped
```
day-lee, line 28: There is another reason why it doesn't work as intended. Can you think of what it might be?

Sprint-2/debug/author.js
```diff
@@ -10,7 +10,7 @@ const author = {
   age: 40,
   alive: true,
 };
-
-for (const value of author) {
+//you have to add the object values to an array first, then log out the array
```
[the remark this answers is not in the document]
khalil-ali1: because the "for...of" loop is designed for iterable objects which are a collection of one value elements(like array, string, Map or Set) while objects are (key-value) pairs.
but what I used in this example that I made an array for the object values. then I logged out this array.

Sprint-2/implement/querystring.test.js
khalil-ali1: I fixed the code and added a test example.

Sprint-2/implement/tally.js
khalil-ali1: Done..
Thank you.

Sprint-2/interpret/invert.js
khalil-ali1: Because this line `invertedObj.key = value;`
assigns values to a literal key named "key".
To get the intended output we use `invertedObj[value] = key;` to swap between key and value.

day-lee: Correct!

Sprint-2/implement/querystring.test.js
day-lee: Good job! Below is some extra knowledge that might be useful for the future. It's totally up to you if you want to refactor the code or not.

In the real world, URLs are usually encoded with percent encoding to represent special characters. In a URL, & can be used as a separator (key=value&key=value...) or as a special character within a key or value. If we want to use & as a special character, we need to encode it.

For example, parseQueryString("a%25b=c%26d") results in { "a%b": "c&d" }. FYI, %25 is the same as %, and %26 is the same as &.

day-lee:
@khalil-ali1 Well done on refactoring the code. I've commented extra knowledge about the query string but you can change the label to complete anytime.

Sprint-2/implement/querystring.test.js
khalil-ali1: @day-lee 
Thank you for the useful information.

PR #382, opened 2025-03-14
LONDON_JAN25 | KHALIL ALI | MODULE_DATA_GROUPS | SPRINT 3 | ALARM-CLOCK

khalil-ali1:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -1,26 +1,46 @@
+let countdownTimeoutId = null;
+let isAlarmSet = false;
 function getFormattedInput(inputInSeconds) {
   const minutes = Math.floor(inputInSeconds / 60);
   const seconds = inputInSeconds % 60;
-  const formattedMinutes = String(minutes).padStart(2, '0');
-  const formattedSeconds = String(seconds).padStart(2, '0');
+  const formattedMinutes = String(minutes).padStart(2, "0");
+  const formattedSeconds = String(seconds).padStart(2, "0");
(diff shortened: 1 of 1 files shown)


Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,5 +1,28 @@
...
+function countdown(inputInSeconds) {
+  function updateAndContinue() {
+    if (inputInSeconds < 0) {
+      playAlarm();
+      document.body.style.backgroundColor = "red";
+      return;//break the loop when inputInSeconds reaches 0
+    }
+    timeRemaining.innerHTML = `Time Remaining: ${getFormattedInput(inputInSeconds)}`;
+    inputInSeconds--;
+    setTimeout(updateAndContinue, 1000);
```
cjyuan, line 17: You can also consider using `setInterval()` to schedule a task to run every 1000 milliseconds.

With your current approach, the would happen every (1000 + X) milliseconds, where X is the time taken to execute the code in lines 10-17.

cjyuan (commented):
Can you fix the issue? (One possible solution is to use a global variable to store the return value of `setTimeout()` or `setInterval()`, and then use `clearInterval()` to terminate the "previous countdown")

Sprint-3/alarmclock/alarmclock.js
khalil-ali1: I solved the problem when the user clicks set alarm.
if the alarm is already set an alert will show to tell the user that alarm is already set.

cjyuan:
Changes look good.

PR #384, opened 2025-03-15
ITP_LONDON_JAN_2025| CRISTINA_MARIA_LEASCU | DATA_GROUPS | Alarmclock

CristinaLeascu:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -2,15 +2,61 @@
 function formatTime(seconds) {
   const minutes = Math.floor(seconds / 60);
   const remainingSeconds = seconds % 60;
+  return `${String(minutes).padStart(2, "0")}:${String(
+    remainingSeconds
+  ).padStart(2, "0")}`;
+}
 
-  const formattedMinutes = String(minutes).padStart(2, "0");
-  const formattedRemainingSeconds = String(remainingSeconds).padStart(2, "0");
+// function to reset the background and stop any background flashing
+function resetBackground() {
+  document.body.style.backgroundColor = "white"; // set background to white
+  if (window.flashInterval) {
+    clearInterval(window.flashInterval);
+  }
+}
 
-  console.log(
-    "formatted time (mm:ss):",
-    `${formattedMinutes}:${formattedRemainingSeconds}`
-  );
-  return `${formattedMinutes}:${formattedRemainingSeconds}`;
+// function to handle countdown clock
+let countdownInterval = null;
+
+function countdownClock(timeInSeconds) {
+  const timeRemaining = document.querySelector("#timeRemaining");
+
+  if (countdownInterval) {
+    clearInterval(countdownInterval);
+  }
+
+  countdownInterval = setInterval(() => {
+    timeInSeconds--;
+
+    // update the display each second
+    timeRemaining.textContent = `Time Remaining: ${formatTime(timeInSeconds)}`;
+
+    // When time reaches 00:00, stop the countdown and play the alarm
+    if (timeInSeconds <= 0) {
+      clearInterval(countdownInterval);
+      flashBackground();
+    }
+  }, 1000);
+}
+
+// function for background flash
+function flashBackground() {
+  let flashCount = 0; //
+  const colors = ["pink", "plum", "indigo", "lavenderblush"];
+
+  window.flashInterval = setInterval(() => {
+    document.body.style.backgroundColor = colors[flashCount % colors.length];
+    flashCount++;
+
+    if (flashCount >= colors.length) {
+      clearInterval(window.flashInterval);
+      document.body.style.backgroundColor = "white"; // set the final background color to white
+
+      setTimeout(() => {
+        playAlarm(); // start the alart af the end of backgroung flash
+      });
+    }
+  }, 500); // flash every 500ms
 }
 
 function setAlarm() {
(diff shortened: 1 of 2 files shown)


Sprint-3/alarmclock/alarmclock.js
```diff
@@ -23,3 +80,4 @@ function pauseAlarm() {
 }
 
 window.onload = setup;
+module.exports = setAlarm;
```
cjyuan, line 83: This statement would cause a runtime error when the app is running in a browser.

When you test your app in the browser, you can check the JS console in the browser to see if there is any runtime error message.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,61 @@
...
+  const minutes = Math.floor(seconds / 60);
+  const remainingSeconds = seconds % 60;
+
+  const formattedMinutes = String(minutes).padStart(2, "0");
+  const formattedRemainingSeconds = String(remainingSeconds).padStart(2, "0");
+
+  console.log(
+    "formatted time (mm:ss):",
+    `${formattedMinutes}:${formattedRemainingSeconds}`
+  );
```
cjyuan, line 12: We should remove or comment out debugging statements in the production code.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,61 @@
...
+
+        if (flashCount >= colors.length) {
+          // Stop flashing after 3 flashes
+          clearInterval(flashInterval); // Stop flashing
+          document.body.style.backgroundColor = "pink"; // Keep background red
+          playAlarm();
+        }
+      }, 500);
+    }
+  }, 1000);
```
cjyuan, line 57: There isn't seem to be a way to reset the changed background to its initial state (white background).

Restarting a new countdown immediately after the countdown reaches 0 does not seem to be able to stop the blinking. I expect restarting a countdown would reset the appearance.

If you have time, you can consider using only CSS to animate the background and then use JS to toggle a CSS class to add/remove the effect.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,61 @@
...
+
+  const formattedTime = formatTime(timeInSeconds); // Use the formatTime function
+  timeRemaining.innerText = `Time Remaining: ${formattedTime}`;
+
+  // Countdown logic to update time and change background color at 00:00
+  const countdown = setInterval(() => {
+    timeInSeconds--;
+
+    // Update the display each second
+    timeRemaining.textContent = `Time Remaining: ${formatTime(timeInSeconds)}`;
```
cjyuan, line 36: If we place the statement that displays the countdown clock (i.e., the code that update the view) in a function,
- we can replace the lines 28-29 by a function call, and
- in case we need to change the wordings or the way to display the time, we only need to change one set of code

cjyuan (commented):

Also, the background color, once changed, cannot be reset to its initial color.

Can you fix the issues?

Sprint-3/alarmclock/alarmclock.js
CristinaLeascu: Thank you for your feedback, log removed,

Sprint-3/alarmclock/alarmclock.js
CristinaLeascu: Thank you. I have implemented a function.

CristinaLeascu:
@cjyuan  Thank you for your feedback.
I have removed log, created functions for countdownClock, resetBackground, flashBackground and updated my code.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -2,63 +2,95 @@
...
+}
+
+// function for background flash
+function flashBackground() {
+  let flashCount = 0; //
+  const colors = ["pink", "plum", "indigo", "lavenderblush"];
 
-  const formattedMinutes = String(minutes).padStart(2, "0");
-  const formattedRemainingSeconds = String(remainingSeconds).padStart(2, "0");
+  window.flashInterval = setInterval(() => {
```
cjyuan, line 47: Why introduce `flashInterval` as a property of `window` object (which has the same effect as declaring it in the global scope using 'var') but declare `countdownInterval` at line 19 using `let`?

Both approaches work. I am just curious why you used two different approaches.

You can ask ChatGPT the difference between declaring global variables using `var` and `let`.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -2,63 +2,95 @@
...
+
+// function to handle countdown clock
+let countdownInterval = null;
+
+function countdownClock(timeInSeconds) {
+  const timeRemaining = document.querySelector("#timeRemaining");
+
+  if (countdownInterval) {
+    clearInterval(countdownInterval);
+  }
```
cjyuan, line 26: The if-statement is not needed. Can you figure out why we can just call `clearInterval(countdownInterval);` without the if-statement?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -2,63 +2,95 @@
...
-    `${formattedMinutes}:${formattedRemainingSeconds}`
-  );
-  return `${formattedMinutes}:${formattedRemainingSeconds}`;
+    if (flashCount >= colors.length) {
+      clearInterval(window.flashInterval);
+      document.body.style.backgroundColor = "white"; // set the final background color to white
+
+      setTimeout(() => {
+        playAlarm(); // start the alart af the end of backgroung flash
+      });
```
cjyuan, line 57: What's the difference between
```
      setTimeout(() => {
        playAlarm();
      });
```
and
```
        playAlarm();
```

Why not just use the latter approach?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -2,63 +2,95 @@
...
+    flashCount++;
 
-  console.log(
-    "formatted time (mm:ss):",
-    `${formattedMinutes}:${formattedRemainingSeconds}`
-  );
-  return `${formattedMinutes}:${formattedRemainingSeconds}`;
+    if (flashCount >= colors.length) {
+      clearInterval(window.flashInterval);
+      document.body.style.backgroundColor = "white"; // set the final background color to white
```
cjyuan, line 53: Why not just call `resetBackground()`?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -2,63 +2,95 @@
...
+// function to reset the background and stop any background flashing
+function resetBackground() {
+  document.body.style.backgroundColor = "white"; // set background to white
+  if (window.flashInterval) {
+    clearInterval(window.flashInterval);
+  }
+}
+
+// function to handle countdown clock
+let countdownInterval = null;
```
cjyuan, line 19: Common best practice is to declare global variables at the top of the files; it makes it easier to see and manage them.

CristinaLeascu: Thank yiu fir yiur feedback, after reaserching I have understood that I can use clearInterval(countdownInterval);, as clearInterval it handles invalid interval gracefuly. I will stop the interval or ingnore the call.

Sprint-3/alarmclock/alarmclock.js
CristinaLeascu: Thank you for your feedback, after reaserching I have understood that I can use clearInterval(countdownInterval);, as clearInterval it handles invalid interval gracefuly. I will stop the interval or ingnore the call.

Sprint-3/alarmclock/alarmclock.js
CristinaLeascu: flashInterval controls the flashing of the background, would have to be accesible for other functions globally accesible. 
countdownInterval it was intended to be used for countdownClock only and the I have added to the setAlarm(). I understood now the diffrence. Thank you.

Sprint-3/alarmclock/alarmclock.js
CristinaLeascu: I has some issues with the alarm starting when countdown reached 00:00 
and not after the background flash, I introduced 2000 delay to give time for flashing then removed and I left it as it worked.

PR #388, opened 2025-03-15
ITP-Jan 2025 | London | Chi Mbah | Module-Data-Groups | Week 7 | Sprint 1

Chi-Mb:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -7,18 +7,16 @@ function calculateMedian(list) {
   if (list.length % 2 == 1){
     // odd length array
     const middleIndex = Math.floor(list.length / 2);
-    const median = list.splice(middleIndex, 1)[0];
+    const median = list[middleIndex];
     return median;
   } else {
     // even length array
     const middleIndex = Math.floor(list.length / 2) - 1;
-    let middle = list.splice(middleIndex, 2);
+    let middle = [list[middleIndex], list[middleIndex + 1]];
     let median = (middle[0] + middle[1]) / 2
-    return median
+    return median;
   }
 
-
-  
 }
 
 module.exports = calculateMedian;
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -2,10 +2,23 @@
 // Start by running the tests for this function
 // If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
 
+[1,2,3,4]
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  if (list.length % 2 == 1){
+    // odd length array
+    const middleIndex = Math.floor(list.length / 2);
+    const median = list.splice(middleIndex, 1)[0];
+    return median;
+  } else {
+    // even length array
+    const middleIndex = Math.floor(list.length / 2) - 1;
+    let middle = list.splice(middleIndex, 2);
+    let median = (middle[0] + middle[1]) / 2
+    return median
+  }
+
+
+  
 }
 
 module.exports = calculateMedian;
--- a/Sprint-1/fix/median.test.js
+++ b/Sprint-1/fix/median.test.js
@@ -19,8 +19,9 @@ describe("calculateMedian", () => {
 
   test("doesn't modify the input", () => {
     const list = [1, 2, 3];
+    const copy = list
     calculateMedian(list);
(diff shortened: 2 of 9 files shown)


Sprint-1/fix/median.js
```diff
@@ -2,10 +2,23 @@
...
 
+[1,2,3,4]
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  if (list.length % 2 == 1){
+    // odd length array
+    const middleIndex = Math.floor(list.length / 2);
+    const median = list.splice(middleIndex, 1)[0];
```
cjyuan, line 10: There is a shorter syntax and more efficient approach to access an array element.

Sprint-1/fix/median.js
```diff
@@ -2,10 +2,23 @@
 // Start by running the tests for this function
 // If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
 
+[1,2,3,4]
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  if (list.length % 2 == 1){
```
cjyuan, line 7: The arrays specified in `median.test.js` are already sorted by coincidence.

Typically, to find the median value of an array, we first need to sort it. However, to prevent modifying `list` directly, it's best to work with a cloned version instead.

Could you add one or more statements to create a cloned copy of `list` and then sort the numbers in the cloned array?

Sprint-1/fix/median.test.js
```diff
@@ -19,8 +19,9 @@ describe("calculateMedian", () => {
 
   test("doesn't modify the input", () => {
     const list = [1, 2, 3];
+    const copy = list
     calculateMedian(list);
 
-    expect(list).toEqual([1, 2, 3]);
+    expect(list).toEqual(copy);
```
cjyuan, line 25: The elements in `[1, 2, 3]` are already sorted. If we use this array as a test value, it will be difficult to determine whether the function modifies the input or not (by sorting the given array).

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,41 @@ const sum = require("./sum.js");
...
+test("given an array containing negative numbers, returns the correct sum", () => {
+    expect(sum([-1, -3, 5])).toEqual(1); 
+});
 
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
+test("given an array with decimal numbers, returns the correct sum", () => {
+    expect(sum([1.2, 3.2, 5.6])).toEqual(10); 
+});
```

So the following could happen
```

```

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,9 @@
 function sum(elements) {
-}
+    const sum = (elements) => {
+        const numericalValues = elements.filter(element => typeof element === 'number');
+        return numericalValues.reduce((total, current) => total + current, 0);
+          }
+    
+    }
```
cjyuan, line 7: What do you expect from the following function call?

```
sum([NaN, 1]);
```

cjyuan (commented):
Code is good. Some of the functions and Jest tests can be improved.

Sprint-1/fix/median.test.js
Chi-Mb: @cjyuan Can I use it in reverse, like [3, 2, 1], to make it easier to tell if the array gets modified?

Sprint-1/fix/median.js
Chi-Mb: @cjyuan Yes. Instead of using splice() to remove the middle element, I can just access it directly with array indexing

Sprint-1/fix/median.test.js
cjyuan: Some people might sort the array in reversed order.

Sprint-1/implement/sum.js
Chi-Mb: @cjyuan I updated my function and added a filter that removes NaN values.

Sprint-1/implement/sum.test.js
Chi-Mb: @cjyuan ChatGPT said:
I used Math.abs(result - expected) to find the difference and toBeLessThan(tolerance) to check if it's within the allowed tolerance.

PR #390, opened 2025-03-15
NOT-A-STUDENT-PR: Update Solution and tests for Sprint 1 'fix median' task

tyzia:
This PR suggests a few improvements in the proposed solution for `calculateMedian()` method.

The solution example now will have two advanced features:

1. Throws an error for invalid input (non-numeric array).
2. Filters array and calculates median for mixed arrays.

It also adds a very important step in calculating a median - sorts the array.

The PR also updates unit tests. If we don't think that student's implementation should handle non-numeric arrays or throw errors, I can remove tests related to these cases.

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -11,18 +11,27 @@
 // }
 
 // Explanation:
-// .splice() is a mutating method, so it updates the array in place.
-// to prevent this, you could take a copy of the array and use .splice(),
-// or you could use bracket notation to element at the desired index
+// The original implementation has several mistakes:
+// 1 - .splice() mutates or changes the original array. We don't want this to happen.
+// 2 - We actually don't need .splice() or any other method to find a median.
+// 3 - We must sort the array before we search for the median.
+// 4 - We should check if the 'list' parameter actually has numbers, otherwise try to filter
+// out all non-numeric values or throw an error if there are no numeric values.
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
+  const onlyNumberList = list.filter(item => typeof item === 'number' && !isNaN(item));
 
-  if (list.length % 2 === 0) { // use the remainder operator (%) to check if the array length is even
-    return (list[middleIndex - 1] + list[middleIndex]) / 2;
+  if (!onlyNumberList.length) {
+    throw new Error('You provided a list without numbers.');
+  } else {
+    const sortedList = onlyNumberList.sort((current, next) => current - next);
+    const indexNearMiddleOfArray = Math.floor(sortedList.length / 2);
+    if (sortedList.length % 2 === 0) {
+      return (sortedList[indexNearMiddleOfArray - 1] + sortedList[indexNearMiddleOfArray]) / 2;
+    } else {
+      return sortedList[indexNearMiddleOfArray];
+    }
   }
-
-  return list[middleIndex];
 }
 
 module.exports = calculateMedian;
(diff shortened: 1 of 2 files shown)


Sprint-1/fix/median.test.js
```diff
@@ -7,20 +7,42 @@
...
 
-  it("doesn't modify the input", () => {
+  it("doesn't modify the input array [1, 2, 3]", () => {
     const list = [1, 2, 3];
     calculateMedian(list);
-
     expect(list).toEqual([1, 2, 3]);
   });
+
+  [ 'not an array', 123, null, undefined, {}, [], ["apple", null, undefined] ].forEach(val =>
```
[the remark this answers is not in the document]
tyzia: Absolutely, the `main` should also be updated. I will first update the tests as to my comment above - not to expect `throw` and after we merge these changes into `solutions`, I will create similar change to `main`.

Sprint-1/fix/median.js
```diff
@@ -11,18 +11,27 @@
...
+// out all non-numeric values or throw an error if there are no numeric values.
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
+  const onlyNumberList = list.filter(item => typeof item === 'number' && !isNaN(item));
 
-  if (list.length % 2 === 0) { // use the remainder operator (%) to check if the array length is even
-    return (list[middleIndex - 1] + list[middleIndex]) / 2;
+  if (!onlyNumberList.length) {
+    throw new Error('You provided a list without numbers.');
```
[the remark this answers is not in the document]
tyzia: Hey @illicitonion please take a look at updated version:

- removed throw error
- return null in case not an array or array with non-numeric values provided

I also keep a commit with both versions of this function (returning null and trowing an error) and related unit tests just in case we would like to add `throw` to learning materials and this example will be needed: a6fd7064504d9f4856c9116ce247d0ce41d879da

Once this PR is approved and merged, I will create the same tests for `main` branch.

Sprint-1/fix/median.js
```diff
@@ -11,18 +11,33 @@
...
+  if (!Array.isArray(list)) {
+    return null;
   }
+  const onlyNumberList = list.filter(item => typeof item === 'number' && !isNaN(item));
 
-  return list[middleIndex];
+  if (!onlyNumberList.length) {
+    return null;
+  } else {
+    const sortedList = onlyNumberList.sort((current, next) => current - next);
```
illicitonion, line 33: The default comparator should work for numbers, right?

```suggestion
    const sortedList = onlyNumberList.sort();
```

illicitonion (approved):
LGTM, thanks!

Sprint-1/fix/median.js
tyzia: Not really, based on what I know about sort() if no comparator function is provided, it will convert numbers to strings and sort them alphabetically. Which will produce `wrong` results.

`console.log([10, 2, 1, 5].sort()); // [1, 10, 2, 5]`

tyzia:
Hi @illicitonion Do I need to get any more approvals or you can merge it?

Sprint-1/fix/median.js
illicitonion: Oh wow, I knew the default comparator wasn't great, but I didn't realise it was _that_ not great!!

illicitonion:
Merged! Thank you! If you can back-port the new/updated tests to `main` that'd be handy too!

tyzia:
Created similar change for `main` branch. @illicitonion please review: https://github.com/CodeYourFuture/Module-Data-Groups/pull/452

PR #393, opened 2025-03-15
LONDON-JAN-25 | GIOSEF FERRARO | Module-Data groups |  SPRINT 1

Giosef92:
Self checklist

- [ X] I have committed my files one by one, on purpose, and for a reason
- [ X] I have titled my PR with COHORT_NAME | FIRST_NAME LAST_NAME | REPO_NAME | WEEK 
- [ X] I have tested my changes
- [ X] My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/contributing/)
- [ X] My changes meet the [requirements](./README.md) of this task

### DIFF
--- a/Sprint-1/implement/dedupe.js
+++ b/Sprint-1/implement/dedupe.js
@@ -1,1 +1,5 @@
-function dedupe() {}
+function dedupe() {}function dedupe(arr){
+    return [... new Set(arr)];
+}
+
+module.exports = dedupe;
--- a/Sprint-1/implement/dedupe.test.js
+++ b/Sprint-1/implement/dedupe.test.js
@@ -1,4 +1,3 @@
-const dedupe = require("./dedupe.js");
 /*
 Dedupe Array
 
@@ -16,12 +15,22 @@ E.g. dedupe([1, 2, 1]) target output: [1, 2]
 // Given an empty array
 // When passed to the dedupe function
 // Then it should return an empty array
-test.todo("given an empty array, it returns an empty array");
+describe("dedupe",()=> {
+    test.todo("given an empty array, it returns an empty array");
+})
 
 // Given an array with no duplicates
 // When passed to the dedupe function
 // Then it should return a copy of the original array
+test("given an array with no duplicates, it returns a copy of the original array", () => {
+    expect(dedupe([1, 2, 3])).toEqual([1, 2, 3]);
+    expect(dedupe(['a', 'b', 'c'])).toEqual(['a', 'b', 'c']);
+  });
 
 // Given an array with strings or numbers
 // When passed to the dedupe function
 // Then it should remove the duplicate values, preserving the first occurence of each element
+test("given an array with duplicates, it returns an array without duplicates", () => {
+    expect(dedupe([1, 2, 2, 3, 3])).toEqual([1, 2, 3]);
+    expect(dedupe(['a', 'a', 'b', 'b', 'c'])).toEqual(['a', 'b', 'c']);
+  });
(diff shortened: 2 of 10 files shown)


Sprint-1/implement/dedupe.js
```diff
@@ -1 +1,5 @@
-function dedupe() {}
+function dedupe() {}function dedupe(arr){
```
day-lee, line 1: You should define function once, Please delete the duplicate.

Sprint-1/implement/dedupe.js
```diff
@@ -1 +1,5 @@
...
+    return [... new Set(arr)];
```
day-lee, line 2: Can you explain the characteristics of Set object and why we need to use spread operator here?

Sprint-1/implement/dedupe.test.js
```diff
@@ -16,12 +15,22 @@ E.g. dedupe([1, 2, 1]) target output: [1, 2]
 // Given an empty array
 // When passed to the dedupe function
 // Then it should return an empty array
-test.todo("given an empty array, it returns an empty array");
+describe("dedupe",()=> {
```
day-lee, line 18: Have you run the test using `npm test`? Your current function has syntax error so it won't pass the test. Please make sure to run the test before you share the code with others.

Sprint-1/implement/max.test.js
```diff
@@ -16,28 +16,47 @@ const findMax = require("./max.js");
...
 // When passed to the max function
 // Then it should return the closest one to zero
+test("given an array with just negative numbers, returns the closest one to zero", () => {
+    expect(findMax([-10, -50, -30, -40])).toBe(-10);
+});
 
 // Given an array with decimal numbers
 // When passed to the max function
 // Then it should return the largest decimal number
+test("given an array with decimal numbers, returns the largest decimal number", () => {
```
day-lee, line 46: JavaScript uses floating-point numbers, which are not always precise and lead to rounding issues.
If you test the following line expect(sum([1.6, 2.5, 3.1])).toBe(7.2) It will cause an error.
Can you refactor the sum() function so it passes this test case?

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,42 @@ const sum = require("./sum.js");
...
 // Then it should still return the correct total sum
+test("given an array containing negative numbers, returns the correct total sum", () => {
+    expect(sum([10, -5, 5])).toBe(10); // 10 + (-5) + 5 = 10
+});
 
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
+test("given an array with decimal/float numbers, returns the correct total sum", () => {
+    expect(sum([1.5, 2.5, 3.0])).toBe(7.0);
```
day-lee, line 38: JavaScript uses floating-point numbers, which are not always precise and lead to rounding issues.
If you test the following line expect(sum([1.6, 2.5, 3.1])).toBe(7.2) It will cause an error.
Can you refactor the sum() function so it passes this test case?

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,7 @@
 function sum(elements) {
+    const numbers = elements.filter(element => typeof element === 'number');
+
+    return numbers.reduce((acc, current) => acc + current, 0);
```
day-lee, line 4: To pass the test `expect(sum([1.6, 2.5, 3.1])).toBe(7.2)`, You will need to convert the types for returning value of reduce function.

PR #398, opened 2025-03-17
Glasgow ITP-JAN-2025 | Aiah Ibrahim | Module-Data-Group | Alarmclock

aiahibrahim:
Self checklist

### DIFF
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -1,6 +1,11 @@
 function setAlarm() {
   let timeLeft = parseInt(document.getElementById("alarmSet").value);
   let display = document.getElementById("timeRemaining");
+  // Input validation
+  if (isNaN(timeLeft) || timeLeft <= 0) {
+      alert("Please enter a valid positive number.");
+      return;
+  }
 
   function updateDisplay() {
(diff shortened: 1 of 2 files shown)


cjyuan (commented):
Code is very clean and elegant.

Here are my comments and suggestions:
- If the user enters a time and then clicks the "Set Alarm" button multiple times, the countdown clock will not display properly.

- Some unusual input values can mess up the display. Can you add code to sanitise the input?

- When the countdown reaches 00:00, there is a one second delay before the alarm sound is played. Is this by design?

aiahibrahim:
ive changed the code, is it ok now?? thanks for the review

cjyuan:
The input validation is good now.

Would you also try to overcome these issues? It should only involve changing/adding few lines of code.
- If the user enters a time and then clicks the "Set Alarm" button multiple times, the countdown clock will not display properly.
- When the countdown reaches 00:00, there is a one second delay before the alarm sound is played (instead of playing the alarm sound immediately)

cjyuan:
You are in the right direction. Where you declare the `countdownInterval` variable can affect the outcome.

aiahibrahim:
i've declared countdownInterval in the global scope so it can be a global variable,

aiahibrahim:
can i add complete label ??

PR #402, opened 2025-03-17
ITP_LONDON_JAN_2025| CRISTINA_MARIA_LEASCU | DATA_GROUPS |Quote-Generator

CristinaLeascu:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-3/quote-generator/quotes.js
+++ b/Sprint-3/quote-generator/quotes.js
@@ -491,3 +491,24 @@ const quotes = [
 ];
 
 // call pickFromArray with the quotes array to check you get a random quote
+
+// event listener for page load
+document.addEventListener("DOMContentLoaded", function () {
+  const randomQuote = pickFromArray(quotes);
+
+  document.getElementById("quote").textContent = `"${randomQuote.quote}"`;
+  document.getElementById("author").textContent = `- ${randomQuote.author}`;
+});
+
+//event listener for button
+
+const quoteElement = document.getElementById("quote");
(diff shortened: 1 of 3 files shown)


Sprint-3/quote-generator/quotes.js
```diff
@@ -491,3 +491,24 @@ const quotes = [
...
+
+const quoteElement = document.getElementById("quote");
+const authorElement = document.getElementById("author");
+const button = document.querySelector("#new-quote");
+
+button.addEventListener("click", () => {
+  const randomQuote = pickFromArray(quotes);
+
+  quoteElement.textContent = `"${randomQuote.quote}"`;
+  authorElement.textContent = `- ${randomQuote.author}`;
```
cjyuan, line 513: If we place these in a function, we can replace these three lines of code (and also code at lines 497-500) by a function call. 

Note: Inside this function, we can access the global variables such as `quoteElement` as long as we initialise the variables before calling the function. Because function definition in JS is hoisted (see: https://developer.mozilla.org/en-US/docs/Glossary/Hoisting), we can define our function "in any order".

cjyuan (commented):
Code works great.

The first line in `quote.js` is `// DO NOT EDIT BELOW HERE`. So it would be a better practice to respect that line and insert our code before that line.

A small part of your code can be refactored (to reduce the amount of code).

Feel free to mark this as "Complete" anytime.

PR #405, opened 2025-03-18
WM | ITP-JAN-25 | HATEF_EIDI | Module-Data-Groups | Sprint 3| Alarm Clock

HatefEidi:
## Learners, PR Template

Self checklist

- [*] I have committed my files one by one, on purpose, and for a reason
- [* ] I have titled my PR with COHORT_NAME | FIRST_NAME LAST_NAME | REPO_NAME | WEEK 
- [ *] I have tested my changes
- [ *] My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/contributing/)
- [ *] My changes meet the [requirements](./README.md) of this task

## Changelist

Alarm Clock project, where the user type the number of seconds in an input, time gets decremented by every second and when time remaining reaches to 0, alarm goes off, and they can stop the alarm by clicking the stop alarm button.
## Questions

I couldn't use the jest file, I have no idea how test looks like for testing JavaScript DOM, do we come across it later?


Sprint-3/Prep/index.html
illicitonion, line 1: It looks like you included files from the Prep in this PR, which should only include the Alarm Clock exercise - please can you remove the prep files from this PR?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,7 +1,61 @@
-function setAlarm() {}
+function setAlarm() {
+  //Converting the time of now value to seconds from milliseconds
+  //Getting the time value from the input field, converting it into number, this will be in seconds
+  const alarmTime = Number(document.getElementById("alarmSet").value)   ;
```
illicitonion, line 4: It looks like you have some inconsistent formatting here - there are a bunch of extra spaces before the `;`, later on you sometimes do or don't have  spaces around operators, etc.

It's generally easier to read code that is formatting consistently - please can you reformat your code so it all looks the same.

If you haven't heard about prettier, maybe look into it, it can help you out with this.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,7 +1,61 @@
...
+  const checkAlarm=setInterval(() => {
+    
+    //Calculating current time inside the interval
+     currentTime++;
+
+    //Calculating the difference between the alarm time and the current time
+    const diff = alarmTime - currentTime;
+
+    //Updating the time remaining text in the HTML element
+    document.getElementById('timeRemaining').innerText="Time Remaining: "+timeFormatter(diff);
```
illicitonion, line 18: This looks like the same code as is on line 6.

Let's imagine we wanted to change the text it showed from "Time Remaining:" to "Time Left:" - we would need to make sure we remembered to change both lines 6 and 18.

Can you think of a way to make it so that we would only need to change _one_ line of code if we wanted to make that change, rather than two?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,7 +1,61 @@
...
+
+    
+
+    //If the difference is less than or equal to zero, stop the interval and play the alarm sound
+    if(diff <= 0){
+      clearInterval(checkAlarm);
+      playAlarm();
+      
+      
+      // return;
```
illicitonion, line 28: Why is this comment here?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,7 +1,61 @@
...
+    }
+  }, 1000);
+}
+
+//formatting remaining seconds in the format of MM:SS
+function timeFormatter(seconds) {
+  const remainingSeconds= seconds % 60;
+  const remainingMinutes = Math.floor(seconds / 60);
+  return `${remainingMinutes.toString().padStart(2, '0')}:${remainingSeconds.toString().padStart(2, '0')}`;
```
illicitonion, line 37: This is a bit hard to read because there's quite a lot on one line inside one string literal - could you think of a way to split this up a bit so that there isn't so much going on in one line?

illicitonion (commented):
This is generally looking good - I left some comments on a few things to think about.

On the testing - how are you trying to run the tests? What happens when you try? Do you get error messages?

Sprint-3/Prep/index.html
HatefEidi: I tried to get rid of the Prep file, however since I pushed the commits to the parent branch (main) whenever I request a pull, the parent changes appear to the PR.

Sprint-3/alarmclock/alarmclock.js
HatefEidi: Oops I missed that, now is fixed, thanks you, that does not normally happen but will be more careful

Sprint-3/alarmclock/alarmclock.js
HatefEidi: The reason is as soon as the user set alarm the value is added to the title, I can only keep line 18, but the title updates after 1 second interval.
Now removed line 6, which means the title is updated after 1 second, if that needs to be changed and you prefer the user to see the title as soon as they click set alarm button, please let me know so I changed it to what it was before.

Sprint-3/alarmclock/alarmclock.js
HatefEidi: I wanted to check if it works the same without having return statement, which it did. Now is removed

Sprint-3/alarmclock/alarmclock.js
HatefEidi: Sure, created a separate function which takes two parameter, min and sec, and return the formatted time to time formatter function

Sprint-3/alarmclock/alarmclock.js
illicitonion: I agree that you need to update the element both when the alarm is set, and when it changes. What I'm asking here is: How can you do that _without repeating exactly the same code_?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,7 +1,67 @@
...
+      clearInterval(checkAlarm);
+      playAlarm();
+      
+      
+    }
+  }, 1000);
+}
+
+//formatting remaining seconds in the format of MM:SS
+function timeFormatter(seconds) {
```
illicitonion, line 33: We normally name functions like verbs because they _do_ things - so `formatTime` rather than `timeFormatter`. (Same comment applies to `timePadder`)

Sprint-3/Prep/index.html
illicitonion: I'd recommend sitting with someone in class on Saturday, or on a call, to try to get this fixed. You should be able to remove this file just on your branch. But also if you've accidentally pushed it to your `main`, you should be able to remove it from there too.

Sprint-3/alarmclock/alarmclock.js
illicitonion: These are much easier to read - thanks!

PR #407, opened 2025-03-18
ITP-LONDON | PRISCILLA-EMEBO | Quote-generator

Priscilla-EM:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-3/quote-generator/quotes.js
+++ b/Sprint-3/quote-generator/quotes.js
@@ -16,6 +16,10 @@
 // pickFromArray(['a','b','c','d'])     // maybe returns 'c'
 
 // You don't need to change this function
+quote = document.getElementById("quote");
+author = document.getElementById("author");
+newQuote = document.getElementById("new-quote");
+
 function pickFromArray(choices) {
   return choices[Math.floor(Math.random() * choices.length)];
 }
@@ -491,3 +495,11 @@ const quotes = [
 ];
 
 // call pickFromArray with the quotes array to check you get a random quote
+function pickRandomQuote() {
+  let randomObject = pickFromArray(quotes);
+  quote.innerHTML = randomObject.quote;
+  author.innerHTML = randomObject.author;
+}
+pickRandomQuote();
+newQuote.addEventListener("click", pickRandomQuote); 
+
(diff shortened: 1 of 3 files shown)


Sprint-3/quote-generator/quotes.js
```diff
@@ -491,3 +495,11 @@ const quotes = [
 ];
 
 // call pickFromArray with the quotes array to check you get a random quote
+function pickRandomQuote() {
+  let randomObject = pickFromArray(quotes);
+  quote.innerHTML = randomObject.quote;
+  author.innerHTML = randomObject.author;
+}
+pickRandomQuote();
+newQuote.addEventListener("click", pickRandomQuote);
```
cjyuan, line 504: Why not place this code together with those at lines 19-21?

Sprint-3/quote-generator/quotes.js
```diff
@@ -16,6 +16,10 @@
 // pickFromArray(['a','b','c','d'])     // maybe returns 'c'
 
 // You don't need to change this function
+quote = document.getElementById("quote");
+author = document.getElementById("author");
+newQuote = document.getElementById("new-quote");
```
cjyuan, line 21: 1. To make the variable names (for storing DOM objects) more meaningful, a good practice is to add a suffix to the variable names. For examples, `quoteEl`, `authorEl`, `newQuoteButton`. 

2. These variables are not properly declared.

cjyuan (commented):
Code is working good but could use some improvement.

The first line in `quote.js` is `// DO NOT EDIT BELOW HERE`. So it would be a better practice to respect that line and insert our code before that line.

Sprint-3/quote-generator/quotes.js
Priscilla-EM: innerText returns just the text without formatting and has better perfomance while innerHtml returns text with formatting. 

Hello @cjyuan Could you please advise me on what to do? When I placed the function pickRandomQuotes before the Quotes array, the quotes weren't displaying in my browser, but when it is placed after the quotes array, I was able to see it.

cjyuan: We could place all the code to be executed once when the page is loaded in 

```
window.addEventListener("load", () => {
  pickRandomQuote();
  newQuote.addEventListener("click", pickRandomQuote); 
});
```

cjyuan: Another reason to avoid using `innerHTML` is that some special characters in the text (when treated as HTML code) may not get displayed properly. If the quotes can be supplied by users, hacker could  supply malicious code as a quote and use this loophole to inject the code into the page. The quotes in this exercise are probably safe, but I just thought it is important for you to recognise the difference.

PR #413, opened 2025-03-19
ITP-london2025| Sabita Shrestha |Module-Data-Groups | Sprint2

SabitaShrestha325:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,8 +1,8 @@
 function contains(obj, property) {
-  if (typeof obj !== "object" || obj === null || typeof property !== "string") {
+  if (!obj || typeof property !== "string" || Array.isArray(obj)) {
     return false;
   }
-  return obj.hasOwnProperty(property);
+  return property in obj;
 }
 
 module.exports = contains;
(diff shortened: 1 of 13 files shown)


Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,8 @@
-function contains() {}
+function contains(obj, property) {
+  if (typeof obj !== "object" || obj === null || typeof property !== "string") {
```
day-lee, line 2: The typeof operator returns "object" for arrays since arrays are objects. Therefore, using typeof obj !== "object" is unnecessary. Can you think of another way to check if obj is an array?

Sprint-2/implement/querystring.js
day-lee, line 1: In the real world, URLs are usually encoded with `percent encoding` to represent special characters. 
In URL, 
1. & can be used as a separator (key=value&key=value...) 
2. special character within a key or value. if we were to use & as a special character, we need to encode it.

SabitaShrestha325:
Hi, day-lee,
Thank for review
I have updated the code as you have mention please review .

day-lee:
Code looks good! Good job👍

PR #435, opened 2025-03-23
ITP LONDON_JAN 2025 | SAMUNTA SUNUWAR | MODULE DATA GROUPS| SPRINT 1

Samunta:
Self checklist

### DIFF
--- a/Sprint-1/implement/dedupe.js
+++ b/Sprint-1/implement/dedupe.js
@@ -4,12 +4,14 @@ function dedupe(arr) {
   for (let i = 0; i < arr.length; i++) {
     let isDuplicate = false;
 
+    // iterates over uniqueArray and checks if arr[i] is in that array
     for (let j = 0; j < uniqueArray.length; j++) {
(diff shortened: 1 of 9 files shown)


Sprint-1/implement/dedupe.js
```diff
@@ -1 +1,23 @@
-function dedupe() {}
+function dedupe(arr) {
```
day-lee, line 1: Good job! You can loop through each element to check if it is a duplicate using a flag. Additionally, you can use a Set object to achieve the same result, taking advantage of the characteristics of a Set. Can you think of another solution using a Set?

Samunta: Thank you for your feedback, I have refactored my code using Set

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,19 @@
 function findMax(elements) {
```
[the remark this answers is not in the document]
Samunta: Thank you for your feedback, I have now refactored my code using Math.Max

PR #437, opened 2025-03-23
ITP_LONDON_JAN 2025|Samunta Sunuwar|Module-Data-Groups|Sprint 2

Samunta:

### DIFF
--- a/Sprint-2/implement/lookup.js
+++ b/Sprint-2/implement/lookup.js
@@ -1,5 +1,11 @@
-function createLookup() {
-  // implementation here
+function createLookup(pairs) {
+  const lookup = {};
+
+  for (const pair of pairs) {
+    lookup[pair[0]] = pair[1];
+  }
+
+  return lookup;
 }
 
 module.exports = createLookup;
--- a/Sprint-2/implement/lookup.js
+++ b/Sprint-2/implement/lookup.js
@@ -1,4 +1,17 @@
 function createLookup(pairs) {
+  //Checking if input is an array and throwing error if it's not
+  if (!Array.isArray(pairs)) {
+    throw new Error("Input must be an array.");
+  }
+
+  // Checking if every element inside the array is also an array (a pair) and throwing error if it's not
+  for (const pair of pairs) {
+    if (!Array.isArray(pair) || pair.length !== 2) {
+      throw new Error(
+        "Each element in the array must be an array with exactly two elements."
+      );
+    }
+  }
   const lookup = {};
 
   for (const pair of pairs) {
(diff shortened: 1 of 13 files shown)


Sprint-2/implement/lookup.js
```diff
@@ -1,5 +1,11 @@
-function createLookup() {
-  // implementation here
+function createLookup(pairs) {
```
day-lee, line 1: Well done! As a challenge, Can you think of edge cases as well? When arguments are not an array, or array of arrays is not an array, then you can exit early.

Sprint-2/implement/querystring.js
```diff
@@ -6,11 +6,15 @@ function parseQueryString(queryString) {
   const keyValuePairs = queryString.split("&");
 
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
+    // const [key, value] = pair.split("=");
+    // queryParams[key] = value;
+    const index = pair.indexOf("="); // Find the first "="
+    const key = pair.slice(0, index); // Extract key
+    const value = pair.slice(index + 1); // Extract everything after "="
     queryParams[key] = value;
```
day-lee, line 14: Good job!
In the real world, URLs are usually encoded with [percent encoding](https://developer.mozilla.org/en-US/docs/Glossary/Percent-encoding) to represent special characters. In URL, `&` can be used as a separator (key=value&key=value...) or special character within a key or value. if we were to use `&` as a special character, we need to encode it.

Can you refactor the code so that parseQueryString("a%25b=c%26d") results in `{ "a%b": "c&d" }`?
FYI, `%25` is same as `%`, `%26` is same as `&`

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,19 @@
-function tally() {}
+function tally(items) {
+  if (!Array.isArray(items)) {
+    throw new Error("Input must be an array");
+  }
+
+  const object = {};
+
+  for (let element of items) {
```
day-lee, line 8: Good job! As a challenge, can you think of a code that can do the same thing in one line, using object bracket notation that takes a variable as a key and logical OR operator `||` ?

Sprint-2/implement/lookup.js
Samunta: Thank you for the feedback, I have refactored my code and have added test case to check the code is working

Sprint-2/implement/querystring.js
Samunta: Thank you for the feedback, I have refactored my code and have added test case to check the code is working

Sprint-2/implement/tally.js
Samunta: Thank you for the feedback, I have refactored my code using object bracket and OR operator

PR #442, opened 2025-03-25
ITP-May-2025 | London | Chi Mbah | Module-Data-Groups | Week 8 | Sprint 2

Chi-Mb:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,19 +1,10 @@
 function contains(object, key) {
-    if (typeof object !== "object") {
+    if (typeof object !== "object" || object === null || Array.isArray(object)) {
         return false;
-    } else if (object === null) {
-        return false;
-    } else if (Array.isArray(object)) {
-        return false;
-    } else {
-        if (key in object) {
-            return true;
-        } else {
-            return false;
-        }
     }
-
+    return key in object;
 }
 
-
 module.exports = contains;
+
+
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,3 +1,19 @@
-function contains() {}
+function contains(object, key) {
+    if (typeof object !== "object") {
+        return false;
+    } else if (object === null) {
+        return false;
+    } else if (Array.isArray(object)) {
+        return false;
+    } else {
+        if (key in object) {
+            return true;
+        } else {
+            return false;
+        }
+    }
+
+}
+
 
 module.exports = contains;
--- a/Sprint-2/implement/lookup.js
+++ b/Sprint-2/implement/lookup.js
@@ -1,14 +1,17 @@
-function createLookup() {
-  // implementation here
+function createLookup(pairs) {
+// implementation here
+  if (!Array.isArray(pairs) || !pairs.every(pair => Array.isArray(pair) && pair.length === 2)) {
+    throw new Error("Invalid input: Expected an array of key-value pairs.");
+  }
+
   let lookup = {};
 
-    for (let pair of pairs) {
-        let [country, currency] = pair;
-        lookup[country] = currency;
-    }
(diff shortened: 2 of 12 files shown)


Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,19 @@
-function contains() {}
+function contains(object, key) {
+    if (typeof object !== "object") {
```
day-lee, line 2: The logic is good! You could use an if-else statement, but since all the results return false, you can use the logical OR operator (||) to present it in one line using if (condition1 || condition 2 || condition3) Can you refactor the code accordingly?

Sprint-2/implement/querystring.js
```diff
@@ -1,13 +1,21 @@
...
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
-    queryParams[key] = value;
+    const firstEqualSign = pair.indexOf("=");  
+    if (firstEqualSign === -1) {
+      queryParams[pair] = null; 
+    } else {
+      const key = pair.substring(0, firstEqualSign); 
+      const value = pair.substring(firstEqualSign + 1); 
+      queryParams[key] = value;
```
day-lee, line 17: Good job!
In the real world, URLs are usually encoded with [percent encoding](https://developer.mozilla.org/en-US/docs/Glossary/Percent-encoding) to represent special characters. In URL, & can be used as a separator (key=value&key=value...) or special character within a key or value. if we were to use & as a special character, we need to encode it.

Sprint-2/implement/lookup.js
```diff
@@ -1,5 +1,14 @@
 function createLookup() {
   // implementation here
+  let lookup = {};
+
```
day-lee, line 4: Well done! As a challenge, Can you think of edge cases as well? When arguments are not an array, or array of arrays is not an array, then you can exit early returning Error.

Sprint-2/interpret/invert.js
```diff
@@ -18,12 +18,67 @@ function invert(obj) {
...
 // e) Fix the implementation of invert (and write tests to prove it's fixed!)
+
+function invert(obj) {
+  if (!obj || typeof obj !== "object" || Array.isArray(obj)) {
+    throw new Error("Invalid input: expected an object.");
+  }
+  
+  const invertedObj = {};
+
+  for (const key in obj) {
```
day-lee, line 51: Well done! You can loop through an object using a for...in loop. As a challenge, can you think of a solution using `Object.entries` as well?

Sprint-2/implement/contains.js
Chi-Mb: @day-lee Thanks for the feedback. I've implemented the changes you suggested

Sprint-2/implement/querystring.js
Chi-Mb: @day-lee I have refactored the code implementing the decodeURIComponent to correctly decode special characters

day-lee: Great work! It would be nice if you could add the test case as well but it's up to you.

Chi-Mb: @day-lee The test cases for this file are located in the test file

Sprint-2/implement/lookup.js
Chi-Mb: @day-lee I've implemented your suggestion

Sprint-2/interpret/invert.js
Chi-Mb: @day-lee I replaced the for...in loop with a for...of loop using Object.entries(). This has gives me direct access to both keys and values without needing obj[key].

day-lee: Great!

PR #443, opened 2025-03-25
ITP-May-2025 | London | Chi Mbah | Module-Data-Groups | Week 9 | Sprint 3 | Alarm Clock

Chi-Mb:
## Learners, PR Template

Self checklist

## Changelist
- Fixed issue with multiple alarms running at once.
- Clicking 'Set alarm' now restarts the timer if one is already running.

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -28,7 +28,7 @@ function setAlarm() {
   }
 
   updateClock();
-  countdownInterval = setInterval(updateClock, 1000);
+  countdownInterval = setInterval(updateClock, ONE_SECOND_MS);
 }
 
 function getClockDigit(number) {
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -1,4 +1,38 @@
-function setAlarm() {}
+function setAlarm() {
+  const initialTime = parseInt(document.getElementById("alarmSet").value);
+  
+  if (isNaN(initialTime) || initialTime <= 0) {
+    alert("Please enter a valid time.");
+    return;
+  }
+
+  let countdownInterval;
+  let remainingTime = initialTime;
+
+  function updateClock() {
+    const minutes = Math.floor(remainingTime / 60);
+    const seconds = remainingTime % 60;
+    document.getElementById("timeRemaining").textContent = `Time Remaining: ${getClockDigit(minutes)}:${getClockDigit(seconds)}`;
+
+    if (remainingTime > 0) {
+      remainingTime--;
+    } else {
+      playAlarm();
+      clearInterval(countdownInterval);
+    }
+  }
+
+  updateClock();
+  countdownInterval = setInterval(updateClock, 1000);
+}
+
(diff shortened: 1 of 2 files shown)


Sprint-3/alarmclock/index.html
jenny-alexander, line 1: The title in the browser tab says 'Title here'. Could you update this with something that provides more context to the user? 
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/title

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,38 @@
-function setAlarm() {}
+function setAlarm() {
+  const initialTime = parseInt(document.getElementById("alarmSet").value);
+  
+  if (isNaN(initialTime) || initialTime <= 0) {
```
jenny-alexander, line 4: You did a good job validating that the user entered a number. I'm able to enter a decimal like '5.5' and the timer will countdown from 5 seconds. Can you think of a way to validate for a decimal too?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,38 @@
...
+    if (remainingTime > 0) {
+      remainingTime--;
+    } else {
+      playAlarm();
+      clearInterval(countdownInterval);
+    }
+  }
+
+  updateClock();
+  countdownInterval = setInterval(updateClock, 1000);
```
jenny-alexander, line 26: What does 1000 mean? It would be even better to represent this as a constant with a name that explains what this value is. :)

jenny-alexander (changes requested):
The alarm clocks works well. Here are a few things I see that should be adjusted:
- I can enter a value, click 'Set alarm', enter another value, click 'Set alarm' and there will be 2 alarms going on at once. Can you think of a way to reset the alarm each time the user clicks on 'Set alarm'?
- The changelist portion of your pull request is currently empty. This is an important part that helps other engineers understand what you've modified.
Could you please add a brief description of your changes in the changelist section? Even a simple summary is will be beneficial. :)

Sprint-3/alarmclock/alarmclock.js
Chi-Mb: @jenny-alexander Thanks for the feedback. I updated the function to use Number() instead of parseInt, and added Number.isInteger() to make sure only whole numbers are accepted

jenny-alexander (changes requested):

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,43 @@
...
+    if (remainingTime > 0) {
+      remainingTime--;
+    } else {
+      playAlarm();
+      clearInterval(countdownInterval);
+    }
+  }
+
+  updateClock();
+  countdownInterval = setInterval(updateClock, ONE_SECOND_MS);
```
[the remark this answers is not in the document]
Chi-Mb: @jenny-alexander  You're right, thanks for pointing that out. I've updated the function now. 
Can I go ahead and mark the assignment as complete?

PR #445, opened 2025-03-25
ITP_Jan_2025 | Aung_Ye_Kyaw | Module-Data-Group | Coursework/Sprint-3

sarawone:
Hi !

I hope you are doing well !

I would like to thank for your time to review my code .
Kindly check and provide feedback, advices and guidance . 
Thank you so much !

Thanks and Regards,
Aung Ye Kyaw

### DIFF
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -1,4 +1,48 @@
-function setAlarm() {}
+function setAlarm() {
+  
+  let timeValue = document.querySelector("input").value;
+
+
+  // if text box have value 
+  if (timeValue.length != 0)
+  {
+
+    function updateTimer()
+    {
+    
+    // get the h1 element 
+    let timeLabel = document.querySelector("#timeRemaining");
+
+    let minutes , seconds = 0;
+
+    minutes = Math.floor(timeValue/60); // get the minutes from the value 
+    seconds = timeValue%60; // get the seconds from the value 
+
+    timeLabel.textContent = "Time Remaining: " + `${String(minutes).padStart(2,'0')} : ${String(seconds).padStart(2,'0')}`;
+
+    //reduce 1 to the time 
+    timeValue--;  
+
+    //if the time counting reach to 0
+    if (timeValue < 0)
+    {
+      // clear the interval 
+      clearInterval(timer);
+
+      //play the sound 
+      playAlarm();
+    }  
+   
+      
+    
+   } 
+
+  }
+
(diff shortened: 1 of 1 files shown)


Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,48 @@
-function setAlarm() {}
+function setAlarm() {
+  
+  let timeValue = document.querySelector("input").value;
+
+
+  // if text box have value 
+  if (timeValue.length != 0)
```
cjyuan, line 7: Some unusual input values can pass this check and mess up the displayed clock. 
Can you add additional check to reject these "unusual input values"?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,48 @@
...
+      playAlarm();
+    }  
+   
+      
+    
+   } 
+
+  }
+
+  let timer = setInterval(updateTimer,1000)
```
cjyuan, line 42: Can you find out what the value returned by `setInterval()` is? I think a name in the form `...IntervalId` would be more descriptive.

cjyuan (commented):

- Code works fine if a user only clicks the "Set Alarm" button once. However, if the user enters a time and then clicks the "Set Alarm" button multiple times, the countdown clock will not display properly. Can you fix the issue? (Hint: The solution may involve declaring some of your variables in the global scope).

- After the user clicks the "Set Alarm" button, currently there is a one second delay before the alarm clock starts working. Can you find a way to remove the delay?

Sprint-3/alarmclock/alarmclock.js
sarawone: I have changed the name . Kindly check . Thank you so much .

Sprint-3/alarmclock/alarmclock.js
sarawone: I have fixed the code . Kindly check and provide feedback. Thank you.

sarawone:
> * You missed updating `index.html` according to an instruction in `readme.md`.
> * Code works fine if a user only clicks the "Set Alarm" button once. However, if the user enters a time and then clicks the "Set Alarm" button multiple times, the countdown clock will not display properly. Can you fix the issue? (Hint: The solution may involve declaring some of your variables in the global scope).
> * After the user clicks the "Set Alarm" button, currently there is a one second delay before the alarm clock starts working. Can you find a way to remove the delay?

I have done the above points . Kindly check and provide feedback. Thank you so much.

PR #452, opened 2025-03-29
Update tests for Sprint 1 'fix median' task

tyzia:
This PR updates the tests for the fixed version of `median` function. The tests expect from the students to implement `median` function which will:

- Sort arrays, leaving only numeric values. Median can be calculated only on numeric values.
- Return `null` if no numeric values provided.


tyzia:
`solutions` branch has been already updated accordingly: https://github.com/CodeYourFuture/Module-Data-Groups/pull/390

illicitonion (approved):
Thanks!

PR #458, opened 2025-03-30
JAN 2025 ITP/ ELFREDAH KEVIN-ALERECHI/ DATA GROUPS/Sprint 2

Elfredah:
I deleted the other sprint to ensure I submit the sprint 1 separately

## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions


day-lee:
Hi, @Elfredah. Please make sure to add the label needs review so it shows up on the volunteers' dashboard. You can get a quicker review once you add a label to your PR.

Sprint-1/fix/median.js
```diff
@@ -2,10 +2,56 @@
...
+    return (sortedList[middleIndex - 1] + sortedList[middleIndex]) / 2;
+  } else {
+    return sortedList[middleIndex];
+  }
+}
+
+module.exports = calculateMedian;
+
+
+//======> solution to the right codes to solve the issue is below
```
day-lee, line 34: The code looks good and passes the test. 
One thing I want to mention is that when you share the code with others, please provide the final version to avoid any confusion. If you need the code for future reference, you can comment it out and share it, but the best practice would be to remove any unnecessary code before PR.

Sprint-1/fix/median.test.js
```diff
@@ -8,19 +8,70 @@ const calculateMedian = require("./median.js");
...
+  });
+
+  test("returns undefined for undefined input", () => {
+    expect(calculateMedian()).toBeUndefined(); // Undefined input case
   });
 });
+
+
+
+//const calculateMedian = require("./median.js");
```
day-lee, line 38: Please delete any unnecessary commented-out lines.

Sprint-1/fix/median.test.js
```diff
@@ -8,19 +8,70 @@ const calculateMedian = require("./median.js");
...
+
+  //test("doesn't modify the input", () => {
+    //const list = [1, 2, 3];
+    //calculateMedian(list);
+
+    //expect(list).toEqual([1, 2, 3]);
+  //});
+//});
+
+//=======> solution to the right codes to solve the issue is below
```
day-lee, line 59: Please delete any unnecessary commented-out lines.

Sprint-1/fix/median.test.js
```diff
@@ -8,19 +8,70 @@ const calculateMedian = require("./median.js");
...
+  const middleIndex = Math.floor(sortedList.length / 2);
+
+  if (sortedList.length % 2 === 0) {
+    return (sortedList[middleIndex - 1] + sortedList[middleIndex]) / 2;
+  } else {
+    return sortedList[middleIndex];
+  }
+}
+
+module.exports = calculateMedian;
```
day-lee, line 77: The est.js file should only contain the test suite. Since the function is already written in median.js so you can delete those lines.

Sprint-1/implement/dedupe.js
```diff
@@ -1 +1,25 @@
-function dedupe() {}
+//function dedupe() {}
+// dedupe.js
+function dedupe(arr) {
```
day-lee, line 3: Nicely done! As a bonus challenge, can you think of a solution that uses `Set` object as well?

Sprint-1/implement/max.test.js
```diff
@@ -41,3 +41,48 @@ test.todo("given an empty array, returns -Infinity");
 // Given an array with only non-number values
 // When passed to the max function
 // Then it should return the least surprising value given how it behaves for all other inputs
+
+const findMax = require("./max");
```
day-lee, line 45: Have you run the test with the `npm test <file path>` command? I tried it on my local machine, and it fails. I noticed that you wrote this line of code twice in the file.

day-lee: You will need to delete one of those duplicate `const findMax = require("./max");` line to make it work.

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,16 @@
-function findMax(elements) {
-}
+//function findMax(elements) {
+//}
 
-module.exports = findMax;
+//module.exports = findMax;
+function findMax(elements) {
```
day-lee, line 5: With current code, test fails under these conditions.
```
given an empty array, returns -Infinity
given an array with non-number values, returns the max and ignores non-numeric values
given an array with only non-number values, returns -Infinity 
```
You need a line of code that checks the type of element in an array(elements) and find max from filteredArray.

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,16 @@
...
+    // Return undefined if the array is empty
+    if (elements.length === 0) {
+      return undefined;
```
day-lee, line 8: In the test, it specifies that it needs to return -Infinity, not undefined. This needs to be changed. Additionally, you should check the length of filteredArray because if there are only strings in the array, it means the array has no valid numbers and is effectively an empty array.
```
test("given an empty array, returns -Infinity", () => {
    expect(findMax([])).toEqual(-Infinity);
  });
```

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,16 @@
...
+    }
+  
+    // Use Math.max to find the largest number in the array
+    return Math.max(...elements);
```
day-lee, line 12: Good job using Math.max() function with spread operator to spread the elements from an array.

Sprint-1/implement/sum.js
```diff
@@ -2,3 +2,16 @@ function sum(elements) {
 }
 
 module.exports = sum;
```
day-lee, line 4: This should be deleted to run the test as it's duplicate and cause an error.

Sprint-1/implement/sum.test.js
```diff
@@ -34,3 +34,43 @@ test.todo("given an empty array, returns 0")
...
+  // Test for an array with negative numbers
+  test("given an array containing negative numbers, returns the correct total sum", () => {
+    expect(sum([10, -5, 5])).toEqual(10); // 10 + (-5) + 5 = 10
+    expect(sum([-10, -20, -30])).toEqual(-60); // -10 + (-20) + (-30) = -60
+  });
+
+  // Test for an array with decimal/float numbers
+  test("given an array with decimal/float numbers, returns the correct total sum", () => {
+    expect(sum([1.5, 2.5, 3.5])).toEqual(7.5); // 1.5 + 2.5 + 3.5 = 7.5
+    expect(sum([1.1, -0.2, 3.2])).toEqual(4.1); // 1.1 + (-0.2) + 3.2 = 4.1
```
day-lee, line 62: When I run the test, this test fails as `sum([1.1, -0.2, 3.2]` returns `4.1000000000000005`. This is because JavaScript has one numeric type which is `float`. This can be not precise and has a rounding issue. Can you think of a solution using `parseFloat()` and `toFixed() function`?  Check if your final code passes all test as well.

day-lee:
Great job on completing Sprint 1! You did well overall. I've commented on a few minor issues. You can run the test using npm test <path/filename.test.js>. If you have any more questions, feel free to ask! :)

PR #473, opened 2025-03-31
WEST MIDLANDS-JAN-ITP|SEGUN FOLAYAN|DATA GROUPS|SPRINT-1

segunfolayan:
## Learners, PR Template

Self checklist

- [x ] I have committed my files one by one, on purpose, and for a reason
- [ x] I have titled my PR with COHORT_NAME | FIRST_NAME LAST_NAME | REPO_NAME | WEEK 
- [ x] I have tested my changes

## Questions

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -1,20 +1,14 @@
-// Fix this implementation
-// Start by running the tests for this function
-// If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
+
+module.exports = calculateMedian;
 
 function calculateMedian(list) {
   const middleIndex = Math.floor(list.length / 2);
-  const median=0;
   
-  if (list.length % 2===1){
-    const median= list[middleIndex];
-    return median;
+  if (list.length % 2 === 1) {
+    return list[middleIndex]; // Return the middle element for odd-length lists
   }
-  if (list.length % 2===0 ){
-    const median=(list[middleIndex] + list[middleIndex-1])/2;
-    return median;
-  }
-  //const median = list.splice(middleIndex, 1)[0];
+  
+  return (list[middleIndex] + list[middleIndex - 1]) / 2; // Return the average of the two middle elements for even-length lists
 }
 
 module.exports = calculateMedian;
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -4,8 +4,17 @@
 
 function calculateMedian(list) {
   const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  const median=0;
+  
+  if (list.length % 2===1){
+    const median= list[middleIndex];
+    return median;
+  }
+  if (list.length % 2===0 ){
+    const median=(list[middleIndex] + list[middleIndex-1])/2;
+    return median;
+  }
+  //const median = list.splice(middleIndex, 1)[0];
 }
 
 module.exports = calculateMedian;
--- a/Sprint-1/implement/dedupe.js
+++ b/Sprint-1/implement/dedupe.js
@@ -5,4 +5,5 @@ function dedupe(arr) {
   }
 
 
+module.exports=dedupe;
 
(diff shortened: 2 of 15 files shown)


Sprint-1/fix/median.js
```diff
@@ -4,8 +4,17 @@
 
 function calculateMedian(list) {
   const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  const median=0;
```
day-lee, line 7: Your logic is good, but the code looks a bit repetitive. The variable median is unnecessary. Could you refactor the code to return the result directly instead of assigning it to the variable median?

Sprint-1/implement/dedupe.js
```diff
@@ -1 +1,9 @@
-function dedupe() {}
+
+
+function dedupe(arr) {
+    return [...new Set(arr)];
```
day-lee, line 4: Could you explain the characteristics of `Set object` and why we need to use the spread operator here?

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,18 @@
...
+  for (let Index=0; Index < elements.length; Index++){
+    if (typeof elements[Index] === "number"){
+        total+=elements[Index];
+    }
+    else {
+        total=total + 0;
+    }
+
+   }
+   return total;
```
day-lee, line 12: JavaScript uses floating-point numbers, which are not always precise and lead to rounding issues.
If you test the following line expect(sum([1.6, 2.5, 3.1])).toBe(7.2) It will cause an error.
Can you refactor the sum() function so it passes this test case?

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,50 @@ const sum = require("./sum.js");
...
+    expect(sum([-2,-2-9-10-2,])).toEqual(-25);
+    }
+    );
 
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
+test("given an array with decimal/float numbers, then it should return the correct sum", () => {
+    expect(sum([1.25,1.75,1.6])).toEqual(4.6);
+    }
```
day-lee, line 45: Can you add `expect(sum([1.6, 2.5, 3.1])).toBe(7.2)` and refactor the code to pass the test?

day-lee:
Good job 👍  Let me know if you have any questions!

Sprint-1/fix/median.js
segunfolayan: function calculateMedian(list) {
  
  if (list.length % 2 === 1) {
    return list[middleIndex];
  }
  
  return (list[middleIndex] + list[middleIndex - 1]) / 2; lists
}

module.exports = calculateMedian;

segunfolayan: Thank you

Sprint-1/implement/dedupe.js
segunfolayan: 1. Uniqueness-It does not contain duplicates.
2. No index access- IT cannot be accessed using indices like arrays.
3. Insertion order-The values in the set are iterated in the order in which they are added.
4. Type flexibility- A Set can store values of any type (primitive or object), and even objects can be stored as unique values (by reference).
5. The spread operator is used to convert the Set back into an array.
6. Thank you

Sprint-1/implement/sum.js
segunfolayan: function sum(elements) {
    let total = 0;
    for (let Index = 0; Index < elements.length; Index++) {
        if (typeof elements[Index] === "number") {
            total += elements[Index];
        } else {
            total = total + 0;
        }
    }
    return parseFloat(total.toFixed(10));
}

.......................Thank you

Sprint-1/implement/sum.test.js
segunfolayan: function sum(elements) {
    let total = 0;
    for (let Index = 0; Index < elements.length; Index++) {
        if (typeof elements[Index] === "number") {
            total += elements[Index];
        } else {
            total = total + 0;
        }
    }
    return parseFloat(total.toFixed(10));
}

module.exports = sum;

segunfolayan: Many Thanks

PR #475, opened 2025-03-31
WESTMIDLANDS JAN ITP|SEGUN FOLAYAN|DATA-GROUPS|SPRINT-2

segunfolayan:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions


Sprint-2/implement/querystring.js
```diff
@@ -6,8 +6,10 @@ function parseQueryString(queryString) {
   const keyValuePairs = queryString.split("&");
 
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
-    queryParams[key] = value;
+    const [key, ...value] = pair.split("=");
```
day-lee, line 9: Good job!

segunfolayan: function parseQueryString(queryString) {
  const queryParams = {};
    return queryParams;
  }

    const decodedPair = decodeURIComponent(pair);

    
    const [key, ...value] = decodedPair.split("=");
    const newValue = value.join("=");

    // Store the decoded key and value in the queryParams object
    queryParams[key] = newValue;
  }

  return queryParams;
}

module.exports = parseQueryString;

segunfolayan: Thank you

PR #485, opened 2025-04-05
Westmidlands/Millena-Mesfin/Module Data Groups Sprint#2/Week 10

Millena28:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-2/debug/author.js
+++ b/Sprint-2/debug/author.js
@@ -11,6 +11,8 @@ const author = {
   alive: true,
 };
 
-for (const value of author) {
-  console.log(value);
-}
+
+ for (const key in author) {
+   console.log(key);
+ }
+ 
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,3 +1,5 @@
-function contains() {}
+function contains(obj,key) {
+    return (key in obj);
+}
 
 module.exports = contains;
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -20,16 +20,35 @@ as the object doesn't contains a key of 'c'
 // Given an empty object
 // When passed to contains
 // Then it should return false
-test.todo("contains on empty object returns false");
+test("contains an empty object returns false", () => {
+const obj = {};
+expect(contains(obj,"m")).toBe(false);
+});
 
 // Given an object with properties
 // When passed to contains with an existing property name
 // Then it should return true
+test("contains an object with an existing property should return true", () => {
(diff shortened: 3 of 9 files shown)


Sprint-2/debug/author.js
```diff
@@ -11,6 +11,8 @@ const author = {
...
 };
 
-for (const value of author) {
-  console.log(value);
-}
+
+ for (const key in author) {
+   console.log(key);
+ }
+
```
cjyuan, line 18: Can you modify this to output the property values instead of property names?

Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,5 @@
-function contains() {}
+function contains(obj,key) {
+    return (key in obj);
+}
```
cjyuan, line 3: Consider the following two approaches for determining if an object contains a property:
```
```

Sprint-2/implement/contains.test.js
```diff
@@ -20,16 +20,35 @@ as the object doesn't contains a key of 'c'
...
+});
 
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+test("contains an invalid parameter like array should return false" ,() => {
+const obj = ["code","your","future"];
+expect(contains(obj,'age')).toBe(false);
+expect(contains(obj,'address')).toBe(false);
+});
```
cjyuan, line 54: An array is a kind of object in JS.
The indexes serve as the keys (or property names) in an array. So `contains(["code","your","future"], '0')` would return `true`.
If the function needs to return `false` when the parameter is an array, the function will need to check if the parameter is an array or not.

Sprint-2/implement/tally.test.js
```diff
@@ -23,12 +23,29 @@ const tally = require("./tally.js");
...
 
 // Given an invalid input like a string
 // When passed to tally
 // Then it should throw an error
+test("An array with invalid input should throw an error", () => {
+    const count  = ["hello", "hey", "hey", "hello", "hello", "hi"];
+    const expectedOutput = {hello: 3, hey: 2, hi: 1 };
+  
+    expect(tally(count)).toEqual(expectedOutput);
+  });
```
cjyuan, line 49: The description and the code does not match.

cjyuan (commented):
Code is very elegant!

I think there are some improvements you can make.

Feel free to mark this PR as "Complete" after you made the changes.

Sprint-2/implement/contains.js
Millena28: Thank you cj, I got to differentiate the (propertyName in obj) suits to my function.

PR #491, opened 2025-04-10
ITP_GLASGOW_APR | HANNA_MYKYTIUK | MODULE_DATA_GROUPS | SPRINT_1

HannaOdud:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-1/implement/max.js
+++ b/Sprint-1/implement/max.js
@@ -1,4 +1,48 @@
 function findMax(elements) {
+    let max;
+    // check if elements empty array
+    if (elements.length == 0){
+        return -Infinity;
+    }
+    // check if list contains numbers 
+    let containsNumbers = false;
+    for (let i = 0; i < elements.length; i++){
+        if (typeof elements[i] == 'number'){
+        containsNumbers = true;
+        break;
+        }
+    }
+    // if elements does not contain numbers we should search for values that can be converted to numbers
+    if (containsNumbers == false){
+        for (let i = 0; i < elements.length; i++){
+            //check if variable can be converted to number and check if variable max is assigned
+            if (!(typeof max == 'number') && !isNaN(elements[i])){
+                // if both true assign converted value to max
+                max = Number(elements[i]);
+            }
+            // if max is already assigned and elements[i] value can be converted to number
+            if (typeof max == 'number' && !isNaN(elements[i])){
+                // check if new walue is greater than max
+                if (Number(elements[i]) > max){
+                    max = Number(elements[i]);
+                }
+            }
+        }
+    }
+    // if array contains numeric value we search max only among them
+    for (let i=0; i<elements.length; i++ ){
+        //if max unassigned - we assigning  first numeric value from elements
+        if (!(typeof max == 'number') && typeof elements[i] == 'number'){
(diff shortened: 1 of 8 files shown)


Sprint-1/implement/max.js
```diff
@@ -1,4 +1,48 @@
...
+            }
+            // if max is already assigned and elements[i] value can be converted to number
+            if (typeof max == 'number' && !isNaN(elements[i])){
+                // check if new walue is greater than max
+                if (Number(elements[i]) > max){
+                    max = Number(elements[i]);
+                }
+            }
+        }
+    }
```
cjyuan, line 31: Most people choose to return `-Infinity` when all elements are non-number. That is, they treat such array as having no numbers. 

If we initialised `max` to `-Infinity`, then we would not need to check `!(typeof max == 'number')` and we also do not need lines 19-21. Too much checking can degrade the performance of a program.

We can also return the value of `max` right before line 31. If the array does not contain any number, then we don't have to execute the remaining code.

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,48 @@
...
+    // if array contains numeric value we search max only among them
+    for (let i=0; i<elements.length; i++ ){
+        //if max unassigned - we assigning  first numeric value from elements
+        if (!(typeof max == 'number') && typeof elements[i] == 'number'){
+            max = elements[i];
+        }
```
cjyuan, line 37: We would not need these few lines if `max` was initialised to `-Infinity`.

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,48 @@
...
+        // compare each numeric value with max, and replace max if new value greater
+        if (typeof max == 'number'){
```
cjyuan, line 39: Shouldn't this check `typeof elements[i] == 'number'` instead?

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,29 @@
...
+        }
+    }
+    // if elements does not contain numbers we should search for values that can be converted to numbers
+    if (containsNumbers == false){
+        for (let i=0; i<elements.length; i++){
+           if(!isNaN(elements[i])){
+            sum += Number(elements[i]);
+           } 
+        }
+    }
```
cjyuan, line 18: We could return the sum right before line 18 and avoid unnecessary processing.

Sprint-1/implement/sum.test.js
```diff
@@ -8,29 +8,47 @@ E.g. sum(['hey', 10, 'hi', 60, 10]), target output: 80 (ignore any non-numerical
...
+        const value = sum([-7, -2, -1]);
+        expect(value).toEqual(-10);
+    });
+    // Given an array with decimal/float numbers
+    // When passed to the sum function
+    // Then it should return the correct total sum
+    it("array with decimal/float numbers should return correct sum", () => {
+        const value = sum([7.1, 2.5, 0.6]);
+        expect(value).toEqual(10.2);
+    });
```

So the following could happen
```

```

cjyuan (commented):
I think you can make some improvements in some of your code.

PR #492, opened 2025-04-10
ITP_GLASGOW_APR | HANNA_MYKYTIUK | MODULE__DATA_\GROUPS | SPRINT_2

HannaOdud:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-1/implement/max.test.js
+++ b/Sprint-1/implement/max.test.js
@@ -12,6 +12,16 @@ We have set things up already so that this file can see your function from the o
 
 const findMax = require("./max.js");
 
+describe("findMax", () => {
+    [
+        { input: [], expected: ['a','b','c'] },
+        { input: [5, 1, 1, 2, 3, 2, 5, 8], expected: [5, 1, 2, 3, 8] },
+        { input: [1, 2, 1], expected: [1, 2] },
+    ].forEach(({ input, expected }) =>
+        it(`deduplicated array for [${input}]`, () => expect(findMax(input)).toEqual(expected))
+    );
+});
+
 // Given an empty array
 // When passed to the max function
 // Then it should return -Infinity
--- a/Sprint-2/implement/querystring.js
+++ b/Sprint-2/implement/querystring.js
@@ -6,11 +6,19 @@ function parseQueryString(queryString) {
   const keyValuePairs = queryString.split("&");
 
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
-    queryParams[key] = value;
+    //console.log(pair.split("="));
+    const array = pair.split("=");
+    let value = '';
+    for(let i = 1; i<array.length; i++){
+      value += array[i];
+      if (i != array.length - 1){
+        value += '=';
+      }
+    }
+    queryParams[array[0]] = value;
   }
-
   return queryParams;
 }
+console.log(parseQueryString("equation=x=y+1"));
 
 module.exports = parseQueryString;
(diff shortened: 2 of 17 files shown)


Sprint-1/implement/max.test.js
day-lee, line 1: Please remove unrelated files when you request a code review.

Sprint-2/implement/querystring.js
```diff
@@ -6,11 +6,19 @@ function parseQueryString(queryString) {
...
+    //console.log(pair.split("="));
+    const array = pair.split("=");
+    let value = '';
+    for(let i = 1; i<array.length; i++){
+      value += array[i];
+      if (i != array.length - 1){
+        value += '=';
+      }
+    }
+    queryParams[array[0]] = value;
```
cjyuan, line 18: Please note that in real querystring, both `key` and `value` are ***percent-encoded*** or ***URL encoded*** in the URL. For example, the string "5%" will be encoded as "5%25". So to get the actual value of "5%25" (whether it is a key or value in the querystring), you should call a function to decode it.

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,31 @@
-function tally() {}
+function tally(array) {
+    try {
+        if (typeof(array) == 'string'){
```
cjyuan, line 3: There is a way to check if a value is an array. (Ask ChatGPT)
You can just return false or throw an error if `array` is NOT an array, instead of just check for "string". There are other types of value that is not array.

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,31 @@
...
+        let object = {};
+        for (let i=0; i < copy.length; i++){
+            let count = 0;
+            for (let j=0; j < array.length; j++){
+                if (array[j] == copy[i]){
+                    count = count + 1;
+                }
+            }
+            object[copy[i]] = count;
+        }
```
cjyuan, line 22: Would you try implement the following approach (which is more efficient by using object as a lookup table):
```
  let object = {};
  for each element, ele, in the arrays
    if object[ele] is undefined
      set object[ele] to 1
    else 
      increment object[ele] by 1
```

Sprint-2/implement/tally.test.js
```diff
@@ -19,16 +19,30 @@ const tally = require("./tally.js");
...
+    expect(tally(['a', 'a', 'b', 'b', 'c', 'c'])).toEqual({ a : 2, b: 2, c: 2 });
+});
+
 
 // Given an invalid input like a string
 // When passed to tally
 // Then it should throw an error
+it("Given an invalid input it should throw an error", () => {
+    expect(tally('string')).toEqual(false);
+});
```
cjyuan, line 48: Your function is not throwing any error when the parameter is not an array (the description of the test is not quite correct).

Sprint-2/stretch/count-words.js
```diff
@@ -26,3 +26,31 @@
...
+
+function countWords(string){
+  let countedWords = {};
+  let noPunctuation= string.replace(/[.,\/#!$%\^&\*;:{}=\-_`~()]/g,"");
+  //remove unnecessary spaces
+  noPunctuation = noPunctuation.replace(/\s{2,}/g," ");
+  noPunctuation = noPunctuation.toLowerCase();
+  const words = noPunctuation.split(" ");
+  for(i=0; i<words.length; i++){
+    if (Object.hasOwn(countedWords, words[i])){
```
cjyuan, line 38: We could also write `if (countWords[words[i]])`, which can be interpreted as "if `countWords[words[i]]` has a value" (and we know that it cannot be zero).

cjyuan (commented):
I left you some comments. Feel free to mark this as "Complete" after you have made the changes accordingly.

PR #499, opened 2025-04-17
ITP_JAN25 | KATARZYNA_KAZIMIERCZUK | MODULE_DATA_GROUPS | SPRINT2

katarzynakaz:
Self checklist

## Changelist

Briefly explain your PR.
Finished Sprint 2.

## Questions
No questions.

### DIFF
--- a/Sprint-2/debug/author.js
+++ b/Sprint-2/debug/author.js
@@ -14,8 +14,8 @@ const author = {
 // for (const value of author) {
 //   console.log(value);
 // }
-for (const value in author) {
-  console.log(value, author[value]);
+for (const key in author) {
+  console.log(key, author[key]);
 }
 
 // it attempts to loop over the object author and log the values of all properties
--- a/Sprint-2/debug/author.js
+++ b/Sprint-2/debug/author.js
@@ -11,6 +11,11 @@ const author = {
   alive: true,
 };
 
-for (const value of author) {
-  console.log(value);
+// for (const value of author) {
+//   console.log(value);
+// }
+for (const value in author) {
+  console.log(value, author[value]);
 }
+
+// it attempts to loop over the object author and log the values of all properties
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,19 +1,20 @@
 const contains = (object, prop) => {
-    // check if the object is empty
-    const isEmpty = (object) => { 
-        return Object.keys(object).length === 0; 
-    }
+  //additional test
+  //null, undfined and array check
+  if (object === null || object === undefined || object instanceof Array) {
+    return false;
+  }
 
-    if (isEmpty(object)) {
-        return false;
-    }
+  // check if the object is empty
+  if (Object.keys(object).length === 0) {
+    return false;
+  }
 
-    if (object.hasOwnProperty(prop)) {
-        return true; 
-    } else {
-        return false;
-    }
+  if (object.hasOwnProperty(prop)) {
+    return true;
+  } else {
+    return false;
+  }
 };
-
 
 module.exports = contains;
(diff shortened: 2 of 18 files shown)


Sprint-2/debug/author.js
```diff
@@ -11,6 +11,11 @@ const author = {
...
 
-for (const value of author) {
-  console.log(value);
+// for (const value of author) {
+//   console.log(value);
+// }
+for (const value in author) {
+  console.log(value, author[value]);
 }
+
```
cjyuan, line 20: Since the variable named `value` actually takes the property name or key of the `author` object in the for-in loop, it would be better to name the variable `key` than `value`.

Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,19 @@
-function contains() {}
+const contains = (object, prop) => {
+    // check if the object is empty
+    const isEmpty = (object) => { 
+        return Object.keys(object).length === 0; 
+    }
+
+    if (isEmpty(object)) {
+        return false;
+    }
```
cjyuan, line 9: Why not just write 

```
  if (Object.keys(object).length === 0)
    return false; 
```

Also, if an object is empty, `object.hasOwnProperty(prop)` would also return `false`.

Sprint-2/implement/contains.test.js
```diff
@@ -12,24 +12,44 @@ as the object doesn't contains a key of 'c'
...
+    expect(contains(cat, 'age')).toEqual(false)
+})
 
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+test('given invalid parameters, it returns false', () => {
+    expect(contains([])).toEqual(false)
+})
+})
```
cjyuan, line 55: Would you also include the following tests to check if your function can return false when the first parameter is invalid?
```
    expect(contains([5, 5, 5]), 0).toEqual(false)
    expect(contains(null, 'key')).toEqual(false)
    expect(contains(1234, 'key')).toEqual(false)
    expect(contains(undefined, 'key')).toEqual(false)
```

Note: Array is a kind of object in JS where indexes serve as keys to an array.

Sprint-2/implement/querystring.js
```diff
@@ -1,16 +1,41 @@
...
+    //split into key and the secondPart
+
+    const [key, ...secondPart] = pair.split("=");
+    // if additional '=' connect for the value
+    queryParamsReturned[key] = secondPart.join("=");
+
+    // so instead I split into 2 parts always
+    // const [key, value] = pair.split("=", 2);
+    // queryParamsReturned[key] = value;
+  });
```
cjyuan, line 36: Please note that in real querystring, both `key` and `value` are ***percent-encoded*** or ***URL encoded*** in the URL. For example, the string "5%" will be encoded as "5%25". So to get the actual value of "5%25" (whether it is a key or value in the querystring), you should call a function to decode it.

cjyuan (commented):
Code is generally good. I have left you some suggestions.

This branch should not contains modified files in the Sprint-1 folder. Can you fix the branch? (One way to fix the branch is to replace the files in Sprint-1 folder by the same files in your `main` branch).

Sprint-2/implement/querystring.js
katarzynakaz: > Please note that in real querystring, both `key` and `value` are _**percent-encoded**_ or _**URL encoded**_ in the URL. For example, the string "5%" will be encoded as "5%25". So to get the actual value of "5%25" (whether it is a key or value in the querystring), you should call a function to decode it. May I suggest looking up any of these terms, and "How to decode URL encoded string in JS"?

thank you I looked it up, there is a decodeURIComponent, I shall add it to both key and value

PR #503, opened 2025-04-19
ITP JAN 25 | Katarzyna Kazimierczuk | Data Groups | Sprint 3 - alarmclock

katarzynakaz:
<!--

You must title your PR like this:

Self checklist

## Changelist

Alarm clock completed

## Questions

No questions

### DIFF
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -1,4 +1,143 @@
-function setAlarm() {}
+const userInput = document.getElementById("alarmSet").value;
+
+//split function
+const splitTime = (userInput) => {
+  return userInput.split(":");
+};
+
+//parse time to a format we can work with
+const splitAndParseTime = (userInput) => {
+  const split = splitTime(userInput);
+
+  let hours = 0;
+  let minutes = 0;
+  let seconds = 0;
+
+  if (split.length >= 1) hours = parseInt(split[0]);
+  if (split.length >= 2) minutes = parseInt(split[1]);
+  if (split.length === 3) seconds = parseInt(split[2]);
+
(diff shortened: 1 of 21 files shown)


Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,143 @@
-function setAlarm() {}
+const userInput = document.getElementById("alarmSet").value;
+
+//split function
+const splitTime = (userInput) => {
+  return userInput.split(":");
+};
+
+//parse time to a format we can work with
+const splitAndParseTime = (userInput) => {
```
shieldo, line 9: I'm not sure that the requirements here were to include an hours column - it looks like it's just minutes and seconds, in an MM:SS format.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,143 @@
...
+    // return it
+    // return `${hours}:${minutes}:${seconds}`;
+    return true;
+  }
+  alert("Please enter a valid time.");
+  return false; // wrong time input
+};
+
+// format single digit time and return as an object
+const formattedTime = (hours, minutes, seconds) => {
```
shieldo, line 56: This is a reasonable way to model a time given how we are modelling it.

shieldo (commented):
You've managed to get the timer counting down to an alarm, so well done! There are a few things missing, such as following the requirement that entering `10` turns into `00:10` (i.e. counts down ten seconds). Also it would be good to be able to cancel existing alarms that are set - i.e. if you set a new alarm it should cancel any existing alarm. It does work OK in the simple case though (if I type in `00:00:10` as the input).

PR #506, opened 2025-04-20
ITP -Northwest- JAN 2025 | Jan Lo | Module-Data groups | SPRINT 2 | WEEK 7

JanLlw:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,3 +1,24 @@
-function contains() {}
+function contains(element, key) {
+    const result = element[key];
 
+    // console.log({
+    //     key,
+    //     typeKey: typeof key,
+
+    //     result,
+    //     typeResult: typeof result
+    // })
+
+    // if undefined
+      // return false as doesnt exist
+    // else, does exist!
+      // return true
+
+    if (result === undefined) return false;
+    else return true;
+    
+}
+console.log(contains({a: 1, b: 2}, 'a'))
+console.log(contains({a: 1, b: 2}, 'c'))
+console.log(contains({firstName: "Cam" }, 'firstName'))
 module.exports = contains;
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,5 +1,6 @@
 function contains(element, key) {
-    const result = element[key];
+ 
+  const result = element[key];
 
     // console.log({
     //     key,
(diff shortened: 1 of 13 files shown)


Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,24 @@
...
+
+    // if undefined
+      // return false as doesnt exist
+    // else, does exist!
+      // return true
+
+    if (result === undefined) return false;
+    else return true;
+    
+}
```
cjyuan, line 20: This approach is not bulletproof because a property can be assigned an `undefined` value. For example,
```
  let obj { x: undefined };  // An object with a property named "x", and its value is undefined.
```

Can you figure out how else to improve the code?

Sprint-2/implement/contains.test.js
```diff
@@ -20,16 +20,26 @@ as the object doesn't contains a key of 'c'
...
+test("contains with a non-existent property name", () => {
+    expect(contains({a: 1, b: 2}, 'c')).toBe (false);
+});
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+test("invalid parameters", () => {
+    expect(contains(["apple", "orange", "pear"])).toBe (false);
+});
+
```
cjyuan, line 45: Invalid parameters could also be other types of values. For examples, `null`, `"some string"`, `123`, `undefined`.

With your current implementation, `contains(["apple", "orange", "pear"], "0")` and `contains("ABC", "0") would return `true`.

You would need to add code to your function to check the first parameter is a valid object and is not an array.

Note: array is a kind of objects in JS.

Sprint-2/implement/querystring.js
```diff
@@ -6,7 +6,8 @@ function parseQueryString(queryString) {
   const keyValuePairs = queryString.split("&");
 
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
+    const [key, ...valueParts] = pair.split("="); 
+    const value = valueParts.join("="); 
     queryParams[key] = value;
```
cjyuan, line 11: Your approach works. 

Please note that in real querystring, both `key` and `value` are ***percent-encoded*** or ***URL encoded*** in the URL. For example, the string "5%" will be encoded as "5%25". So to get the actual value of "5%25" (whether it is a key or value in the querystring), you should call a function to decode it.

cjyuan:
Changes look good.

cjyuan:
It is good practice to respond to each reviewer's comment to acknowledge their feedback and to explain any changes made in response.

image

cjyuan:
It's also a good practice to give a brief description to each PR.

image

PR #507, opened 2025-04-20
ITP-Jan | NW | Jovy So | Module-Data-Groups | WEEK 1

jovyso:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -3,9 +3,19 @@
 // If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
 
 function calculateMedian(list) {
+  // const oddMiddleIndex = Math.floor(list.length / 2);
+  // const oddMedian = list.splice(oddMiddleIndex, 1)[0];
   const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
-}
+  // const evenMiddleIndex = list.length / 2;
+  // const evenMedian = ((list.splice(evenMiddleIndex -1 ,1)[0]) + (list.splice(evenMiddleIndex ,1)[0])) / 2;
+  if (list.length % 2 === 1) return middleIndex ;
+  else return (list[middleIndex - 1] + list[middleIndex]) / 2;
+};
+
 
 module.exports = calculateMedian;
+
+console.log(calculateMedian([1, 2, 3, 4, 5]));
+console.log(calculateMedian([1, 2, 3, 4, 5, 6]));
+console.log(calculateMedian([1, 2, 3, 4]));
+console.log(calculateMedian([1, 2, 3]));
--- a/Sprint-1/implement/dedupe.js
+++ b/Sprint-1/implement/dedupe.js
@@ -1,1 +1,22 @@
-function dedupe() {}
+function dedupe(elements) {
+  if (elements.length === 0) return [];
+  else if (elements.length === 1) return elements;
+
+  for (let i = elements.length -1; i >= 0; i--) {
+    for (let j = 0; j < i; j++) {
+      if (elements[j] === elements[i]) {
+        elements.splice(i, 1);
+        break;
+      }
+    }
+  }
+  return elements;
+};
+
+module.exports = dedupe;
+
+console.log(dedupe([]));
+console.log(dedupe(["a", "b", "c"]));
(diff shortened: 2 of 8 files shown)


Sprint-1/fix/median.js
```diff
@@ -3,9 +3,19 @@
...
   const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
-}
+  // const evenMiddleIndex = list.length / 2;
+  // const evenMedian = ((list.splice(evenMiddleIndex -1 ,1)[0]) + (list.splice(evenMiddleIndex ,1)[0])) / 2;
+  if (list.length % 2 === 1) return middleIndex ;
+  else return (list[middleIndex - 1] + list[middleIndex]) / 2;
+};
+
```
cjyuan, line 14: The arrays specified in `median.test.js` are already sorted by coincidence.

Typically, to find the median value of an array, we first need to sort it. However, to prevent modifying `list` directly, it's best to work with a cloned version instead.

Could you add one or more statements to create a cloned copy of `list` and then sort the numbers in the cloned array?

Sprint-1/implement/dedupe.js
```diff
@@ -1 +1,22 @@
...
+  for (let i = elements.length -1; i >= 0; i--) {
+    for (let j = 0; j < i; j++) {
+      if (elements[j] === elements[i]) {
+        elements.splice(i, 1);
+        break;
+      }
+    }
+  }
+  return elements;
+};
```
cjyuan, line 14: You can also consider using built-in functions to improve performance. 
Some of the built-in functions/methods you can consider are `Array.prototype.indexOf()` and `Array.prototype.includes()`.

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,20 @@
-function findMax(elements) {
-}
+function findMax(elements) {  
+    let numberOnly = elements.filter(item => typeof item === "number");
+    // let nonNumber = elements.filter(item => typeof item !== "number");
+    if (elements.length === 0) return -Infinity;
+    else if (elements.length === 1 && elements[0] > 0) return elements[0];
```
cjyuan, line 5: With this condition, `findMax(["10"])` would return `"10"`, `findMax([true])` would return `true`.

Sprint-1/implement/max.js
cjyuan, line 1: You can also treat "an array with only non-number values" as an array of no numbers or an empty array.

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,41 @@ const sum = require("./sum.js");
...
+test("given an array containing negative number, returns correct total sum", () => {
+  expect(sum([-1, 1, -2, 3])).toEqual(1);
+});
 
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
+test("given an array with decimal/float number, returns correct total sum", () => {
+  expect(sum([1.1, 2.2, 3, 4.5])).toEqual(10.8);
+});
```

So the following could happen
```

```

Sprint-1/implement/max.test.js
```diff
@@ -16,28 +16,50 @@ const findMax = require("./max.js");
...
+  expect(findMax(["Apple", true, 10, 98])).toEqual(98);
+});
+
 
 // Given an array with only non-number values
 // When passed to the max function
 // Then it should return the least surprising value given how it behaves for all other inputs
+test("given an array with only non-number values, returns largest number", () => {
+  expect(findMax([true, "Apple", false, "Banana"])).toEqual(NaN);
+});
```
cjyuan, line 65: What do you mean by "the largest number"? `NaN` is "Not a number".

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,24 @@
...
+  else if (elements.length === 1) return elements[0];
+
+  let sumOfElements = 0;
+  for (let i = 0; i < elements.length; i++) {
+    if (typeof elements[i] === "number") {
+    sumOfElements += elements[i];
+    }
+  }
+  return sumOfElements; 
+};
```
cjyuan, line 13: You can also consider filter out all non-number values first, and then just work on the numbers in the filtered array.

cjyuan (commented):
Code is generally good.

I left you some suggestions.

cjyuan:
It is a good practice to leave a brief description in every PR.

image

PR #515, opened 2025-04-22
ITP London_Jan 2025| Samunta Sunuwar | Module Data Groups | Sprint 3| Slideshow

Samunta:

### DIFF
--- a/Sprint-3/slideshow/index.html
+++ b/Sprint-3/slideshow/index.html
@@ -13,10 +13,11 @@
       <p id="image-count">Image 1 of 3</p>
     </section>
     <section id="buttons">
-      <button type="button" id="auto-forward-btn">Auto Forward</button>
+      <button type="button" id="auto-forward">Auto Forward</button>
       <button type="button" id="backward-btn">Backwards</button>
+      <button id="stop">Stop</button>
       <button type="button" id="forward-btn">Forward</button>
-      <button type="button" id="auto-backward-btn">Auto Backward</button>
+      <button type="button" id="auto-backward">Auto Backward</button>
     </section>
   </body>
 </html>
--- a/Sprint-3/slideshow/index.html
+++ b/Sprint-3/slideshow/index.html
@@ -10,6 +10,7 @@
   <body>
     <section id="images">
       <img id="carousel-img" src="./assets/cute-cat-a.png" alt="cat-pic" />
+      <p id="image-count">Image 1 of 3</p>
     </section>
     <section id="buttons">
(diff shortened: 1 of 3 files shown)


Sprint-3/slideshow/index.html
```diff
@@ -3,12 +3,21 @@
...
+    <section id="images">
+      <img id="carousel-img" src="./assets/cute-cat-a.png" alt="cat-pic" />
+      <p id="image-count">Image 1 of 3</p>
+    </section>
+    <section id="buttons">
+      <button type="button" id="auto-forward">Auto Forward</button>
+      <button type="button" id="backward-btn">Backwards</button>
+      <button id="stop">Stop</button>
+      <button type="button" id="forward-btn">Forward</button>
+      <button type="button" id="auto-backward">Auto Backward</button>
```
illicitonion, line 20: I found it a little surprising that "Auto Backwards" is in the "Forwards" direction end of the buttons, and "Auto Forwards" was in the "Backwards" direction end of the buttons. Is there a reason you ordered them like this? Would it perhaps be more clear if you swapped them?

Sprint-3/slideshow/slideshow.js
```diff
@@ -1,8 +1,63 @@
...
+  autoBackwardBtn.disabled = true;
+}
+
+// Auto Backward logic
+function startAutoBackward() {
+  stopSlideshow();
+  slideshowInterval = setInterval(goBackward, 2000);
+  autoForwardBtn.disabled = true;
+  autoBackwardBtn.disabled = true;
+}
```
illicitonion, line 50: These two functions seem to mostly do the same thing.

Imagine if we decided we also wanted to disable the "Forward" and "Backward" buttons when in auto mode, and _only_ allow the "Stop" button to be pressed. We would need to make the same change to both of these functions. Maybe we wouldn't remember this, and would only change one, and have a bug.

Can you think of a way to avoid this, so that we re-use this code across both functions, rather than having two copies of it?

Sprint-3/slideshow/slideshow.js
```diff
@@ -1,8 +1,63 @@
...
+  "./assets/cute-cat-b.jpg",
+  "./assets/cute-cat-c.jpg",
 ];
 
+// Write your code here
 
-// Write your code here
\ No newline at end of file
+let currentIndex = 0;
+const carouselImg = document.getElementById("carousel-img");
```
illicitonion, line 10: In general your naming in this PR is really good. The variables are really clear what they are and what they're for, the functions have really clear names - I can tell what they do from just their names, etc. Good job!

PR #519, opened 2025-06-10
ZA | ITP-May-2025 | Rashaad Ebrahim | Module Data Groups | Sprint 1

Rashaad-Ebrahim:
## Learners, PR Template

Self checklist

### Sprint 1 Coursework complete
* Solutions provided for exercises and necessary explanations provided

## Questions

No questions at this time.

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -17,7 +17,7 @@ function calculateMedian(list) {
     .sort((a, b) => a - b);
 
   // Throw out empty arrays
-  if (sortedList.length < 2) {
+  if (sortedList.length < 1) {
     return null;
   }
 
@@ -30,3 +30,5 @@ function calculateMedian(list) {
 }
 
 module.exports = calculateMedian;
+
+console.log(typeof calculateMedian([5]))
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -6,8 +6,26 @@
 // or 'list' has mixed values (the function is expected to sort only numbers).
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
+  // Throw out invalid arrays
+  if (!Array.isArray(list)) {
+    return null;
+  }
+
+  // Filter and sort numbers only
+  const sortedList = list
+    .filter((item) => typeof item === "number" && !isNaN(item))
+    .sort((a, b) => a - b);
+
+  // Throw out empty arrays
+  if (sortedList.length < 2) {
+    return null;
+  }
+
+  const middleIndex = Math.floor(sortedList.length / 2);
+  const median =
+    sortedList.length % 2 === 0
+      ? (sortedList[middleIndex - 1] + sortedList[middleIndex]) / 2
+      : sortedList[middleIndex];
   return median;
 }
 
--- a/Sprint-1/implement/dedupe.test.js
+++ b/Sprint-1/implement/dedupe.test.js
@@ -13,15 +13,41 @@ E.g. dedupe([1, 2, 1]) target output: [1, 2]
 
 // Acceptance Criteria:
 
-// Given an empty array
-// When passed to the dedupe function
-// Then it should return an empty array
-test.todo("given an empty array, it returns an empty array");
+describe("dedup Tests", () => {
+  // Given an empty array
+  // When passed to the dedupe function
+  // Then it should return an empty array
+  test("Given an empty array, it returns an empty array", () => {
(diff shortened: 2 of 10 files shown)


chinar-amrutkar:
@chinar-amrutkar

Sprint-1/fix/median.js
```diff
@@ -6,8 +6,26 @@
...
+    return null;
+  }
+
+  // Filter and sort numbers only
+  const sortedList = list
+    .filter((item) => typeof item === "number" && !isNaN(item))
+    .sort((a, b) => a - b);
+
+  // Throw out empty arrays
+  if (sortedList.length < 2) {
```
illicitonion, line 20: This is an interesting check - how did you decide `sortedList.length < 2` is a good check here? Why not 3, or 1?

Is there a test we  could add to the test suite to show we've done the right thing?

Sprint-1/implement/dedupe.test.js
```diff
@@ -13,15 +13,41 @@ E.g. dedupe([1, 2, 1]) target output: [1, 2]
...
+    const expected = [];
+    expect(input).toEqual(expected);
+  });
 
-// Given an array with no duplicates
-// When passed to the dedupe function
-// Then it should return a copy of the original array
+  // Given an array with no duplicates
+  // When passed to the dedupe function
+  // Then it should return a copy of the original array
```
illicitonion, line 28: The requirement here says returns a _copy_. But this test doesn't seem to test that a copy is returned - if you just returned the original array, I think this test would still pass.

Can you think of a way of verifying we actually returned a copy? Maybe break your implementation to return the original array, and use that to test your test, before fixing your code again.

Sprint-1/implement/dedupe.test.js
```diff
@@ -13,15 +13,41 @@ E.g. dedupe([1, 2, 1]) target output: [1, 2]
...
+  // When passed to the dedupe function
+  // Then it should remove the duplicate values, preserving the first occurrence of each element
+  test.each([
+    [
+      ["a", "a", "a", "b", "b", "c"],
+      ["a", "b", "c"],
+    ],
+    [
+      [5, 1, 1, 2, 3, 2, 5, 8],
+      [5, 1, 2, 3, 8],
```
illicitonion, line 45: This is a nice test making sure it returns the _first_ copy!

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,13 @@
 function sum(elements) {
+  const numericalElements = [];
+  for (const element of elements) {
+    if (typeof element === "number") {
+      numericalElements.push(element);
+    }
+  }
+  return numericalElements.reduce((a, c) => a + c, 0);
```
illicitonion, line 8: This works, but I think `reduce` is often quite hard to understand and follow. Can you think of a way of implementing this function which _doesn't_ require accumulating an array of values and then using `reduce`?

Sprint-1/fix/median.js
Rashaad-Ebrahim: I agree, it should really be `sortedList.length < 1`. I've checked and the function can handle arrays with only 1 entry.

I've updated the function and added a test to check sing entry arrays.

Sprint-1/implement/dedupe.test.js
Rashaad-Ebrahim: I've updated my testing to do a check but I'm not sure its exactly what you meant.

Sprint-1/implement/dedupe.test.js
```diff
@@ -13,15 +13,41 @@ E.g. dedupe([1, 2, 1]) target output: [1, 2]
...
-// Given an array with no duplicates
-// When passed to the dedupe function
-// Then it should return a copy of the original array
+  // Given an array with no duplicates
+  // When passed to the dedupe function
+  // Then it should return a copy of the original array
+  test("Given an array with no duplicates, it returns a copy of the original array", () => {
+    const input = dedupe([7, 2, 8, 9]);
+    const expected = [7, 2, 8, 9];
+    expect(input === expected).toEqual(false);
```
illicitonion, line 32: This does now check that the returned array isn't the same as the passed one (though `expect(input).not.toBe(expected)` may be more clear/idiomatic as its error message will be better on failure), but _doesn't_ check that the contents are the same any more. You may need two assertions to check both thing :)

Sprint-1/implement/sum.js
illicitonion: This is simpler, but you're still ending up accumulating an array of values. Can you think of a way to not need a second array at all?

illicitonion (commented):
Good improvements, congratulations! A couple more things, then you're good to go! :)

PR #530, opened 2025-06-26
ZA | May-2025 | Malusi Skunyana | Sprint-1

MalusiS:
## Learners, PR Template

Self checklist

## Changelist

Implemented functions, wrote tests, refactored implementation, and solved the stretch problems.

## Questions


Sprint-1/implement/max.test.js
```diff
@@ -16,28 +16,48 @@ const findMax = require("./max.js");
...
 // Then it should return the closest one to zero
+test("given an array with just negative numbers, it returns the closest one to zero", () => {
+  expect(findMax([-10, -20, -30, -5])).toBe(-5);
+});
 
 // Given an array with decimal numbers
 // When passed to the max function
 // Then it should return the largest decimal number
+test("given an array with decimal numbers, it returns the largest decimal number", () => {
+  expect(findMax([1.5, 2.5, 3.5])).toBe(3.5);
```
Waldo-Strydom, line 48: Decimal numbers in most programming languages (including JS) are internally represented in "floating point number" format. Floating point arithmetic is not exact. For example, the result of 46.5678 - 46 === 0.5678 is false because 46.5678 - 46 only yield a value that is very close to 0.5678.

So expect(sum([1.1, 1.1, 1.1])).toEqual(3.3); would probably fail.

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,41 @@ const sum = require("./sum.js");
...
 // Then it should still return the correct total sum
+test("sum([-5, -10, -15]) should return -30", () => {
+  expect(sum([-5, -10, -15])).toBe(-30);
+});
 
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
+test("sum([1.5, 2.5, 3.5]) should return 7.5", () => {
+  expect(sum([1.5, 2.5, 3.5])).toBe(7.5);
```
Waldo-Strydom, line 38: See my last comment

Sprint-1/implement/max.test.js
MalusiS: I updated the test for an array with decimal numbers to use `toBeCloseTo` rather than `toBe`.

PR #535, opened 2025-06-30
London | ITP-May-2025 | Mo Muchunu | Module-Data-Groups | Sprint 1

Mo-Muchunu:
## Learners, PR Template

Self checklist

## Changelist

-   Fixed broken function to pass all test cases in fix directory.

-  Implemented functions and test suites for 'max', 'sum' and 'dedupe' according to requirements.

-   Refactored existing implementations to meet requirements and ensured all tests still pass.

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -5,10 +5,27 @@
 // Hint: Please consider scenarios when 'list' doesn't have numbers (the function is expected to return null)
 // or 'list' has mixed values (the function is expected to sort only numbers).
 
+
+
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  // Validate input: must be an array with at least one number
+  if (!Array.isArray(list)) return null; // Returns null for non-arrays.
+
+  const numericValues = list.filter( // Keeps only numbers, filtering out non-numeric values: empty slots, null, undefined, strings.
+    item => typeof item === "number" && !isNaN(item) // Ensures all elements are numbers, and not NaN.
+  );
+
+  if (numericValues.length === 0) return null; // Returns null if no numeric values left.
+
+  // Using spread operator (...list) to create copy of original list without modifying it, then sort copied array in ascending order.
+  const sorted = [...numericValues].sort((a, b) => a - b); // Ensures numeric based sorting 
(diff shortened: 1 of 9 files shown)


Sprint-1/fix/median.js
```diff
@@ -5,10 +5,27 @@
...
+
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  // Validate input: must be an array with at least one number
+  if (!Array.isArray(list)) return null; // Returns null for non-arrays.
+
+  const numericValues = list.filter( // Keeps only numbers, filtering out non-numeric values: empty slots, null, undefined, strings.
+    item => typeof item === "number" && !isNaN(item) // Ensures all elements are numbers, and not NaN.
```
cjyuan, line 15: `Infinity` is also a value of type number. Do you consider `[-Infinity, Infinity]` a valid input.

Sprint-1/implement/sum.test.js
```diff
@@ -6,31 +6,50 @@ E.g. sum([10, 20, 30]), target output: 60
...
+    test('given an array containing negative numbers, it should still return the correct total sum', () => {
+        expect(sum([-5, 7, -12, -6, 1, 3, -20])).toBe(-32);
+    });
+
+    // Given an array with decimal/float numbers
+    // When passed to the sum function
+    // Then it should return the correct total sum
+    test('given an array with decimal/float numbers, it should return the correct total sum', () => {
+        expect(sum([ 1.4, 1.75, 2.51, 3.523])).toBe(9.183);
+    });
```

So the following could happen
```

```

Suggestion: Look up 

cjyuan (commented):
Code is very pretty solid.

I just have a few suggestions.

Sprint-1/implement/sum.test.js
Mo-Muchunu: After looking up how to check equality for operations involving floating point numbers in JavaScript and with Jest, I can see that the use of `toBe` expects strict equality. As you mentioned floating point arithmetic can result in small rounding errors- causing tests to fail even when the values are logically the same.

I realised it’s better to use the `toBeCloseTo()` method which allows you to compare numbers within a small margin of error. This is because you can set a default precision point to be close within a specified number of decimal places, making it better for floating point calculations.

I've corrected the function and added a test case.

Mo-Muchunu:
> Well done!

Once again CJ- thank you for your time!

PR #537, opened 2025-06-30
LONDON | MAY-2025 | JESUS DEL MORAL | SPRINT-1 MODULE DATA-GROUPS

delmorallopez:
PR sprint-1 Module Data Groups

Including 2 commits 
- Fix -Implement -Refactor - Stretch

### DIFF
--- a/Sprint-1/implement/max.js
+++ b/Sprint-1/implement/max.js
@@ -4,7 +4,7 @@ function findMax(elements) {
   }
 
   // Filter only numbers
-  const numericElements = elements.filter(el => typeof el === "number");
+  const numericElements = elements.filter(el => Number.isFinite(el));
 
   if (numericElements.length === 0) {
     return -Infinity;


Sprint-1/implement/max.test.js
```diff
@@ -16,28 +16,73 @@ const findMax = require("./max.js");
...
+  });
+
 // Given an array with only non-number values
 // When passed to the max function
 // Then it should return the least surprising value given how it behaves for all other inputs
+  it("Given an array with only non-number values, return the least surprising value given how it behaves for all other inputs", () => {
+    const list = ["London", "Hello", "CYF" , "Zebra"];
+    const result = findMax(list);
+    expect(result).toEqual(-Infinity);
+  });
```
cjyuan, line 86: Do the following function calls return the value you expected?
```
findMax(["A"]);
findMax(["A", 99]);
```

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,17 @@
 function sum(elements) {
+    if (elements.length < 1){
+        return 0;
+    } else if (elements.length == 1){
+        return elements;
+    }
```
cjyuan, line 6: What if `elements` is `["A"]`?

Could consider using the code at lines 8-14 to handle array of any size; no need special treatments for empty array or array with one element.

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,54 @@ const sum = require("./sum.js");
...
+  });
 
 // Given an array containing non-number values
 // When passed to the sum function
 // Then it should ignore the non-numerical values and return the sum of the numerical elements
+it("Given an array containing non-number values", () => {
+    const list = [1, 2.2, -5.2, 5, "Jesus", null];
+    const result = sum(list);
+    expect(result).toEqual(3);  
+  });
```
cjyuan, line 57: Decimal numbers in most programming languages (including JS) are internally represented in "floating point number" format. Floating point arithmetic is not exact. For example, the result of `46.5678 - 46 === 0.5678` is false because `46.5678 - 46` only yield a value that is very close to `0.5678`. Even changing the order in which the program add/subtract numbers can yield different values.

So the following could happen
```

```

Suggestion: Look up 

Sprint-1/implement/max.js
```diff
@@ -1,21 +1,18 @@
...
-      return max;
-    } else return -Infinity;
+  if (elements.length < 1) {
+    return -Infinity;
   }
-  module.exports = findMax;
-  
+
+  // Filter only numbers
+  const numericElements = elements.filter(el => typeof el === "number");
```
cjyuan, line 7: `NaN` is a value of type "number" and can pass this check. You could consider excluding `NaN` in your computation.

cjyuan (commented):
- Changes look good. Great job.

- Did you share your code with other trainees? I am asking this because I came across similar test scripts submitted by another trainee.

- Leaving responses directly in the comment threads makes tracking the discussion easier. You can try this approach in future PRs.
image

Here is a simplified version of best practices ChatGPT suggested for responding to inline comments in a pull request:
> - ✅ Reply to every comment – Let the reviewer know you saw it.
> - ✏️ Make the change if needed – Fix the code if the comment points out a real issue.
> - 🤔 Explain if you don't agree – If you think the code is fine, politely explain why.
> - ✅ Mark as resolved when done – Only mark comments resolved after you fix or respond.
> - 💬 Keep replies short and polite – Be respectful and to the point.
> - ⏱️ Respond soon – Don’t wait too long to reply.
> - 🧪 Test your changes – Make sure your fixes actually work.
> - 📍 Reply directly under the comment – This keeps the conversation easy to follow.

delmorallopez:
@cjyuan 
I didn't share the code with anyone, just I did with my colleague Hakan Murat Kavut (hmkavut@gmail.com) some exercises together

PR #539, opened 2025-07-01
LONDON | MAY-2025 | HAKAN MURAT KAVUT | MODULE DATA-GROUPS SPRINT-1 

hmkavut:
## Learners, PR Template

Self checklist

- [x] have titled my PR with REGION | COHORT_NAME | FIRST_NAME LAST_NAME | PROJ_NAME

## Changelist

Sprint 1 exercises are completed.

## Questions

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -2,13 +2,30 @@
 // Start by running the tests for this function
 // If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
 
-// Hint: Please consider scenarios when 'list' doesn't have numbers (the function is expected to return null)
-// or 'list' has mixed values (the function is expected to sort only numbers).
-
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  if (!Array.isArray(list)) return null;
+  if (list.length == 0) return null;
+
+  let numeric = false;
+
+  for (let i = 0; i < list.length; i++) {
+    if (typeof list[i] === "number") {
+      numeric = true;
+    }
+  }
+  if (!numeric) return null;
+
+  const numbers = list.filter((n) => typeof n === "number");
+
+  numbers.sort((a, b) => a - b);
+
+  const middleIndex = Math.floor(numbers.length / 2);
+
+  if (numbers.length % 2 === 0) {
+    return (numbers[middleIndex - 1] + numbers[middleIndex]) / 2;
+  } else {
+    return numbers[middleIndex];
+  }
 }
 
 module.exports = calculateMedian;
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -6,16 +6,8 @@ function calculateMedian(list) {
   if (!Array.isArray(list)) return null;
   if (list.length == 0) return null;
 
-  let numeric = false;
-
-  for (let i = 0; i < list.length; i++) {
-    if (typeof list[i] === "number") {
-      numeric = true;
-    }
-  }
-  if (!numeric) return null;
-
   const numbers = list.filter((n) => typeof n === "number");
+  if (numbers.length==0) return null;
 
   numbers.sort((a, b) => a - b);
 
--- a/Sprint-1/implement/dedupe.test.js
+++ b/Sprint-1/implement/dedupe.test.js
@@ -31,7 +31,7 @@ describe("dedupe arrays  ", () => {
     { input: ["a", "b", "c"], expected: ["a", "b", "c"] },
     { input: [5, 1, 2, 3, 8], expected: [5, 1, 2, 3, 8] },
   ].forEach(({ input, expected }) =>
-    it(`returns Dedupe Array with no duplicate]`, () => expect(dedupe(input)).toEqual(expected))
+    it(`returns Dedupe Array with no duplicate]`, () => expect(dedupe(input)).not.toBe(input))
   );
 // Given an array with strings or numbers
 // When passed to the dedupe function
--- a/Sprint-1/implement/dedupe.test.js
+++ b/Sprint-1/implement/dedupe.test.js
@@ -22,7 +22,7 @@ describe("dedupe arrays  ", () => {
     { input: "it is not array", expected: null },
     { input: null, expected: null },
   ].forEach(({ input, expected }) =>
-    it(`returns empty or null ]`, () => expect(dedupe(input)).toEqual(expected))
+    it(`dedupe should returns empty or null `, () => expect(dedupe(input)).toEqual(expected))
   );
 // Given an array with no duplicates
 // When passed to the dedupe function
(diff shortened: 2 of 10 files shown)


Sprint-1/fix/median.js
```diff
@@ -2,13 +2,30 @@
...
+  let numeric = false;
+
+  for (let i = 0; i < list.length; i++) {
+    if (typeof list[i] === "number") {
+      numeric = true;
+    }
+  }
+  if (!numeric) return null;
+
+  const numbers = list.filter((n) => typeof n === "number");
```
AdnanGondal, line 18: You are using a for loop once to check for numbers, and again doing a filter of the numbers. The first for-loop is not needed if you combine these right?

After you have filtered the numbers, you can just use if (numbers.length===0) to check if the list contained any numbers.

Sprint-1/refactor/includes.test.js
AdnanGondal, line 1: For the refactor challenge, you should leave the test file as it is, and instead only change the includes.js file as required by the instructions.

AdnanGondal (changes requested):
A good attempt - but please address the comments and I will review it again.

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,23 @@
...
+  for (let i = 0; i < elements.length; i++) {
+    if (typeof elements[i] === "number") {
+      sum += elements[i];
+      flag = true;
+    }
+  }
+  if (flag) {
+    return sum;
+  } else return -Infinity;
 }
```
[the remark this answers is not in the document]
hmkavut: I modified with adding number control.

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,55 @@ const sum = require("./sum.js");
 // Given an empty array
 // When passed to the sum function
 // Then it should return 0
-test.todo("given an empty array, returns 0")
+describe("find Max", () => {
```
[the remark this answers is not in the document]
hmkavut: I updated.

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,55 @@ const sum = require("./sum.js");
...
-// When passed to the sum function
-// Then it should ignore the non-numerical values and return the sum of the numerical elements
+  // Given an array containing non-number values
+  // When passed to the sum function
+  // Then it should ignore the non-numerical values and return the sum of the numerical elements
+  it("Given an array with non-number values, return the sum of the numerical elements", () => {
+    const list = [-10.05, 82.15, "1000", "London", -2.85, true];
+    const result = sum(list);
+    expect(result).toBeCloseTo(69.25, 2);
+  });
```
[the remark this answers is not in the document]
hmkavut: I added sum = parseFloat(sum.toFixed(5));

Sprint-1/refactor/includes.test.js
hmkavut: I misunderstood :) I modified includes.js and removed changes at test file

hmkavut:
Thank you for detailed review.
Sorry for late update. I modified my PR as your comments.

Sprint-1/implement/dedupe.test.js
```diff
@@ -18,10 +18,37 @@ E.g. dedupe([1, 2, 1]) target output: [1, 2]
...
+
 // Given an array with no duplicates
 // When passed to the dedupe function
 // Then it should return a copy of the original array
+describe("calculateMedian", () => {
+  [
+    { input: ["a", "b", "c"], expected: ["a", "b", "c"] },
+    { input: [5, 1, 2, 3, 8], expected: [5, 1, 2, 3, 8] },
+  ].forEach(({ input, expected }) =>
+    it(`returns Dedupe Array ]`, () => expect(dedupe(input)).toEqual(expected))
```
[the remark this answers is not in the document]
cjyuan: What I meant was, preparing the test description in the following manner:

```javascript
 ].forEach(({ input, expected }) =>
    it(`unique elements: dedupe of ${input} should return a copy of the array`, () =>  ... 
```
Note: When `expect(X).toEqual(Y)` fails, jest will show the value of X and Y so embedding Y in the test description is optional.

---
I forgot to mention the following in my previous review:
The current test code cannot check if the returned array is a **copy** of the original array because `toEqual()` compares objects (including arrays) by value. To illustrate,

```
  const A = [2, 3, 1];
    
```

In order to check if the returned array is a **copy** of the original array, we would need an **additional** check.
Can you find out what code we need to add in order to ensure the returned value is not the original array?

Sprint-1/implement/max.js
```diff
@@ -1,22 +1,24 @@
 function findMax(elements) {
-  //First test should return -Infinity
-  if (elements.length < 1) {
+  
+  if (elements.length == 0 || !Array.isArray(elements)) {
     return -Infinity;
   }
+
```
cjyuan: The checking order matters. if `elements` were `null` or `undefined`, then `elements.length == 0` would have thrown an error.

However, if we wrote the condition as `( !Array.isArray(elements) || elements.length == 0)`, then by the time the 2nd operand is evaluated, `elements` is guaranteed to be an array).

Sprint-1/implement/max.js
```diff
@@ -1,22 +1,24 @@
...
   }
 
-  let max = elements[0];
-  for (let i = 0; i < elements.length; i++) {
-    if (elements[i] > max && typeof elements[i] === "number") {
-      max = elements[i];
+  let max = numbers[0];
+  for (let i = 0; i < numbers.length; i++) {
+    if (numbers[i] > max) {
+      max = numbers[i];
```
cjyuan, line 18: With this approach, since we have already assumed the first element is the max, we could start the loop with `i=1` instead of `i=0`. That is, we do not need to compare max to the first element).

Sprint-1/implement/dedupe.test.js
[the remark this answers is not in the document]
hmkavut: I found not.tobe function and I updated the test method.

Sprint-1/implement/sum.js
```diff
@@ -1,23 +1,26 @@
...
-  for (let i = 0; i < elements.length; i++) {
-    if (typeof elements[i] === "number") {
-      sum += elements[i];
+  for (let i = 0; i < numbers.length; i++) {
+    if (typeof numbers[i] === "number") {
+      sum += numbers[i];
       flag = true;
     }
   }
+  sum = parseFloat(sum.toFixed(5));
```
cjyuan, line 19: Doing such conversion would make the calculated result less accurate.

Rather than adjusting the implementation to match the test, we should handle the conversion within the test itself. Jest’s `.toBeCloseTo()` matcher is designed for this purpose and allows for comparing floating-point values with precision.

Sprint-1/implement/sum.js
```diff
@@ -1,23 +1,26 @@
...
   if (flag) {
     return sum;
-  } else return -Infinity;
+  } 
+  else return -Infinity;
```
cjyuan, line 23: Why not consider an array with no numbers the same as array with 0 elements, and return 0 in both cases?

Sprint-1/implement/sum.js
```diff
@@ -1,23 +1,26 @@
...
+    return numbers[0];
   }
 
   let sum = 0;
   let flag = false;
-  for (let i = 0; i < elements.length; i++) {
-    if (typeof elements[i] === "number") {
-      sum += elements[i];
+  for (let i = 0; i < numbers.length; i++) {
+    if (typeof numbers[i] === "number") {
```
cjyuan, line 14: If you want to exclude `Infinity` and `NaN` from the calculation, you can consider using `Number.isFinite()`

cjyuan (commented):
Changes were good. I have few more suggestions.

Sprint-1/implement/sum.js
hmkavut: I removed them because I check number elements in line 5

cjyuan (commented):
Changes look good.

PR #541, opened 2025-07-01
WESTMIDLANDS| ITP-MAY-2025|SARATAHIR|SPRINT-1/COURSEWORK

SaraTahir28:
## Learners, PR Template

Self checklist

- [X]  I have tested my changes
- [X] My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/contributing/)

## Changelist

Updated functions and tests. All tests are passing and mandatory coursework is done 

## Questions

### DIFF
--- a/Sprint-1/implement/dedupe.test.js
+++ b/Sprint-1/implement/dedupe.test.js
@@ -16,12 +16,22 @@ E.g. dedupe([1, 2, 1]) target output: [1, 2]
 // Given an empty array
 // When passed to the dedupe function
 // Then it should return an empty array
-test.todo("given an empty array, it returns an empty array");
+test("given an empty array, it returns an empty array",()=>{
+    const result = dedupe([])
+    expect(result).toEqual([])
+})
 
 // Given an array with no duplicates
 // When passed to the dedupe function
 // Then it should return a copy of the original array
-
+test("given an array with no duplicates, it returns copy of the array",()=>{
+    const result = dedupe([1,2,3,4,5])
+    expect(result).toEqual([1,2,3,4,5])
+})
 // Given an array with strings or numbers
 // When passed to the dedupe function
 // Then it should remove the duplicate values, preserving the first occurence of each element
+test("given an array with strings or numbers, it removes duplicate values and preserve first occurence",()=>{
+    const result = dedupe([1,1,2,4,3,4,5,3,5])
+    expect(result).toEqual([1,2,4,3,5])
+})
--- a/Sprint-1/implement/dedupe.test.js
+++ b/Sprint-1/implement/dedupe.test.js
@@ -1,3 +1,4 @@
+
 const dedupe = require("./dedupe.js");
 /*
 Dedupe Array
@@ -32,6 +33,8 @@ test("given an array with no duplicates, it returns copy of the array",()=>{
 // When passed to the dedupe function
 // Then it should remove the duplicate values, preserving the first occurence of each element
 test("given an array with strings or numbers, it removes duplicate values and preserve first occurence",()=>{
-    const result = dedupe([1,1,2,4,3,4,5,3,5])
-    expect(result).toEqual([1,2,4,3,5])
-})
+   let result1 = dedupe([1,1,2,4,3,4,5,3,5])
+    expect(result1).toEqual([1,2,4,3,5])
+    let result2 = dedupe([1, "a", 1, 2, "b", "a", 4, 3, "b", 4, 5, 3, 5]);
+    expect(result2).toEqual([1, "a", 2, "b", 4, 3, 5]);
+});
--- a/Sprint-1/implement/max.js
+++ b/Sprint-1/implement/max.js
@@ -1,4 +1,17 @@
-function findMax(elements) {
+function findMax(list) {
+    const numericList = list.filter(
+    item => typeof item === "number" && !Number.isNaN(item))
+
+    if(list.length === 0){
+        return -Infinity
+  }
+    if(numericList.length === 0){
+        return null
+  }
+    else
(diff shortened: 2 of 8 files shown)


Sprint-1/implement/max.js
```diff
@@ -1,4 +1,17 @@
...
+function findMax(list) {
+    const numericList = list.filter(
+    item => typeof item === "number" && !Number.isNaN(item))
+
+    if(list.length === 0){
+        return -Infinity
+  }
+    if(numericList.length === 0){
+        return null
+  }
```
cjyuan, line 10: Could consider treating array with no numeric values the same as empty array (which also does not have any numeric values). 

Additionally, it's good practice to design functions so that they consistently return values of the same type.

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,14 @@
...
+function sum(list) {
+  if (list.length === 0) {
+    return 0;
+  }
+
+  const numericList = list.filter(item => typeof item === "number" && !Number.isNaN(item));
 
+  if (numericList.length === 0) {
+    return null;
+  }
```
cjyuan, line 10: Could also consider treating array with no numeric values the same as empty array (which also does not have any numeric values).

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,45 @@ const sum = require("./sum.js");
...
+    expect(result).toBe(11);
+})
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
-
+test("given an array containing float numbers ,return correct total sum", ()=>{
+    const result = sum([2.5,4.6,3.5,2.2]);
+    expect(result).toBe(12.8);
+})
```
cjyuan, line 43: Decimal numbers in most programming languages (including JS) are internally represented in "floating point number" format. Floating point arithmetic is not exact. For example, the result of `46.5678 - 46 === 0.5678` is false because `46.5678 - 46` only yield a value that is very close to `0.5678`. Even changing the order in which the program add/subtract numbers can yield different values.

So the following could happen
```

```

Suggestion: Look up 

Sprint-1/refactor/includes.js
```diff
@@ -1,13 +1,22 @@
...
 
 module.exports = includes;
+function includes (list, target){
+  if (!Array.isArray(list) && list.length=== 0) return []
+  if(list.includes(target)){
+  return true
+}
+  else {
+  return false
+}
```
cjyuan, line 21: - This exercise expects one to rewrite the code using a for...of loop.
 
- The current implementation does not behave the same as the original function; the original function doesn't return an empty array when `list` is empty.

Sprint-1/implement/dedupe.test.js
```diff
@@ -16,12 +16,22 @@ E.g. dedupe([1, 2, 1]) target output: [1, 2]
...
 // Then it should return a copy of the original array
-
+test("given an array with no duplicates, it returns copy of the array",()=>{
+    const result = dedupe([1,2,3,4,5])
+    expect(result).toEqual([1,2,3,4,5])
+})
 // Given an array with strings or numbers
 // When passed to the dedupe function
 // Then it should remove the duplicate values, preserving the first occurence of each element
+test("given an array with strings or numbers, it removes duplicate values and preserve first occurence",()=>{
```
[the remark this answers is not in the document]
SaraTahir28: Added a test value with string inputs

Sprint-1/refactor/includes.js
SaraTahir28: Refactored the code with a for loop.

Sprint-1/implement/sum.test.js
SaraTahir28: yes switched toBe with toBeCloseTo

Sprint-1/implement/max.test.js
```diff
@@ -16,9 +16,9 @@ const findMax = require("./max.js");
 // When passed to the max function
 // Then it should return -Infinity
 // Delete this test.todo and replace it with a test.
-test("given an empty array, returns -Infinity",()=>{
+test("given an empty array, returns null",()=>{
     const result = findMax([]);
-    expect(result).toBe(-Infinity);
+    expect(result).toBe(null);
```
cjyuan, line 21: The test specification is:
```
// Given an empty array
```

Sprint-1/refactor/includes.js
```diff
@@ -11,12 +11,10 @@
...
 module.exports = includes;
-function includes (list, target){
-  if (!Array.isArray(list) && list.length=== 0) return []
-  if(list.includes(target)){
-  return true
-}
-  else {
-  return false
-}
+function includes(list, target = undefined) {
```
cjyuan, line 14: When an argument is omitted, parameter is assigned `undefined` by default, there is no need to explicitly set `undefined` as the default value.

Sprint-1/implement/max.test.js
SaraTahir28: I dont understand what to do When I try returning infinity for this. I have to seperate the numeric list vs list=== 0 to pass my tests. I tried trying to understand through Chat Gpt but it gives me explanations beyong the scope of my understanding where I should return infinity for all cases. I dont und

cjyuan: Case 1 (Empty array): `[]` => return `-Infiity`
Case 2: (Array with no numbers): `["A", "B"]` => Filter out non-numeric elements => `[]` (Same as case 1).

In general, we should avoid designing a function to return different **types** of value. In the case of `max()`, it is normally expected to return a number, so we should avoid returning `null` from the function.

PR #543, opened 2025-07-01
London | ITP-May-2025 | Adnaan Abo | Module-Data-Groups | Sprint 1

AdnaanA:
Self checklist

## Changelist

Briefly explain your PR.
Created test cases with jest and implemented test and made sure it all  passed

## Questions

### DIFF
--- a/Sprint-1/implement/sum.js
+++ b/Sprint-1/implement/sum.js
@@ -1,22 +1,7 @@
 function sum(elements) {
-    if (elements.length === 0) {
-        return 0; // Return 0 for an empty array
-    }
-
-    // Filter out non-numeric elements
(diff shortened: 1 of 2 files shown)


Sprint-1/implement/sum.js
AdnanGondal, line 1: This method works great - but in live code bases it is preferred to avoid variable mutation and take a more functional approach. 
Perhaps you can use array methods like filter and reduce to achieve the same effect?

AdnaanA: I have changed the code to use the filter and reduce methods and the test are still passing.

AdnanGondal:
Great submission - marking as Complete

PR #552, opened 2025-07-05
London | ITP-May-2025 | Mo Muchunu | Module-Data-Groups | Sprint 2

Mo-Muchunu:
## Learners, PR Template

Self checklist

## Changelist

 - Debugged files by predicting, verifying, and fixing bugs to ensure correct code execution.

- Implemented required functions with comprehensive tests covering various input scenarios.

- Interpreted complex programs, researched unfamiliar syntax, and used logging to understand and validate behavior.

### DIFF
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -17,19 +17,21 @@ as the object doesn't contains a key of 'c'
 // When passed an object and a property name
 // Then it should return true if the object contains the property, false otherwise
 
-// Given an empty object
-// When passed to contains
-// Then it should return false
-test.todo("contains on empty object returns false");
+it("contains an empty object, returns false", () => {
+  expect(contains({}, "a")).toBe(false);
+});
 
-// Given an object with properties
-// When passed to contains with an existing property name
-// Then it should return true
 
-// Given an object with properties
-// When passed to contains with a non-existent property name
-// Then it should return false
+it("contains an object with existing property, returns true", () => {
+  expect(contains({ a: 1, b: 2 }, "a")).toBe(true);
+});
 
-// Given invalid parameters like an array
-// When passed to contains
-// Then it should return false or throw an error
+
+it("contains an object with non-existent property, returns false", () => {
+  expect(contains({ a: 1, b: 2 }, "c")).toBe(false);
+});
+
+
+it("contains invalid parameters, returns false", () => {
+  expect(contains(null, 42)).toBe(false);
+});
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -17,21 +17,22 @@ as the object doesn't contains a key of 'c'
 // When passed an object and a property name
 // Then it should return true if the object contains the property, false otherwise
 
-it("contains an empty object, returns false", () => {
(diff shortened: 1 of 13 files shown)


Sprint-2/implement/contains.test.js
```diff
@@ -17,19 +17,21 @@ as the object doesn't contains a key of 'c'
...
+
+
+it("contains an object with non-existent property, returns false", () => {
+  expect(contains({ a: 1, b: 2 }, "c")).toBe(false);
+});
+
+
+it("contains invalid parameters, returns false", () => {
+  expect(contains(null, 42)).toBe(false);
+});
```
cjyuan, line 37: To clearly indicate a name refers to a function, we could use the notation `contains()`. 

Alternatively, we could prepare the tests in the following way:
```
// Mention the function once in "describe"
describe('Testing contains()', () => {

  // No need to mention the function name in each test description
  test('given an empty object, returns false', () => {
    // test code here
  });

  test('given an object with existing property, returns true', () => {
    // more test code
  });

  ...
});

```

Sprint-2/implement/querystring.js
```diff
@@ -1,16 +1,16 @@
...
-    const [key, value] = pair.split("=");
-    queryParams[key] = value;
-  }
+  return Object.fromEntries(                                    // Using Object.fromEntries to convert array of [key, value] pairs into an object
+    query.split("&").map((pair) => {                            // Split the query string into individual key-value pairs
+
+
+      const [key, ...rest] = pair.split("=");                   // Split each pair at the first '='
+      const value = rest.length > 0 ? rest.join("=") : "";      // If at least one after the first '=', join values with '=' as separator or default to empty string
+      return [key, value];
```
cjyuan, line 10: Please note that in real query string, both `key` and `value` are ***percent-encoded*** or ***URL encoded*** in the URL. For example, the string "5%" will be encoded as "5%25". So to get the actual value of "5%25" (whether it is a key or value in the querystring), you should call a function to decode it.

cjyuan (commented):
- Code is solid. Good job. 

- Branch is not clean. Can you revert the changes made in "Sprint-1" folder?

Sprint-2/implement/querystring.js
Mo-Muchunu: I looked into how query string encoding works and added `decodeURIComponent` to ensure keys and values get properly decoded. I'll bear this in mind moving forward.

Mo-Muchunu:
> Changes look good.
> 
> I wonder how you "rebase" your commits. It seems like the commits on this branch were newly created after my review.

@cjyuan 

Thank you! 

Well I used the command: `git rebase --onto main Sprint-1 Sprint-2` which let me re-apply only the commits that were unique to Sprint-2, skipping over the Sprint-1 commits that were present on the branch. 

Then, I ran: `git log Sprint-2 -- Sprint-1` to double-check if there were any overlapping commits, and since it returned nothing, I confirmed with `git log --oneline` and there are now no commit conflicts between the two branches.

`git log Sprint-2` shows my original commit history but you're right, the way I corrected the branch did create new commit hashes, so even though the content is the same, it may appear as if these are fresh commits post-review. 

That's probably why the PR timeline looks a bit different now.

PR #557, opened 2025-07-05
London | ITP-May-2025 | Jamal Laqdiem| Module-Data-Groups | Sprint 1 

jemaljemy:
## Learners, PR Template

Self checklist

- [ ] I have titled my PR with REGION | COHORT_NAME | FIRST_NAME LAST_NAME | PROJ_NAME

## Changelist

Briefly explain your PR.

## Questions


Sprint-1/fix/median.js
fearcyf, line 1: Shouldn't there be a test.js file with test and expect statements..?

Sprint-1/implement/dedupe.test.js
fearcyf, line 1: Tests passed, well done. Could your approach be more succinct? Research test.each

Sprint-1/implement/max.test.js
fearcyf, line 1: Tests passed, well done

Sprint-1/refactor/includes.js
fearcyf, line 1: Code works fine, just one consideration, do you really need lines 4-6. It's not an error just consider it.

Sprint-1/fix/median.js
```diff
@@ -6,9 +6,40 @@
...
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  if (!Array.isArray(list) || list.length ===0) {
+    return null
+  }
+  const onlyNumbers = list.filter(item => typeof item  === "number" &&  !isNaN(item))
+  if (onlyNumbers.length === 0) {
+    return null
+  }
+  const sortedList = [...onlyNumbers].sort((a,b) => a - b)
```
illicitonion, line 16: Why do you do the `[...onlyNumbers]` here instead of just using `onlyNumbers`? What problem are you avoiding with it?

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,27 @@
 function findMax(elements) {
+  if (elements.length === 0) {
```
illicitonion, line 2: What are you guarding against with this `if` statement? If you removed it, what would break in your code?

illicitonion (commented):
This is generally looking good, but sometimes a bit more complicated than it could be. Please take a look at my comments and update your code or reply to them with thoughts!

illicitonion:
Sorry, I can't see any answers - is it possible you have some comments pending, or some code changes you haven't pushed to your remote?

I see these comments without answers:

Screenshot 2025-07-21 at 11 22 35

And this is the only code I see having changed since I left those comments, which only answers one of the comments:

Screenshot 2025-07-21 at 11 23 01

illicitonion (commented):
I'm glad we could work that out, thanks! Just one comment left to work through :)

Sprint-1/implement/max.js
jemaljemy: Thanks for reviewing, after examining the code again i can clearly notice that the if condition in the line 2 is redundant, as the for loop handle that case and return -infinity.

illicitonion: Sounds good! If you push a change fixing this, we can call this PR complete :)

PR #558, opened 2025-07-07
London | ITP-May-2025 | Mohamed Ibrahim | Module-Data-Groups | Sprint-2

sidicamoli:
## Learners, PR Template

Self checklist

## Changelist

A range of solutions has been provided for the exercises, as well as the necessary explanations.

## Questions

### DIFF
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -38,20 +38,12 @@ test("Should return false for an empty object", () => {
 // Given an object with properties
 // When passed to contains with an existing property name
 // Then it should return true
-test("Should return true when passed with an existing property name", () => {
-  const currentOutput = contains({ z: 26, y: 25, x: 24 }, "x");
-  const targetOutput = true;
-  expect(currentOutput).toBe(targetOutput);
-});
+
 
 // Given an object with properties
 // When passed to contains with a non-existent property name
 // Then it should return false
-test("Should return false when passed with a non-existing property name", () => {
-  const currentOutput = contains({ w: 23, v: 22, u: 21 }, "x");
-  const targetOutput = false;
-  expect(currentOutput).toBe(targetOutput);
-});
+
 
 // Given invalid parameters like an array
 // When passed to contains
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -11,25 +11,54 @@ E.g. contains({a: 1, b: 2}, 'c') // returns false
 as the object doesn't contains a key of 'c'
 */
 
-// Acceptance criteria:
-
 // Given a contains function
 // When passed an object and a property name
 // Then it should return true if the object contains the property, false otherwise
+test("Should return true if the object contains the property", () => {
+  const currentOutput = contains({ a: 1, b: 2 }, "a");
+  const targetOutput = true;
+  expect(currentOutput).toBe(targetOutput);
+});
+
+test("Should return false if the object doesn't contain the property", () => {
+  const currentOutput = contains({ a: 1, b: 2 }, "c");
+  const targetOutput = false;
+  expect(currentOutput).toBe(targetOutput);
+});
 
 // Given an empty object
 // When passed to contains
 // Then it should return false
-test.todo("contains on empty object returns false");
+test("Should return false for an empty object", () => {
+  const currentOutput = contains({}, "b");
+  const targetOutput = false;
+  expect(currentOutput).toBe(targetOutput);
+});
 
 // Given an object with properties
 // When passed to contains with an existing property name
 // Then it should return true
+test("Should return true when passed with an existing property name", () => {
+  const currentOutput = contains({ z: 26, y: 25, x: 24 }, "x");
+  const targetOutput = true;
+  expect(currentOutput).toBe(targetOutput);
+});
 
 // Given an object with properties
 // When passed to contains with a non-existent property name
 // Then it should return false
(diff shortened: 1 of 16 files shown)


Sprint-2/implement/contains.test.js
```diff
@@ -11,25 +11,54 @@ E.g. contains({a: 1, b: 2}, 'c') // returns false
...
+});
 
 // Given an object with properties
 // When passed to contains with a non-existent property name
 // Then it should return false
+test("Should return false when passed with a non-existing property name", () => {
+  const currentOutput = contains({ w: 23, v: 22, u: 21 }, "x");
+  const targetOutput = false;
+  expect(currentOutput).toBe(targetOutput);
+});
```
cjyuan, line 54: These two tests are similar to those at lines 17-27.

Sprint-2/implement/contains.test.js
```diff
@@ -11,25 +11,54 @@ E.g. contains({a: 1, b: 2}, 'c') // returns false
...
 
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+test("Should return false when passed with an invalid property", () => {
+  const currentOutput = contains({ g: 7, h: 8, i: 9 }, []);
+  const targetOutput = false;
+  expect(currentOutput).toBe(targetOutput);
+});
+
```
cjyuan, line 64: Does you function return what you expect from the following function calls?
```
contains(["A", "B"], "0");
contains(null, "0");
contains(undefined, "0");

```

Sprint-2/implement/querystring.js
```diff
@@ -1,16 +1,18 @@
...
 
-  for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
-    queryParams[key] = value;
+  const pairs = queryString.split("&");
+
+  for (const pair of pairs) {
+    const [key, ...rest] = pair.split("=");
+    const value = rest.join("="); // Handles multiple '=' signs
+    params[key] = value;
```
cjyuan, line 11: Please note that in real querystring, both `key` and `value` are ***percent-encoded*** or ***URL encoded*** in the URL. For example, the string "5%" will be encoded as "5%25". So to get the actual value of "5%25" (whether it is a key or value in the querystring), you should call a function to decode it.

Sprint-2/implement/tally.test.js
```diff
@@ -1,34 +1,26 @@
...
+  expect(tally(["x"])).toEqual({ x: 1 });
+});
 
-// Given a function called tally
-// When passed an array of items
-// Then it should return an object containing the count for each unique item
+// Test: array with repeated items
+test("Counts how many times each item appears", () => {
+  expect(tally(["a", "a", "b", "a", "b", "c"])).toEqual({ a: 3, b: 2, c: 1 });
+});
```
cjyuan, line 16: Can you check if your function returns what you expect from the following function call?
```
  tally(["toString", "toString", "toString"]); 
```

Sprint-2/stretch/count-words.js
```diff
@@ -1,28 +1,41 @@
...
+    } else {
+      result[word] = 1;
+    }
+  }
+
+  // Sort the result by count descending
+  const sorted = Object.entries(result).sort(([, a], [, b]) => b - a);
+
+  return Object.fromEntries(sorted);
+}
```
cjyuan, line 37: Can you check if your function returns what you expect from the following function calls?

```
```

cjyuan:
May I suggest asking ChatGPT the following:

>In JS, what are the differences among the following ways to check if `obj` has a property specified in a variable named `key`?
`obj[key]`
`key in obj`
`Object.hasOwn(obj, key)`

Sprint-2/implement/contains.test.js
sidicamoli: Thanks for the feedback! I've removed the duplicate tests as suggested.

Sprint-2/implement/tally.test.js
sidicamoli: Thanks for the feedback! I handled edge case for 'toString' key and added a test to confirm correct behavior.

Sprint-2/stretch/count-words.js
sidicamoli: Thanks! Updated countWords to handle edge cases — now passes all tests.

sidicamoli:
> May I suggest asking ChatGPT the following:
> 
> > In JS, what are the differences among the following ways to check if `obj` has a property specified in a variable named `key`?
> > `obj[key]`
> > `key in obj`
> > `Object.hasOwn(obj, key)`

✅ obj[key] – Accesses value
Gets the value, not ideal for checking existence.

Fails if value is falsy (0, false, '', undefined, etc.).

✅ Use to read value, not check existence.

✅ "key" in obj – Checks anywhere in prototype chain
Returns true if key exists anywhere, including inherited properties.

⚠️ Might give true for things like "toString" from Object.prototype.

✅ Use if you care about access, not ownership.

✅ Object.hasOwn(obj, key) – Best & safest check
Returns true only if the key is directly on the object (not inherited).

Handles falsy values correctly.

✅ Most robust and preferred method for existence checks.

✅ In Summary:
Use obj[key] → to get a value.

Use key in obj → to check if property exists anywhere (own + inherited).

Use Object.hasOwn(obj, key) → to check if property exists directly on the object.
🔒 Most reliable and safest

cjyuan (commented):
Changes look great. Well done!

PR #561, opened 2025-07-08
WM | May-2025 | Abdullah Saleh | Module-Data-Groups | Sprint1

3bdullah-saleh:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -6,9 +6,32 @@
 // or 'list' has mixed values (the function is expected to sort only numbers).
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
-}
+  // Check if the list is an array and not empty
+  if (!Array.isArray(list) || list.length === 0) {
+    return null; 
+  }
+  // Filter out non-numeric values (NaN, undefined, etc.) 
+  const numbersOnly = list.filter(item => typeof item === 'number')
 
+  // If no numeric values remain after filtering, return null
+  if (numbersOnly.length === 0) {
+    return null;
+  }
+
+  // Sort the filtered array in ascending order
+  numbersOnly.sort((a, b) => (a - b))
+  const middleIndex = Math.floor(numbersOnly.length / 2);
+
+  // If odd length, return the middle element
+  if (numbersOnly.length % 2 !== 0) {
+    return numbersOnly[middleIndex]
+  }
+  // If even length, return average of two middle elements
+  else {
+    const middle1 = numbersOnly[middleIndex -1];
+    const middle2 = numbersOnly[middleIndex];
+    return (middle1 + middle2) / 2;
+  }
+
+}
 module.exports = calculateMedian;
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -8,10 +8,11 @@
 function calculateMedian(list) {
   // Check if the list is an array and not empty
   if (!Array.isArray(list) || list.length === 0) {
-    return null; 
+    return null;
   }
-  // Filter out non-numeric values (NaN, undefined, etc.) 
-  const numbersOnly = list.filter(item => typeof item === 'number')
+
+  // Filter out non-numeric values (NaN, undefined, etc.)
+  const numbersOnly = list.filter((item) => Number.isFinite(item));
 
   // If no numeric values remain after filtering, return null
   if (numbersOnly.length === 0) {
@@ -19,19 +20,18 @@ function calculateMedian(list) {
   }
 
   // Sort the filtered array in ascending order
-  numbersOnly.sort((a, b) => (a - b))
+  numbersOnly.sort((a, b) => a - b);
   const middleIndex = Math.floor(numbersOnly.length / 2);
 
   // If odd length, return the middle element
   if (numbersOnly.length % 2 !== 0) {
-    return numbersOnly[middleIndex]
+    return numbersOnly[middleIndex];
   }
   // If even length, return average of two middle elements
   else {
-    const middle1 = numbersOnly[middleIndex -1];
+    const middle1 = numbersOnly[middleIndex - 1];
     const middle2 = numbersOnly[middleIndex];
     return (middle1 + middle2) / 2;
   }
-
 }
 module.exports = calculateMedian;
--- a/Sprint-1/implement/dedupe.js
+++ b/Sprint-1/implement/dedupe.js
@@ -1,1 +1,19 @@
-function dedupe() {}
+function dedupe(array) {
+    let arrayCopy = [];
+    if (array.length == 0) {
+        return [];
+    }
+    else if (array.length != 0) {
+        array.forEach(element => {
+            if (!arrayCopy.includes(element)) {
+                arrayCopy.push(element);    
+        }
+    });
+        return arrayCopy;
+    }
+
+
+}
+
+console.log(dedupe(["a", "b", "c"]))
+module.exports = dedupe;
--- a/Sprint-1/implement/dedupe.js
+++ b/Sprint-1/implement/dedupe.js
@@ -1,19 +1,16 @@
 function dedupe(array) {
-    let arrayCopy = [];
-    if (array.length == 0) {
-        return [];
-    }
-    else if (array.length != 0) {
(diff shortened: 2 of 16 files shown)


Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,50 @@ const sum = require("./sum.js");
...
 
 // Given an array with only non-number values
 // When passed to the sum function
 // Then it should return the least surprising value given how it behaves for all other inputs
+
+test("Given an array with only non-number values, should return Infinity", () => {
+    const input = ["2.4", "hello", null, undefined];
+    const output = Infinity
+    expect(sum(input)).toEqual(output)
+});
```
cjyuan, line 62: Can you check if your function returns the expected values in the following function calls?

```
sum(["A"]);
sum([-Infinity, 1, 2, 3]);
sum([-Infinity, Infinity]);
sum([Infinity, 1, 2, 3]);
```

Sprint-1/fix/median.js
```diff
@@ -6,9 +6,32 @@
...
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
-}
+  // Check if the list is an array and not empty
+  if (!Array.isArray(list) || list.length === 0) {
+    return null; 
+  }
+  // Filter out non-numeric values (NaN, undefined, etc.) 
+  const numbersOnly = list.filter(item => typeof item === 'number')
```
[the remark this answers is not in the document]
3bdullah-saleh: Thanks for your feedback, I made the changes.

Sprint-1/implement/dedupe.test.js
```diff
@@ -17,11 +17,24 @@ E.g. dedupe([1, 2, 1]) target output: [1, 2]
...
+});
 
 // Given an array with no duplicates
 // When passed to the dedupe function
 // Then it should return a copy of the original array
-
+test.todo("Given an array with no duplicates, it return a copy of the original array");
+test("Given an array with no duplicates, it return a copy of the original array", () => {
+    expect(dedupe(["a", "b", "c"])).toEqual(["a", "b", "c"]);
+});
```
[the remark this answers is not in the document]
3bdullah-saleh: I missed that by saying a `copy`  means we need to check that the reference is difference. I update the test compare the reference of the returned array with the input array. Thank you for brining this to my attention.

Sprint-1/implement/max.test.js
```diff
@@ -16,28 +16,62 @@ const findMax = require("./max.js");
...
+});
 
 // Given an array with only non-number values
 // When passed to the max function
 // Then it should return the least surprising value given how it behaves for all other inputs
+test("Given an array with only non-number values, should return Infinity", () => {
+    const input = ['hey', "10", 'hi', "60", "10"]
+    const output = Infinity
+    expect(findMax(input)).toEqual(output)
+});
```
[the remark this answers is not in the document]
3bdullah-saleh: Thanks for pointing that out! Initially, the test was passing, and I assumed it was due to the if (condition) block. However, after reviewing it closely, I realized it was actually passing because of the Math.max return. I've now refactored the code to properly handle the intended cases and ensure the logic is correct.

Sprint-1/implement/sum.test.js
3bdullah-saleh: I refactor the code and It returns:
Infinity
6
Infinity
6
which is the expected output.

Sprint-1/implement/dedupe.js
```diff
@@ -1,19 +1,16 @@
...
+        arrayCopy.push(element);
+      }
     });
-        return arrayCopy;
-    }
-
-
+    return arrayCopy;
+  }
 }
```
cjyuan, line 13: Because the if block on lines 3-5 has a return statement, we can omit the else on line 5. Further more, the condition at line 5 `array.length != 0` is the opposite of the condition on line 3; if one of them is `true`, the other will be `false`.
So we could write the code as

```javascript
if (array.length == 0)
  return [];
  
// If execution can reach this point, then (array.length == 0)  must be false.
array.forEach(
   ...
); 

return arrayCopy; 
```

Sprint-1/implement/dedupe.test.js
```diff
@@ -16,25 +16,22 @@ E.g. dedupe([1, 2, 1]) target output: [1, 2]
...
 // Then it should return a copy of the original array
-test.todo("Given an array with no duplicates, it return a copy of the original array");
-test("Given an array with no duplicates, it return a copy of the original array", () => {
-    expect(dedupe(["a", "b", "c"])).toEqual(["a", "b", "c"]);
+test("Given an array with no duplicates, it returns a copy of the original array", () => {
+    const input = ["a", "b", "c"];
+    const output = dedupe(input);
+    expect(input).not.toBe(output);
+    expect(input).toEqual(output);
 });
```
cjyuan, line 30: I think it is more natural to express
```
    expect(output).not.toBe(input);
    expect(output).toEqual(input);
   //  expect( computedResult ).toEqual( expectedResult );
```
instead of the other way around.

Sprint-1/implement/max.js
```diff
@@ -1,17 +1,15 @@
...
+ */
 
-    }
+function findMax(elements) {
+    // Function will return '-Infinity', If the array is empty or has only non numeric value.
+    // If the array has only one element, function will return that element as the largest value.
+    // If the array has only negative value, function will return the closest number to zero.
+    // Otherwise, function will return the largest value.
+    const numbersOnly = elements.filter(element => Number.isFinite(element));
+    return  Math.max(...numbersOnly);
```
cjyuan, line 11: This code does not quite return `-Infinity` when `numbersOnly` is empty.

Sprint-1/implement/sum.test.js
```diff
@@ -6,49 +6,49 @@ E.g. sum([10, 20, 30]), target output: 60
...
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
 test("Given an array with decimal/float numbers, should return the correct total sum", () => {
     const input = [1.5, 2.4, 3.5];
-    const output = 7.4
-    expect(sum(input)).toEqual(output)
+    const result = sum(input);
+    const expected = 7.4;
+    expect(closeEnough(result, expected)).toBe(true);
```
cjyuan, line 42: We could also use the Jest built-in `toBeCloseTo()` method.

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,18 @@
...
+    if (elements.length == 0) {
+        return 0;
+    }
+    else if (elements.length == 1) {
+        return sum[0];
+    }
+    else if (elements.length > 1) {
+        const numbersOnly = elements.filter(element => typeof element === 'number');
+        if (numbersOnly.length == 0) {
+            return Infinity;
```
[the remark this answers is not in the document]
cjyuan: The definition of "least surprising value" is subjective.
I just have a different candidate for such a value but yours is also acceptable.

I wonder what AI will recommend. :D

Sprint-1/implement/sum.test.js
cjyuan: In JS, 
```
  -Infinity + 1 + 2 + 3 yields -Infinity
  -Infinity + Infinity yields NaN
   Infinity + 1 + 2 + 3 yields Infinity
```

Shouldn't `sum()` behave in a similar fashion?

Sprint-1/implement/dedupe.js
3bdullah-saleh: You're right — the else is unnecessary since the if block already returns. Removing it makes the code cleaner and easier to follow. Thanks for the suggestion!

Sprint-1/implement/dedupe.test.js
3bdullah-saleh: Oh right, that was just a typo! I usually use the expected pattern, so thanks for catching it.

Sprint-1/implement/max.js
cjyuan: > This code does not quite return `-Infinity` when `numbersOnly` is empty.

This comment has not yet been addressed.

3bdullah-saleh: I’ve updated the code to explicitly check for an empty numeric array and return -Infinity before calling Math.max.

PR #568, opened 2025-07-09
LONDON | MAY-2025 | HAKAN MURAT KAVUT | MODULE DATA-GROUPS SPRINT-2

hmkavut:
## Learners, PR Template

Self checklist

## Changelist

In this PR, I implemented function in Sprint 2 folder and test cases are written for these functions

## Questions

### DIFF
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,3 +1,8 @@
-function contains() {}
+function contains(list, prop) {
+  if (typeof list !== "object" || list === null || Array.isArray(list))
+    return false;
+  if (list[prop] == null) return false;
+  return true;
+}
 
 module.exports = contains;
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,8 +1,9 @@
 function contains(list, prop) {
-  if (typeof list !== "object" || list === null || Array.isArray(list))
-    return false;
-  if (list[prop] == null) return false;
-  return true;
+  return (
+    typeof list === "object" &&
+    !Array.isArray(list) &&
+    list[prop] != null
+  );
 }
 
 module.exports = contains;
--- a/Sprint-2/implement/lookup.js
+++ b/Sprint-2/implement/lookup.js
@@ -1,5 +1,11 @@
-function createLookup() {
-  // implementation here
+function createLookup(countryCurrencyPairs) {
+  const queryParams = {};
+  for (let i = 0; i < countryCurrencyPairs.length; i++) {
+    const key = countryCurrencyPairs[i][0];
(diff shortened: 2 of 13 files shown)


Sprint-2/implement/lookup.js
```diff
@@ -1,5 +1,11 @@
-function createLookup() {
-  // implementation here
+function createLookup(countryCurrencyPairs) {
+  const queryParams = {};
+  for (let i = 0; i < countryCurrencyPairs.length; i++) {
```
day-lee, line 3: Good job! It works as intended. As a challenge, would you like to try to refactor the code using for...of loop? It's up to you.

Sprint-2/implement/querystring.js
```diff
@@ -3,10 +3,11 @@ function parseQueryString(queryString) {
   if (queryString.length === 0) {
```
day-lee, line 3: Good job! 

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,14 @@
-function tally() {}
+function tally(input) {
+  if(!Array.isArray(input)) throw new Error("Input is not valid.");;
+  const queryParams = {};
+  for (const char of input) {
+    queryParams[char]=0;
+  }
+  for (const char of input) {
```
day-lee, line 7: I think this code could be combined with the above for...of loop using the OR operator `||`. Can you think of a less repetitive way to implement it?

day-lee, line 7: I think this code can be combined with the upper for...of loop using or operator `||`. Can you think of less repetitive way to implement it?

Sprint-2/implement/lookup.js
hmkavut: The iteration is changed based on comment.

Sprint-2/implement/querystring.js
hmkavut: Thank you for this useful information. I added decodeURIComponent function.

Sprint-2/implement/tally.js
hmkavut: I combined and put ternary operator/

Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,8 @@
-function contains() {}
+function contains(list, prop) {
+  if (typeof list !== "object" || list === null || Array.isArray(list))
```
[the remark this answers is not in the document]
hmkavut: This checks if list is not of type "object".
It filters out things like strings, number, boolean, function

hmkavut:
Thank you for review, I updated PR based on your comments.

day-lee:
Code looks good! 👍

hmkavut:
Thank you @day-lee

PR #569, opened 2025-07-10
Glasgow | May 2025 | Salah Ahmed | Data Groups | Sprint 1

avatarit:
Self checklist

- [ ] My changes meet the requirements of this task

## Changelist

I have solved all issues relaed to sprint 1 in Data Group


Sprint-1/fix/median.js
```diff
@@ -6,9 +6,32 @@
...
+    return null; // Return null if the input is not an array
+  }
+  const sortedList = list
+  .filter(item => typeof item === 'number')
+  .sort((a, b) => a - b);
+  
+  if (sortedList.length === 0) {
+    return null; // Return null if there are no numbers in the list
+  }
+
```
LonMcGregor, line 19: Good solution.
I would suggest that you could put your input tests (like checking for length 0) all at the start, rather than having your filter and sort in between them, but that is a matter of code style and not a bug

Sprint-1/implement/dedupe.js
```diff
@@ -1 +1,28 @@
...
+  //         }
+  //      }
+  //  }
+  //  return list;
+  //}
+
+//Notes / Limitations:
+//Time complexity is O(n²): Inefficient for large arrays.
+//Mutates the input array: If you need to preserve the original, you'll need to copy it first.
+
```
LonMcGregor, line 17: Good analysis of this solution's complexity. Lots of nested loops within loops should be avoided where possible. Your other solution is clever, making use of existing javascript functionality.

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,21 @@
...
+        return null; // Return null if the input is not an array
+    }
+    // Check if the input is an empty array
+    if (elements.length === 0) {
+        return -Infinity; // Return -Infinity if the array is empty
+    }
+    
+    // Filter out non-numeric values and find the maximum
+    const numericElements = elements.filter(element => typeof element === 'number').sort((a, b) => a - b);      
+    if (numericElements.length === 0) {
```
LonMcGregor, line 13: If you are already doing a sort, do you need to later use the `Math.max()` function? Is there maybe a quicker way to access the maximum of an already sorted array?

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,19 @@
...
+    if (elements.length === 0) {
+        return 0; // Return 0 if the array is empty
+    }
+    
+    // Filter out non-numeric values and calculate the sum
+    const numericElements = elements.filter(element => typeof element === 'number');
+    
+    // Return the sum of numeric elements
+    return numericElements.reduce((acc, curr) => acc + curr, 0);
 }
```
LonMcGregor, line 17: Good use of the reduce function here

LonMcGregor (commented):
Good solutions here, I have left one questions for you to answer

Sprint-1/fix/median.js
avatarit: I have added a test for empty array

PR #574, opened 2025-07-10
Sheffield|May-2025|Sheida Shabankari|Module-Data -Groups| Sprint-2

sheida-shab:
Self checklist

Briefly explain your PR.

I have implemented and modified functions and wrote test cases that all passed and answer the questions.

### DIFF
--- a/Sprint-2/debug/author.js
+++ b/Sprint-2/debug/author.js
@@ -11,6 +11,6 @@ const author = {
   alive: true,
 };
 //console.log(author);
-for (const value in author) {
-  console.log(value);
+for (const key in author) {
+  console.log(author[key]);
 }
--- a/Sprint-2/debug/author.js
+++ b/Sprint-2/debug/author.js
@@ -10,7 +10,7 @@ const author = {
   age: 40,
   alive: true,
 };
-
-for (const value of author) {
+//console.log(author);
+for (const value in author) {
   console.log(value);
 }
--- a/Sprint-2/debug/recipe.js
+++ b/Sprint-2/debug/recipe.js
@@ -12,4 +12,4 @@ const recipe = {
 
 console.log(`${recipe.title} serves ${recipe.serves}
   ingredients:
-${recipe}`);
+${recipe.ingredients}`);
--- a/Sprint-2/debug/recipe.js
+++ b/Sprint-2/debug/recipe.js
@@ -10,6 +10,8 @@ const recipe = {
   ingredients: ["olive oil", "tomatoes", "salt", "pepper"],
 };
 
-console.log(`${recipe.title} serves ${recipe.serves}
-  ingredients:
-${recipe.ingredients}`);
+console.log(`${recipe.title} serves ${recipe.serves}`);
+console.log("ingredients:");
+for(const item of recipe.ingredients){
+  console.log(item);
+}
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -55,3 +55,8 @@ test("Given invalid parameters like an array",() => {
     const result = contains(obj, "a");
     expect(result).toBe(false);
 });
+test("Given invalid parameters like an array2", () => {
+  const obj = ["a", 1, "b", 2];
+  const result = contains(obj, "1");
+  expect(result).toBe(false);
+});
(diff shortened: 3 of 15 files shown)


Sprint-2/debug/author.js
```diff
@@ -10,7 +10,7 @@ const author = {
   age: 40,
   alive: true,
 };
-
-for (const value of author) {
+//console.log(author);
+for (const value in author) {
   console.log(value);
 }
```
cjyuan, line 16: With for-in loop, `value` takes the **name of the keys/properties** and not the value of the properties.

Normally for-in loop is used as `for (const key in object) { ... }`

Sprint-2/debug/recipe.js
```diff
@@ -12,4 +12,4 @@ const recipe = {
 
 console.log(`${recipe.title} serves ${recipe.serves}
   ingredients:
-${recipe}`);
+${recipe.ingredients}`);
```
cjyuan, line 15: The requirement is to print each ingredient (element of the array) on a new line.

Sprint-2/implement/contains.test.js
```diff
@@ -20,16 +20,38 @@ as the object doesn't contains a key of 'c'
...
+  const result = contains(obj, "c");
+  expect(result).toBe(false);
+});
 
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+test("Given invalid parameters like an array",() => {
+    const obj = ["a",1,"b", 2 ];
+    const result = contains(obj, "a");
```
cjyuan, line 55: Array is a kind of object where indexes serve as its keys. So we could call `contains(["a",1,"b", 2 ], "1")` to check if the function can return `false` when the first parameter is an array and the second parameter is a valid key.

Sprint-2/implement/lookup.js
```diff
@@ -1,5 +1,14 @@
-function createLookup() {
-  // implementation here
+
+function createLookup(countryCurrArray) {
+    if (
+      !Array.isArray(countryCurrArray) ||
+      !countryCurrArray.every((item) => Array.isArray(item))
+    ) {
+      return false;
```
cjyuan, line 7: The function normally returns an object. For invalid parameters, it's better to return `null` or, alternatively, throw an error to indicate the failure explicitly.

Sprint-2/implement/querystring.js
```diff
@@ -6,8 +6,9 @@ function parseQueryString(queryString) {
   const keyValuePairs = queryString.split("&");
 
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
-    queryParams[key] = value;
+    const [key, ...value] = pair.split("=");
+    const strvalue = value.join("=");
+    queryParams[key] = strvalue;
```
cjyuan, line 11: Please note that in real querystring, both `key` and `value` are ***percent-encoded*** or ***URL encoded*** in the URL. For example, the string "5%" will be encoded as "5%25". So to get the actual value of "5%25" (whether it is a key or value in the querystring), you should call a function to decode it.

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,21 @@
-function tally() {}
+function tally(itemArray) {
+    if (itemArray.length == 0 ) {
+      return {};
+    }else{
+        if(typeof itemArray==="string"){
+            throw new Error("invalid input");
+        }
```
cjyuan, line 7: Why only consider string as invalid parameter? What if the parameter is null, undefined, an object?

Sprint-2/implement/contains.test.js
sheida-shab: Thanks for your comment.
The task was to check if an object has a specific property.
Since arrays are also objects but work differently, I made the function return false if the input is an array.
So the function only checks plain objects.I also added a test to confirm arrays return false, even if the key exists.
Please let me know your thoughts!

Sprint-2/implement/lookup.js
sheida-shab: I did it.

Sprint-2/implement/tally.js
sheida-shab: I changed my code .

cjyuan:
Changes look good. Well done!

PR #577, opened 2025-07-10
Sheffield | May-2025 | Hassan Osman | Sprint 2 Coursework

HassanOHOsman:
## Learners, PR Template

Self checklist

## Changelist

This PR includes bug fixes, function implementations, and code interpretation tasks completed for Sprint 2. All required tests pass, and unfamiliar concepts were explored using documentation and logging.

## Questions

### DIFF
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,3 +1,14 @@
-function contains() {}
+function contains(object, property) {
+    if (typeof object !== "object" || object == null || Array.isArray(object)) {
+        return false;
+    }
+    if (Object.keys(object).length === 0) {
+        return false;
+    } else {
+        return object.hasOwnProperty(property); 
+    }
+    
+
+}
 
 module.exports = contains;
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -20,16 +20,62 @@ as the object doesn't contains a key of 'c'
 // Given an empty object
 // When passed to contains
 // Then it should return false
-test.todo("contains on empty object returns false");
+test("contains on empty object returns false", () => {
+    
+    const currentOutput = contains({}, "2");
+    const targetOutput = false;
+    expect(currentOutput).toEqual(targetOutput);
+});
 
 // Given an object with properties
 // When passed to contains with an existing property name
 // Then it should return true
+const user = {
+  id: 101,
+  name: "Amina Yusuf",
+  email: "amina@example.com",
+  isVerified: true,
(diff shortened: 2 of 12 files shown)


Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,14 @@
-function contains() {}
+function contains(object, property) {
+    if (typeof object !== "object" || object == null || Array.isArray(object)) {
+        return false;
+    }
+    if (Object.keys(object).length === 0) {
+        return false;
+    } else {
```
cjyuan, line 7: If object is empty, the statement at line 8 would also return `false`. We don't have to treat empty object as a special case.

Sprint-2/implement/contains.test.js
```diff
@@ -20,16 +20,62 @@ as the object doesn't contains a key of 'c'
...
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+
+const programmingLanguages = ["JavaScript", "Python", "Java", "C++", "Go", "Rust", "Ruby"];
+const languages = null;
+const name = "Hassan";
+
+test("Returns false when invalid parameters are passed", () => {
+    const currentOutput = contains(programmingLanguages);
```
cjyuan, line 72: Array is a kind of object where indexes serve as its keys. So we could call `contains(programmingLanguages, "1")` to check if the function can return `false` when the first parameter is an array and the second parameter is a valid key.

Sprint-2/implement/querystring.js
```diff
@@ -6,11 +6,21 @@ function parseQueryString(queryString) {
...
-  }
+    const separatorIndex = pair.indexOf("=");
 
+     if (separatorIndex === -1) {
+      queryParams[pair] = "";
+       } else {
+       const key = pair.slice(0, separatorIndex);
+       const value = pair.slice(separatorIndex + 1);
+       queryParams[key] = value;
+       }
```
cjyuan, line 17: - Please note that in real querystring, both `key` and `value` are ***percent-encoded*** or ***URL encoded*** in the URL. For example, the string "5%" will be encoded as "5%25". So to get the actual value of "5%25" (whether it is a key or value in the querystring), you should call a function to decode it.

- Can you improve the indentation of this code?

Sprint-2/implement/contains.test.js
HassanOHOsman: Didn't think indexes could be treated as keys! Thanks for letting me know :)

cjyuan:
Changes look good. Well done!

PR #580, opened 2025-07-11
LONDON | MAY_2025 | EMILIANO_URUENA | DATA_GROUPS | SPRINT2

Emilianouz:
Self checklist

- [X] My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/contributing/)

## Changelist

**Sprint  2 - Module DATA GROUPS**
COMMITS:
1. Debugs: address.js, author.js, recipe.js, were debuged. | Sprint-2 MODULE-DATA-GROUPS
2. Implemented: contains, lookup, querystring, and tally were implemented with their test, according to instructions. | Sprint-2 MODULE-DATA-GROUPS
3. Interpret: Questions answered in invert.js, created the test and solve it. | Sprint-2 MODULE-DATA-GROUPS

### DIFF
--- a/Sprint-2/debug/address.js
+++ b/Sprint-2/debug/address.js
@@ -12,4 +12,4 @@ const address = {
   postcode: "XYZ 123",
 };
 
-console.log(`My house number is ${address[0]}`);
+console.log(`My house number is ${address["houseNumber"]}`);
--- a/Sprint-2/debug/author.js
+++ b/Sprint-2/debug/author.js
@@ -11,6 +11,7 @@ const author = {
   alive: true,
 };
 
-for (const value of author) {
+for (const value of Object.keys(author)) {
   console.log(value);
-}
+} 
+// An object is not iterable.
--- a/Sprint-2/implement/lookup.js
+++ b/Sprint-2/implement/lookup.js
@@ -1,5 +1,12 @@
-function createLookup() {
+function createLookup(countryCurrencyPairs) {
   // implementation here
+  let objectLookup = {}
+  for (i = 0 ; i<  countryCurrencyPairs.length ; i++){
+    //const [country, currency] = countryCurrencyPairs[i]
+     //objectLookup[country] = currency;
+    objectLookup[countryCurrencyPairs[i][0]] = countryCurrencyPairs[i][1];
+  }
+  return objectLookup
 }
 
 module.exports = createLookup;
--- a/Sprint-2/implement/querystring.js
+++ b/Sprint-2/implement/querystring.js
@@ -6,11 +6,11 @@ function parseQueryString(queryString) {
   const keyValuePairs = queryString.split("&");
 
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
-    queryParams[key] = value;
+    //const [key, value] = pair.split("=", 2);
+     const [key, value] = [pair.slice(0,pair.indexOf('=')),pair.slice(pair.indexOf('=')+1)];
+     queryParams[key] = value;
   }
-
   return queryParams;
 }
-
+console.log(parseQueryString("a=1&b=2"))
 module.exports = parseQueryString;
--- a/Sprint-2/implement/querystring.test.js
+++ b/Sprint-2/implement/querystring.test.js
@@ -7,6 +7,22 @@ const parseQueryString = require("./querystring.js")
 
 test("parses querystring values containing =", () => {
   expect(parseQueryString("equation=x=y+1")).toEqual({
-    "equation": "x=y+1",
+    equation: "x=y+1",
   });
 });
+
+test("parses querystring with multiple key-value pairs", () => {
+  expect(parseQueryString("a=1&b=2")).toEqual({
+    a: "1",
+    b: "2",
+  });
+});
+
+test("Handle more than 1 pair in the query string", () => {
+  expect(parseQueryString("sort=newest&color=blue")).toEqual({
+    sort: "newest",
+    color: "blue",
+  });
+});
+
+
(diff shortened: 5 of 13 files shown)


Sprint-2/debug/address.js
```diff
@@ -12,4 +12,4 @@ const address = {
   postcode: "XYZ 123",
 };
 
-console.log(`My house number is ${address[0]}`);
+console.log(`My house number is ${address["houseNumber"]}`);
```
MorganDavid, line 15: This is correct, another correct option is `address.houseNumber`

Sprint-2/debug/author.js
```diff
@@ -11,6 +11,7 @@ const author = {
   alive: true,
 };
 
-for (const value of author) {
+for (const value of Object.keys(author)) {
   console.log(value);
-}
+}
```
MorganDavid, line 16: An object is made up of keys and values:
```
{
<key>: <value>
}
```

The keys in this example are firstName, lastName, etc. 
The values are "Zadie", "Smith", etc.

The comment at the top asks you to log the values. Is that what your code does?

Sprint-2/implement/lookup.js
```diff
@@ -1,5 +1,12 @@
-function createLookup() {
+function createLookup(countryCurrencyPairs) {
   // implementation here
+  let objectLookup = {}
+  for (i = 0 ; i<  countryCurrencyPairs.length ; i++){
+    //const [country, currency] = countryCurrencyPairs[i]
+     //objectLookup[country] = currency;
+    objectLookup[countryCurrencyPairs[i][0]] = countryCurrencyPairs[i][1];
+  }
+  return objectLookup
```
MorganDavid, line 9: This is good! Alternatively, there's actually a built in javascript function to do this:
```suggestion
 return Object.fromEntries(countryCurrencyPairs);
```
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries

Sprint-2/implement/querystring.js
```diff
@@ -6,11 +6,11 @@ function parseQueryString(queryString) {
   const keyValuePairs = queryString.split("&");
 
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
-    queryParams[key] = value;
+    //const [key, value] = pair.split("=", 2);
+     const [key, value] = [pair.slice(0,pair.indexOf('=')),pair.slice(pair.indexOf('=')+1)];
```
MorganDavid, line 10: I find this line a bit hard to understand. Maybe we could split it up into multiple variables?

```suggestion
const equalsIndex = pair.indexOf('=');
     const [key, value] = [pair.slice(0,equalsIndex),pair.slice(pair.indexOf('=')+1)];
```

Breakingn it up into multiple lines with clear names makes it easier to follow

Sprint-2/implement/querystring.test.js
```diff
@@ -7,6 +7,22 @@ const parseQueryString = require("./querystring.js")
...
+    b: "2",
+  });
+});
+
+test("Handle more than 1 pair in the query string", () => {
+  expect(parseQueryString("sort=newest&color=blue")).toEqual({
+    sort: "newest",
+    color: "blue",
+  });
+});
```
MorganDavid, line 26: We could also test for these additional cases:

What happens if the string contains multiple equals signs?  e.g. `a=b=c=c&b=a`,

What happens if the query string is null, e.g. `a` in this example: `a=&b=hello`

Sprint-2/implement/tally.test.js
```diff
@@ -32,3 +32,15 @@ test.todo("tally on an empty array returns an empty object");
...
+    [
+        {input:['a'], expected:{ a: 1 }},
+        {input:['a', 'a', 'a'], expected:{ a: 3 }},
+        {input:['a', 'a', 'b', 'c'], expected:{ a : 2, b: 1, c: 1 }},
+        {input:[], expected:{}},
+        {input: 'String', expected: new Error("Input must be array")},
+    ].forEach(({input,expected})=>
+    it(`return an object containing the count for each unique item for [${input}]`,()=>
+    expect(tally(input)).toEqual(expected))
+);
```
MorganDavid, line 45: This testing pattern is great for adding new test cases easily, but the downside is that if a test fails, it can be hard to work out why it has failed, if each test has a clear message, it's more obvious what has gone wrong.

I think either approach is valid though!

Sprint-2/interpret/invert.test.js
```diff
@@ -0,0 +1,12 @@
+const invert = require("./invert.js");
+// E.g. invert({x : 10, y : 20}), target output: {"10": "x", "20": "y"}
+describe("invert", () =>{
+    [
+        {input:{x : 10, y : 20}, expected:{"10": "x", "20": "y"}},
+        {input:{ a : 1 }, expected:{ 1 : "a" }},
+        {input:{a : 1, b: 2}, expected:{ 1 : "a" , 2 : "b" }},
```
MorganDavid, line 7: what happens if there are multiple identical values? Objects cannot have multiple identical keys. 

I'm not sure what should happen in this case.

You could test for it with 
 ```suggestion
        {input:{a : 1, b: 1}, expected:/* What happens here? */ },
```

Emilianouz (commented):
Thank you for the feedback.

cjyuan (commented):
@MorganDavid 
Seeing that this PR has been reviewed a while ago and the trainee has made the necessary changes, I will mark this PR as "complete" first.

@Emilianouz I also have some suggestions and questions but you can address them when you have time.

PR #582, opened 2025-07-11
Glasgow | May-2025 | Bassam Alshujaa | Sprint-1

Bassam-Radwan:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

I have tested and implemented all files in this Sprint

## Questions

### DIFF
--- a/Sprint-1/implement/dedupe.test.js
+++ b/Sprint-1/implement/dedupe.test.js
@@ -26,8 +26,9 @@ test("given an empty array, it returns an empty array", () => {
 
 test("given an array with no duplicates, it returns a copy of the original array", () => {
   const input = [1, 2, 3, 4];
-  expect(dedupe(input)).toEqual(input);
-  expect(dedupe(input)).not.toBe(input); // different array instance
+  const output = dedupe(input);
+  expect(dedupe(input)).toEqual(output);
+  expect(dedupe(input)).not.toBe(output); // different array instance
 });
 // Given an array with strings or numbers
 // When passed to the dedupe function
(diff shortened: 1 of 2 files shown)


Sprint-1/implement/dedupe.test.js
```diff
@@ -16,12 +16,25 @@ E.g. dedupe([1, 2, 1]) target output: [1, 2]
...
+});
 
 // Given an array with no duplicates
 // When passed to the dedupe function
 // Then it should return a copy of the original array
 
+test("given an array with no duplicates, it returns a copy of the original array", () => {
+  const input = [1, 2, 3, 4];
+  expect(dedupe(input)).toEqual(input);
+  expect(dedupe(input)).not.toBe(input); // different array instance
```
MorganDavid, line 30: We should also test that the returned array contains the same values as the input. Otherwise dedupe could return any array here and the test will still pass.

 ```suggestion
  expect(dedupe(input)).not.toBe(input); // different array instance
  expect(dedupe(input)).toEqual(input);
```

File: Sprint-1/refactor/includes.js [hunk 2]
```diff
@@ -1,6 +1,9 @@
 // Refactor the implementation of includes to use a for...of loop
 
 function includes(list, target) {
+  if (!Array.isArray(list) || list.length === 0) {
+    return false;
+  }
   for (let index = 0; index < list.length; index++) {
     const element = list[index];
     if (element === target) {
```
MorganDavid, line 9: You'll need to refactor this to use a `for ... of` loop. 

You can see how that works here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of

MorganDavid (changes requested):
Looks really good! just two small things to change

Bassam-Radwan:
Thanks for your feedback @MorganDavid , I have fixed the issue and tested it and I updated the for loop to for.. .of

LonMcGregor:
Good work, it looks like you did resolve the suggested changes. You can close this PR as complete now.

PR #605, opened 2025-07-14
ZA | ITP-May-2025 | Revive Munashe Mapfumo | Module-Data-Groups | Alarm clock app

rivakutu:
## Learners, PR Template

Self checklist

## Changelist

Added countdown functionality that begins immediately after user input.

Validated user input to ensure it’s a non-negative number of seconds.

Converted total seconds into a readable format (MM:SS) for time display.

Implemented a soft gradient background for visual appeal

Ensured proper interval handling by clearing existing timers before starting a new countdown

## Questions

### DIFF
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -26,8 +26,13 @@ function setAlarm() {
   }, 1000);
 }
 
+// minutes and seconds to be displayed properly
 function updateDisplay(seconds) {
-  const formatted = `00:${seconds.toString().padStart(2, "0")}`;
+  const mins = Math.floor(seconds / 60);
+  const secs = seconds % 60;
+  const formatted = `${mins.toString().padStart(2, "0")}:${secs
+    .toString()
+    .padStart(2, "0")}`;
   document.getElementById(
     "timeRemaining"
   ).textContent = `Time Remaining: ${formatted}`;
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -11,19 +11,26 @@ function setAlarm() {
   updateDisplay(timeRemaining);
 
   // Clear any existing interval
-  if (window.alarmInterval) clearInterval(window.alarmInterval);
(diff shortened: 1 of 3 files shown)


Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,42 @@
...
+  if (isNaN(timeRemaining) || timeRemaining < 0) {
+    alert("Please enter a valid number of seconds.");
+    return;
+  }
+
+  // Update the display immediately
+  updateDisplay(timeRemaining);
+
+  // Clear any existing interval
+  if (window.alarmInterval) clearInterval(window.alarmInterval);
```
jenny-alexander, line 14: Instead of accessing window.alarmInterval, it would be better to declare a local variable to access this.
Imagine if you install a 3rd party library that also uses uses window.alarmInterval and it sets the timer -> this would overwrite your timer!

By using your own local variable, this will make sure that your interval is not overwritten by another library or 3rd party code.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,42 @@
...
+
+  // Start countdown
+  window.alarmInterval = setInterval(() => {
+    timeRemaining--;
+
+    updateDisplay(timeRemaining);
+
+    if (timeRemaining <= 0) {
```
jenny-alexander, line 22: After the alarm goes off and I hit 'Stop Alarm', I expected the input to be cleared too. This might be a nice user experience (UX) change to implement. What do you think?

jenny-alexander (changes requested):
Your Alarm Clock App is done very well 🙌
- Indentation is good
- Variable names are clear
- Timer is reset when I start a new timer and the previous timer didn't finish yet.
- Good check to ensure the ensure has only entered numbers.

I suggested a few changes for you to implement in order to make your code even more solid.

Sprint-3/alarmclock/alarmclock.js
rivakutu: ive declared alarmInterval variable

PR #607, opened 2025-07-14
 London | 25-ITP-May | Halyna Kozlovska | Sprint 3 | Alarm Clock

halyna-k:
## Learners, PR Template

Self checklist

- [ ] I have titled my PR with Region | Cohort | FirstName LastName | Sprint | Assignment Title
- [ ] My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/reviewing/style-guide/)

## Changelist

- Updated page title to "Alarm clock app" and added defer to script.
- Implemented countdown timer with input validation and display update.
- Plays alarm sound and flashes background when time reaches zero.
- Allows stopping alarm and clears previous timers on new set.

## Questions

### DIFF
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -1,7 +1,15 @@
 let countdownTimer;
+let flashInterval;
 
 function setAlarm() {
-  document.body.style.backgroundColor = "";
+  document.getElementById("stop").addEventListener("click", () => {
+    if (flashInterval) {
+      clearInterval(flashInterval);
+      flashInterval = null;
+      document.body.style.backgroundColor = "";
+    }
+  });
+
   let alarmTime = Number(document.querySelector("input").value);
 
   if (!alarmTime || isNaN(alarmTime) || alarmTime <= 0) {
@@ -27,6 +35,14 @@ function setAlarm() {
 
   updateTitle(alarmTime);
 
+  function flashingBackground() {
+    let isRed = false;
+    flashInterval = setInterval(() => {
+      document.body.style.backgroundColor = isRed ? "#ffffff" : "#ff4444";
+      isRed = !isRed;
+    }, 400);
+  }
+
   if (countdownTimer) {
     clearInterval(countdownTimer);
   }
@@ -37,7 +53,7 @@ function setAlarm() {
 
     if (timeLeft <= 0) {
       clearInterval(countdownTimer);
-      document.body.style.backgroundColor = "#ff4444";
+      flashingBackground();
       playAlarm();
     }
   }, 1000);
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -1,6 +1,7 @@
 let countdownTimer;
 
 function setAlarm() {
+  document.body.style.backgroundColor = "";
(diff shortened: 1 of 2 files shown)


Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,63 @@
...
+      clearInterval(flashInterval);
+      flashInterval = null;
+      document.body.style.backgroundColor = "";
+    }
+  });
+
+  let alarmTime = Number(document.querySelector("input").value);
+
+  if (!alarmTime || isNaN(alarmTime) || alarmTime <= 0) {
+    alert("Please enter a valid positive number for the alarm time");
```
illicitonion, line 16: Good job making sure the user sees what's wrong.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,63 @@
...
+    return;
+  }
+
+  let timeLeft = alarmTime;
+
+  function formatTime(time) {
+    let minutes = Math.floor(time / 60);
+    let seconds = time % 60;
+    return `${minutes.toString().padStart(2, "0")}:${seconds
```
illicitonion, line 25: This is quite a lot of complex code to be inside a string interpolation - it's quite hard to read at a glance what it's doing. Can you think how to make this easier to read?

Also, you have some repeated code here - the 0-padding is the same for both minutes and hours - can you think how to reduce that duplication, e.g. so that if I wanted to change the number of leading zeroes I could do that with just one change not two?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,63 @@
...
+      .toString()
+      .padStart(2, "0")}`;
+  }
+
+  function updateTitle(time) {
```
illicitonion, line 30: Is there a reason you're defining all of these functions inside `setAlarm`? It makes it a bit harder to work out what `setAlarm` is doing because there are a bunch of functions in between the code actually doing stuff. Generally we tend to define functions at the top-level of a file instead of nested, unless there's a reason not to.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,63 @@
...
+    document.getElementById(
+      "timeRemaining"
+    ).innerHTML = `Time Remaining: ${formatTime(time)}`;
```
illicitonion, line 33: Are you including any HTML in the generated text here? If not, could you use `innerText` instead? Have a read of https://owasp.org/www-community/attacks/xss/ to see why we may prefer `innerText`

illicitonion (commented):
This is generally looking good, but I've left a few comments about code style to consider

Sprint-3/alarmclock/alarmclock.js
halyna-k: @illicitonion Thanks for the suggestion. Since the content is plain text without any HTML, I’ve switched from innerHTML to textContent to ensure safer rendering and avoid any XSS vulnerabilities.

Sprint-3/alarmclock/alarmclock.js
halyna-k: @illicitonion Thanks! I’ve moved the helper functions outside setAlarm to keep the main function focused and improve readability.

PR #620, opened 2025-07-16
London | 25-ITP-May | Halyna Kozlovska | Sprint 3 | Quote Generator

halyna-k:
## Learners, PR Template

Self checklist

## Changelist

- Create container dynamically with quote, author, button, toggle checkbox, and status text.
- Display a random quote and author on load and on button click.
- Add event listener to the button to load a new random quote.
- Add a checkbox toggle for auto-play, which could automatically changes the quote every 60 seconds.

## Questions

### DIFF
--- a/Sprint-3/quote-generator/quotes.js
+++ b/Sprint-3/quote-generator/quotes.js
@@ -495,11 +495,12 @@ function createContainer() {
   const container = document.createElement('div');
   container.id = 'quote-container';
   container.className = 'container';
-  
-  document.body.appendChild(container);
-  [...document.body.children]
-    .filter((el) => el !== container)
-    .forEach((el) => container.appendChild(el));
+
+  ['quote', 'author', 'new-quote'].forEach(id => {
+    container.appendChild(document.getElementById(id));
+  });
+
+  document.body.prepend(container);
 }
 
 function createAutoPlayCheckbox() {
@@ -548,14 +549,10 @@ function handleButton() {
   });
 }
 
-function renderQuote() {
-  setQuoteAndAuthor();
-}
-
 function setup() {
   createContainer();
   createAutoPlayCheckbox()
-  renderQuote();
+  setQuoteAndAuthor();
   handleButton();
   handleAutoPlayToggle();
 }


Sprint-3/quote-generator/quotes.js
```diff
@@ -491,3 +491,71 @@ const quotes = [
 ];
 
 // call pickFromArray with the quotes array to check you get a random quote
+function createContainer() {
+  const h1 = document.querySelector("h1");
+  if (h1) {
+    h1.remove();
```
LonMcGregor, line 497: If you never want the header, you should remove it from the index.html instead of through javascript

Sprint-3/quote-generator/quotes.js
```diff
@@ -491,3 +491,71 @@ const quotes = [
...
+
+  container.innerHTML = document.body.innerHTML;
+
+  document.body.innerHTML = "";
+  document.body.appendChild(container);
+
+  const autoPlayWrapper = document.createElement("div");
+  autoPlayWrapper.id = "auto-play-wrapper";
+  autoPlayWrapper.innerHTML = `
+    <label for="auto-play" class="auto-play-label">
```
LonMcGregor, line 512: The opposite problem of the header - if you always want an element on the page it is best to add it to the HTML rather than dynamically in JavaScript.

Sprint-3/quote-generator/quotes.js
```diff
@@ -491,3 +491,71 @@ const quotes = [
...
+  }
+
+  const container = document.createElement('div');
+  container.id = 'quote-container';
+  container.className = 'container';
+
+  container.innerHTML = document.body.innerHTML;
+
+  document.body.innerHTML = "";
+  document.body.appendChild(container);
```
LonMcGregor, line 507: I am confused what is going on in lines 500-507 - can you explain what you are doing here?

Sprint-3/quote-generator/quotes.js
```diff
@@ -491,3 +491,71 @@ const quotes = [
...
+function handleButton() {
+  const button = document.getElementById("new-quote");
+
+  button.addEventListener("click", () => {
+    setQuoteAndAuthor();
+  });
+}
+
+function renderQuote() {
+  createContainer();
```
LonMcGregor, line 555: You shouldn't re-create HTML every time you run this, if you can re-use what is already there. In a more complex web application this could be inefficient.

This also means you won't need tore-add the listeners every time

LonMcGregor (commented):
Good start on this task  - there are some parts I am confused about, particularly the createContainer() function. Could you please take a look at my comments.

Sprint-3/quote-generator/quotes.js
halyna-k: @LonMcGregor Good point — thanks! I’ve removed the <h1> directly from index.html instead of handling it with JavaScript.

Sprint-3/quote-generator/quotes.js
LonMcGregor: OK. If that was part of the requirements for practice, adding it dynamically is fine. I would suggest though that you only need to add this autoplay html once. Right now your solution involves re-creating it every time the renderQuote() function is called.

halyna-k:
@LonMcGregor You're right. I removed renderQuote and now call setQuoteAndAuthor directly.
As for createContainer, I initially moved all body children, but that wasn’t necessary — now it only wraps the specific elements (quote, author, new-quote).

PR #625, opened 2025-07-16
London | May_25 | Emiliano_Uruena | MDG_Sprint-3 | AlarmClock

Emilianouz:
Self checklist

## Changelist

Alarmclock implemented, including extra buttons: pause, resume and resent. All buttons are adapted.

Given the user has entered a number in the input field
When the user clicks the "Set Alarm" button
Then the "Time Remaining" title should update to show the entered number in mm:ss format

Given the alarm is set with a valid time
When one second passes
Then the "Time Remaining" title should decrement by 1 second

Given the alarm is set with a time of 00:00
When the timer reaches 00:00
Then the alarm sound should play continuously

Given the alarm sound is currently playing
When the user clicks the "Stop Alarm" button
Then the alarm sound should stop playing

Given the alarm is set with a time of 00:10
When the timer reaches 00:00
Then the background color should change
And the alarm sound should play

Given the user has not set an alarm
When the page first loads
Then the "Time Remaining" title should show 00:00
And no alarm sound should play

### DIFF
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -1,4 +1,115 @@
-function setAlarm() {}
+// disable/enable buttons
+  function buttonsInitial(){
+  document.getElementById("stop").disabled = true;
+  document.getElementById("pause").disabled = true;
+  document.getElementById("reset").disabled = true;
+  document.getElementById("resume").disabled = true;
+  document.getElementById("set").disabled = false;
+  }
+  function buttonsAlarmRunning(){ // Buttons alarmRunning 
+  document.getElementById("stop").disabled = true;
+  document.getElementById("pause").disabled = false;
+  document.getElementById("reset").disabled = true;
+  document.getElementById("resume").disabled = true;
+  document.getElementById("set").disabled = true;
+  }
+  function buttonsAlarmPlayed(){
+    // buttons alarm played
+  document.getElementById("stop").disabled = false;
+  document.getElementById("pause").disabled = true;
+  document.getElementById("reset").disabled = true;
+  document.getElementById("resume").disabled = true;
+  document.getElementById("set").disabled = true;
+  }
+  function buttonsAlarmPaused(){
+  document.getElementById("stop").disabled = true;
+  document.getElementById("pause").disabled = true;
+  document.getElementById("reset").disabled = false;
+  document.getElementById("resume").disabled = false;
+  document.getElementById("set").disabled = true;
+  }
+  function buttonsAlarmResume(){
+  document.getElementById("stop").disabled = true;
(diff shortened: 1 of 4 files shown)


package.json
LonMcGregor, line 1: Did you mean to commit these files? If the changes here to package files are not necessary for your app you shouldn't commit them

Sprint-3/alarmclock/index.html
```diff
@@ -4,16 +4,21 @@
...
   <body>
     <div class="centre">
       <h1 id="timeRemaining">Time Remaining: 00:00</h1>
       <label for="alarmSet">Set time to:</label>
-      <input id="alarmSet" type="number" />
+      <input id="alarmSet" type="number"/>
 
       <button id="set" type="button">Set Alarm</button>
       <button id="stop" type="button">Stop Alarm</button>
+      <!-- new functions pause, resume, reset. -->
```
LonMcGregor, line 17: You do a good job of disabling buttons that should not be used depending on the state of the app. Can you figure out how to make the right ones disabled when you first start the app?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,115 @@
...
+  document.getElementById("resume").addEventListener("click", function () {
+    resume();
+  });
+  function resume() {
+    runInterval = setInterval(labelUpdate, 1000); 
+    buttonsAlarmResume();
+  }
+
+  // reset function
+  document.getElementById("reset").addEventListener("click", function () {
```
LonMcGregor, line 96: If you are adding an event listener with a function that only calls another function, there may be a way to simplify this to one line. Can you find out how?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,115 @@
-function setAlarm() {}
+// disable/enable buttons
+  function buttonsInitial(){
+  document.getElementById("stop").disabled = true;
```
LonMcGregor, line 3: There is a lot of code duplication here - can you think of any way it could be cleaned up?

LonMcGregor (commented):
Good work on this - I like the use of state management. I've left some comments for you to look at

package.json
Emilianouz: Precisely we have a talk about that last week.  I was committing all files that were changed. But now I know this files are not necessary.

Sprint-3/alarmclock/alarmclock.js
Emilianouz: I didn’t notice that. I can clean it up by creating a single function that sets the disabled state of all buttons based on an object. Is that what you meant?

LonMcGregor:
Good work on these changes - you are now done with this part of the sprint

PR #626, opened 2025-07-16
London | ITP-May-2025 | Mo Muchunu | Module-Data-Groups | Sprint 3 | Alarm-Clock

Mo-Muchunu:
## Learners, PR Template

Self checklist

## Changelist

- Set Alarm functionality based on user input to show Time Remaining countdown in the title.

- Implemented countdown timer with setInterval, updating title every second and triggering playAlarm() at 00:00.

- Added "Stop Alarm" button to stop sound and reset the timer with clearInterval.

- Added background color change and flashing effect when timer ends (bonus features).

- Validated input, structured DOM interactions into clear functions, and tested all functionality in browser.

### DIFF
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -5,6 +5,7 @@ let flashInterval;                                                              
 const random = num => Math.floor(Math.random() * (num + 1));                                // Return random integer to generate RGB color values for flashing background
 
 const setAlarm = () => {
+  pauseAlarm();                                                                             // Stop sound and flashing from any previous alarm
   clearInterval(runningTimer);                                                              // Clear any previous running timer
 
   const timeInput = document.getElementById("alarmSet");                                    // Get user input element 


Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,60 @@
-function setAlarm() {}
+let runningTimer;                                                                           // Store the interval ID for countdown timer                        
+let timeLeft = 0;                                                                           // Remaining time in seconds
+let flashInterval;                                                                          // Store interval ID for flashing background effect
+
+const random = num => Math.floor(Math.random() * (num + 1));                                // Return random integer to generate RGB color values for flashing background
+
+const setAlarm = () => {
+  clearInterval(runningTimer);                                                              // Clear any previous running timer
```
LonMcGregor, line 8: Setting a new alarm does not seem to stop the buzzer or the flashing colours - which means if you set a new alarm without stopping the one before it, it starts flashing really fast.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,60 @@
...
+
+  window.playAlarm = () => {
+    startFlash();                                                                           // Run the original alarm response
+    flashInterval = setInterval(() => {                                                     // Start a new interval that changes the background colour 
+      document.body.style.backgroundColor =
+        `rgb(${random(255)}, ${random(255)}, ${random(255)})`;
+    }, 300);                                                                                //  Change background every 300ms to create flashing effect
+  };
+
+  window.pauseAlarm = () => {
```
LonMcGregor, line 50: Is there a reason you use `window.fn_name` to create these functions? It's not incorrect, but I'm curious why you don't use `function name () {}`?

LonMcGregor (commented):
Hi, good work on this clock. I've left some comments for you to answer.

Sprint-3/alarmclock/alarmclock.js
Mo-Muchunu: You're absolutely right - I overlooked that and didn’t test thoroughly enough. The alarm sound gets pretty irritating, so I was just hitting the stop button instead of trying to set a new alarm while it was still buzzing!

I have fixed this by calling `pauseAlarm()` at the beginning of `setAlarm()` to stop the alarm sound, clear the flashing interval, and reset the background color. This now prevent any overlap when setting a new alarm.

Sprint-3/alarmclock/alarmclock.js
Mo-Muchunu: I used `window.playAlarm = () => {}` to override the original playAlarm function at runtime, while first saving the original `playAlarm` and `pauseAlarm` into local constants (`startFlash` and `pauseFlash`). This allowed me to extend the original behaviour - adding a flashing background effect -without modifying the code block marked "DO NOT EDIT BELOW HERE".

By re-assigning `window.playAlarm`, I effectively wrapped the original logic and made the enhanced version available globally, so it could still be called from anywhere in the codebase while including a new visual feature.

Mo-Muchunu:
> Hi, good work on this clock. I've left some comments for you to answer.

Hello @LonMcGregor, thank you for your feedback and time in reviewing my work!

PR #647, opened 2025-07-18
LONDON | MAY 2025 | JESUS DEL MORAL | SPRINT 2 | DATA GROUPS

delmorallopez:
Sprint 2 Module Data Groups

### DIFF
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,3 +1,11 @@
-function contains() {}
+function contains(input, containsvalues) {
+  if (!input || !containsvalues) {
+    return false;
+  }
+
+  return Object.values(input).includes(containsvalues);
+
+}
 
 module.exports = contains;
+
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,11 +1,19 @@
-function contains(input, containsvalues) {
-  if (!input || !containsvalues) {
-    return false;
+function contains(input, value) {
+    if (
+      input === null ||                          // explicitly reject null
+      typeof input !== 'object' || 
+      Array.isArray(input) || 
+      value === undefined || value === null      // allow empty string
+    ) {
+      return false;
+    }
+  
+    const valueExists = Object.values(input).includes(value);
+    const keyExists = Object.hasOwn(input, value);
+  
+    return valueExists || keyExists;
   }
-
-  return Object.values(input).includes(containsvalues);
-
-}
-
+  
+  
 module.exports = contains;
 
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -1,15 +1,14 @@
 const contains = require("./contains.js");
 
-/*
-Implement a function called contains that checks an object contains a
-particular property
+//Implement a function called contains that checks an object contains a
+//particular property
 
-E.g. contains({a: 1, b: 2}, 'a') // returns true
-as the object contains a key of 'a'
+//E.g. contains({a: 1, b: 2}, 'a') // returns true
+//as the object contains a key of 'a'
 
-E.g. contains({a: 1, b: 2}, 'c') // returns false
-as the object doesn't contains a key of 'c'
-*/
+//E.g. contains({a: 1, b: 2}, 'c') // returns false
+//as the object doesn't contains a key of 'c'
+//*/
 
 // Acceptance criteria:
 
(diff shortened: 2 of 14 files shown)


Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,11 @@
-function contains() {}
+function contains(input, containsvalues) {
+  if (!input || !containsvalues) {
+    return false;
+  }
+
+  return Object.values(input).includes(containsvalues);
```
cjyuan, line 6: You could also consider the following two approaches for determining if an object contains a property:
```
```

Sprint-2/implement/contains.test.js
```diff
@@ -20,16 +19,72 @@ as the object doesn't contains a key of 'c'
...
+    occupation: "writer",
+    age: 40,
+    alive: true,
+  };
+
+  const currentOutput = contains(input, "errortest");
+  const targetOutput = false;
+
+  expect(currentOutput).toEqual(targetOutput);
+
```
cjyuan, line 88: - Why are these parameters considered invalid?

- Can you make your function return the following expected result?
```
  // An array is a kind of object, and its keys are its indexes.
  contains(["A", "B", "C"], "1"); // expect false because the first parameter is an array

  // Empty string can also be key
  contains({ "": "Boo" }, "");  // expect true
  
  contains(null, "null"); // expect false even though type of null is equal to "object"
```

Sprint-2/implement/querystring.js
```diff
@@ -1,12 +1,15 @@
...
   }
+
   const keyValuePairs = queryString.split("&");
 
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
+    const [key, ...rest] = pair.split("=");
+    const value = rest.join("=");
     queryParams[key] = value;
   }
```
cjyuan, line 14: Please note that in real querystring, both `key` and `value` are ***percent-encoded*** or ***URL encoded*** in the URL. For example, the string "5%" will be encoded as "5%25". So to get the actual value of "5%25" (whether it is a key or value in the querystring), you should call a function to decode it.

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,27 @@
...
+    if (!Array.isArray(items)) {
+      throw new Error("Input must be an array");
+    }
+  
+    const counts = {};
+  
+    for (const item of items) {
+      counts[item] = (counts[item] || 0) + 1;
+    }
+
```
cjyuan, line 17: Does the following function call returns the value you expect?
```
```

Sprint-2/stretch/count-words.js
```diff
@@ -26,3 +26,20 @@
 
 3. Order the results to find out which word is the most common in the input
 */
+
+function countWords(querystring) {
+  const wordCounts = {};
```
cjyuan, line 31: Same issue as tally(). 

Could consider using an empty object with no inherited properties.

Sprint-2/implement/contains.js
delmorallopez: To check --  Property name (key) exists, we use propertyName in obj or hasOwn(obj, key)
To check -- Property value exists, we use Object.values(obj).includes(value)

    const valueExists = Object.values(input).includes(value);
    const keyExists = Object.hasOwn(input, value);
  
    return valueExists || keyExists;

Sprint-2/implement/contains.test.js
delmorallopez: The parameters are considers invalid because "errortest" is not a Key or value fo the object 
I added all the tests considering the expected result on your suggestion 

function contains(input, value) {
    if (
      input === null ||                          // explicitly reject null
      typeof input !== 'object' || 
      Array.isArray(input) || 
      value === undefined || value === null      // allow empty string
    ) {
      return false;
    }
  
    const valueExists = Object.values(input).includes(value);
    const keyExists = Object.hasOwn(input, value);
  
    return valueExists || keyExists;
  }

Sprint-2/implement/querystring.js
delmorallopez: I used the function decodeURIComponent()

    const [key, ...rest] = pair.split("=");
    const decodedKey = decodeURIComponent(key);
    const decodedValue = decodeURIComponent(rest.join("="));
    queryParams[decodedKey] = decodedValue;
  }

Sprint-2/stretch/count-words.js
delmorallopez: const wordCounts =  Object.create(null); // No inherited properties;

PR #652, opened 2025-07-19
Glasgow| May-2025| Bassam Alshujaa| Sprint-2 | module-Data-Groups

Bassam-Radwan:
## Learners, PR Template

Self checklist

## Changelist

I have answered and implemented all files 

## Questions

### DIFF
--- a/Sprint-2/implement/lookup.js
+++ b/Sprint-2/implement/lookup.js
@@ -1,11 +1,12 @@
 function createLookup(obj) {
+  if (!Array.isArray(obj) || !obj.every(Array.isArray)) {
+    throw new Error("Input must be an array of arrays");
+  }
   const cur = {};
   for (let list of obj) {
-    for (let country = 0; country < list.length; country++) {
       cur[list[0]] = list[1];
-    }
   }
-  return cur
+  return cur;
 }
 
 module.exports = createLookup;
--- a/Sprint-2/implement/lookup.js
+++ b/Sprint-2/implement/lookup.js
@@ -1,5 +1,11 @@
-function createLookup() {
-  // implementation here
+function createLookup(obj) {
+  const cur = {};
+  for (let list of obj) {
+    for (let country = 0; country < list.length; country++) {
+      cur[list[0]] = list[1];
+    }
+  }
+  return cur
 }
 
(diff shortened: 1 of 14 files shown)


Sprint-2/implement/lookup.js
LonMcGregor, line 1: You might want to consider validating the input here. Can you see what happens if you pass in a string instead of an array?

Sprint-2/implement/lookup.test.js
```diff
@@ -1,6 +1,10 @@
 const createLookup = require("./lookup.js");
 
-test.todo("creates a country currency code lookup for multiple codes");
+test("creates a country currency code lookup for multiple codes" , () =>{
+  const input = createLookup([['US', 'USD'], ['CA', 'CAD']]);
+  const outPut = { US: "USD", CA: "CAD" };
+  expect(input).toEqual(outPut);
+});
```
LonMcGregor, line 8: You could also consider testing the cases where invalid input is given, if you add validation checks to lookup,js

package-lock.json
LonMcGregor, line 1: Try to avoid comitting package-lock.json unless you really mean to make changes to it. Can you remove it (revert it back to what it was originally)?

Sprint-2/implement/tally.test.js
LonMcGregor, line 1: Can you think of some extra test cases to check for? You only have a single test here.

Sprint-2/implement/querystring.js
LonMcGregor, line 1: If I give this a string with more than one query, e.g. `"name=John&name=Jane"` it doesn't work right. Can you check your implementation and maybe add some more test cases to look for these?

LonMcGregor (commented):
You have made a good start on this sprint, but there are still a few areas where you need to make changes. Can you attempt these?

Bassam-Radwan:
Thanks for your feedback @LonMcGregor , I have fixed and tested all issues.

LonMcGregor:
Good work. In general, if you can pass invalid arguments as a parameter which give an output back instead, you should do validation checks, and you should try write tests for every possible branch point in a complex function. You are complete with this sprint now.

PR #677, opened 2025-07-22
LONDON | 25-ITP-MAY | TEWODROS BEKERE | M-D-G | SPRINT-3 | Quote_Generator_App

iteddy16:
…ndom quote functionality in JavaScript for Quote Generator App

## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

- It displays a quote randomly from the quote.js file and I apply some CSS features to make it look good. 
## Questions

### DIFF
--- a/Sprint-3/quote-generator/quotes.js
+++ b/Sprint-3/quote-generator/quotes.js
@@ -1,3 +1,22 @@
+function getRandomQuote() {
+  let quoteDom = document.getElementById("quote");
+  let authorDom = document.getElementById("author");
+  let randomEntry = pickFromArray(quotes);
+  let quote = randomEntry.quote;
+  let author = randomEntry.author;
+  quoteDom.textContent = quote;
+  authorDom.textContent = author;
+}
+// add quote when page loaded first time
+document.addEventListener('DOMContentLoaded', function () {
+  getRandomQuote();
+});
+// add event to button for changing quotes
+let newQuoteButton = document.getElementById("new-quote");
+newQuoteButton.addEventListener("click", () => {
+  getRandomQuote();
+});
+
 // DO NOT EDIT BELOW HERE
 
 // pickFromArray is a function which will return one item, at
@@ -16,10 +35,11 @@
 // pickFromArray(['a','b','c','d'])     // maybe returns 'c'
 
 // You don't need to change this function
+
 function pickFromArray(choices) {
+  if (!Array.isArray(choices) || choices.length === 0) return null;
   return choices[Math.floor(Math.random() * choices.length)];
 }
-
 // A list of quotes you can use in your app.
 // DO NOT modify this array, otherwise the tests may break!
 const quotes = [
@@ -490,4 +510,5 @@ const quotes = [
   },
 ];
 
+console.log(pickFromArray(quotes));
 // call pickFromArray with the quotes array to check you get a random quote
--- a/Sprint-3/quote-generator/style.css
+++ b/Sprint-3/quote-generator/style.css
@@ -1,1 +1,67 @@
-/** Write your CSS in here **/
+/* Basic Reset */
+* {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+}
+
+/* Fullscreen Flex Centered Layout */
+body {
+  font-family: 'Segoe UI', Tahoma, Geneva, Verdana, sans-serif;
+  background: linear-gradient(to right, #ffecd2, #fcb69f);
+  min-height: 100vh;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  padding: 2rem;
+}
+
+/* Wrapper to stack everything vertically */
+.container {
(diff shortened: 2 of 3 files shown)


Sprint-3/quote-generator/quotes.js
```diff
@@ -490,4 +510,5 @@ const quotes = [
   },
 ];
 
-// call pickFromArray with the quotes array to check you get a random quote
+console.log(pickFromArray(quotes));
```
jenny-alexander, line 513: It's a good idea to remove console logs before submitting your work for review.

Sprint-3/quote-generator/quotes.js
```diff
@@ -1,3 +1,22 @@
+function getRandomQuote() {
```
jenny-alexander, line 1: If your variable will not be reassigned within it's scope, it's better practice to use 'const' instead of 'let'.
Further reading: https://www.freecodecamp.org/news/differences-between-var-let-const-javascript/

Sprint-3/quote-generator/quotes.js
jenny-alexander, line 1: You did a great job assigning quote and author from the random quote object. 

🌱 This suggestion isn't required but I wanted point out that you can use destructuring here. You can use object destructuring to extract both properties (quote and author) in one line. For example:
```let { quote, author } = pickFromArray(quotes);```
This is a common JS feature that lets you pull out several properties from an object at once

File: Sprint-3/quote-generator/quotes.js [hunk 3]
```diff
@@ -16,10 +35,11 @@
 // pickFromArray(['a','b','c','d'])     // maybe returns 'c'
 
 // You don't need to change this function
+
 function pickFromArray(choices) {
+  if (!Array.isArray(choices) || choices.length === 0) return null;
```
jenny-alexander, line 40: Since pickFromArray could possibly return null (i.e. if the quotes array was empty), what can you do to safeguard your code to ensure that if null is returned, your function doesn't throw an error?

Sprint-3/quote-generator/style.css
jenny-alexander, line 1: Can you run Lighthouse on your app? There are a few suggestions to implement in order to make your page a bit more accessibile.
(In order to run lighthouse, go to dev tools, and choose 'Lighthouse' from the menu of items available).

jenny-alexander (changes requested):
Your quote generator works very well. I like your styling a lot too! Please review the comments and implement the suggested changes.

Sprint-3/quote-generator/quotes.js
iteddy16: Ok, thank you.

Sprint-3/quote-generator/quotes.js
iteddy16: Fixed, it's handled by displaying an alert to the user.

Sprint-3/quote-generator/style.css
iteddy16: No, I checked it was 93%. Now it's fixed the background color (white), and the new-quote button text (white) was the same; that's why.

jenny-alexander: It's comes back as 90% - which is pretty good. I think the accessibility issues are easy enough to improve on. Not necessary for this PR but something for you to keep in mind in future work!
Screenshot 2025-08-01 at 5 11 12 PM

jenny-alexander:
> @jenny-alexander, thank you for your comments and suggestions. I think I fixed the issue raised. Will you check it now? Thank you.

Hi @iteddy16 - I don't see your latest commit with the fixes. Maybe you forgot to commit your changes?

iteddy16:
@jenny-alexander, Yes, you are right. I forgot to push it to my remote, sorry. Here I fixed it. Will you check it and if it's good, complete my tag? Thank you.

jenny-alexander:
> @jenny-alexander, Yes, you are right. I forgot to push it to my remote, sorry. Here I fixed it. Will you check it and if it's good, complete my tag? Thank you. 

Hi,

Yes I can review it again, but it's still missing improvements for these:
- Removal of console log
- Use const instead of let
- Check for null returned from pickFromArray function.

iteddy16:
@jenny-alexander, I'm sorry. I assumed the first two were just recommendations, and I recall adding an alert display to handle the null; maybe I didn't push it to the remote. It happened to me a lot. 
Can you check it for me now, thank you.

iteddy16:
@jenny-alexander Hi Jenny, please can you review my latest commit? I have to submit it, thank you.

jenny-alexander:
> @jenny-alexander Hi Jenny, please can you review my latest commit? I have to submit it, thank you.

Hi, so sorry for the delay. I was away for the past week. I'll review it tomorrow (Sunday Aug. 10).

iteddy16:
@jenny-alexander, i removed the comments from the JS files. Thank you,

PR #680, opened 2025-07-22
ZA | ITP-MAY-25 | Shannon Davids | Module : Data groups | Sprint-2

Shannon-Davids:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-2/implement/querystring.js
+++ b/Sprint-2/implement/querystring.js
@@ -6,8 +6,26 @@ function parseQueryString(queryString) {
   const keyValuePairs = queryString.split("&");
 
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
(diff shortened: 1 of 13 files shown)


Sprint-2/implement/querystring.js

Sprint-2/implement/querystring.test.js
day-lee, line 1: Good job! 

In the real world, URLs are usually encoded with percent encoding to represent special characters. In URL, & can be used as a separator (key=value&key=value...) or special character within a key or value. if we were to use & as a special character, we need to encode it. Your code already use decodeURIComponent() so it will work as I described.

Can you add the test code so that we can check parseQueryString("a%25b=c%26d") results in { "a%b": "c&d" }?

day-lee:
Good job! 👍

PR #687, opened 2025-07-23
London | 25-ITP-May | Surafel Workneh | Sprint-3 | Alarmclock

SuWebOnes:
## Learners, PR Template

Self checklist

- [x]  I have titled my PR with Region | Cohort | FirstName LastName | Sprint | Assignment Title
- [x]  I have tested my changes
- [x]  My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/reviewing/style-guide/)
- [x]  My changes meet the [requirements](./README.md) of this task

## Changelist

Briefly explain your PR.

## Questions


Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,5 +1,38 @@
...
+  // Get the input value (number of seconds)
+  const input = document.getElementById("alarmSet");
+  timeRemaining = parseInt(input.value, 10);
+
+  // Immediately update the display
+  updateTimeDisplay();
+
+  // Clear any existing timer
+  if (intervalId) {
+    clearInterval(intervalId);
```
LonMcGregor, line 14: You are correctly clearing the interval when starting a new timer - but can you see any problems that may happen if you start a timer after the alarm sound has started playing?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,5 +1,38 @@
-function setAlarm() {}
+let timeRemaining = 0;
+let intervalId;
 
+function setAlarm() {
+  // Get the input value (number of seconds)
+  const input = document.getElementById("alarmSet");
+  timeRemaining = parseInt(input.value, 10);
```
LonMcGregor, line 7: Good idea to parse the input as an integer. Can you think of any validation checks you might want to put on this value before you use it?

LonMcGregor (commented):
Hi, good work on this sprint.
1. Typically we would only commit package files if we are making absolutely necessary changes to the node configuration. Is that the case here?
2. Good work on the alarm clock, I can see some areas where you could improve it a bit and have left comments there.

Sprint-3/alarmclock/alarmclock.js
SuWebOnes: yes, i put pauseAlarm(true); ensures the sound stops before starting a new timer.

LonMcGregor:
Great work on this. Your app is functioning well now. I still see the package files in this commit - do they need to be there? If you didn't mean to change them you don't need to commit them. Do you know how to remove them?

SuWebOnes:
can i remove them my project folder if it resolves?

SuWebOnes:
can you guide me to resolve " Great work on this. Your app is functioning well now. I still see the package files in this commit - do they need to be there? If you didn't mean to change them you don't need to commit them. Do you know how to remove them?"

LonMcGregor:
I suggest looking online for help with how to revert files. When you find out how, revert the changes to the package.json files and make a new commit on the branch - this will return them to their original state, and will clean up the commit. It is good practice to only commit files relevant to the feature being worked on, in this case the alarm clock.

LonMcGregor:
If you look at the files tab here on github, you can see that the `package.json` files and the `package-lock.json` files are still present. Only files that you are working on during the current sprint / feature should be included, and in this case that doesn't include the node package files.

When you are doing git changes and you see merge conflicts, you need to manually check the files and change them to include only what you expect. If you look in your files now, you will see lines like `<<<<<<< HEAD` - these should never be committed without fixing them.

LonMcGregor:
Looking at the first commit you made (ce05dda) - that looks like the output I would expect to fix this. If you repeat that for the other files, that will revert them.

Using the command I suggested, if you only give it the names of package files, it won't affect your changes made in other files.

SuWebOnes:
Good morning @LonMcGregor ,  know I completed and resolved the merge conflicts

LonMcGregor (commented):
Great. Take care in future when picking which files to commit. You are done with this sprint, well done.

SuWebOnes:
thank you

PR #696, opened 2025-07-24
Sheffield | 25-ITP-May | Declan Williams | Data-Groups-Sprint-2

CatchingKiyoko:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-2/debug/recipe.js
+++ b/Sprint-2/debug/recipe.js
@@ -14,6 +14,4 @@ const recipe = {
   ingredients: ["olive oil", "tomatoes", "salt", "pepper"],
 };
 
-console.log(`${recipe.title} serves ${recipe.serves}
-  ingredients:
-${recipe}`);
+console.log(`${recipe.title}\n serves ${recipe.serves}\n ingredients:\n ${recipe.ingredients[0]}\n ${recipe.ingredients[1]}\n ${recipe.ingredients[2]}\n ${recipe.ingredients[3]}`);
--- a/Sprint-2/debug/recipe.js
+++ b/Sprint-2/debug/recipe.js
@@ -1,8 +1,12 @@
 // Predict and explain first...
+// i expect the code to run but not produce the ingredients as its not calling the key pair values of ingredients array with the recipe.ingredients call
+// i also expect that each part will not be produced on a new line when called as there is no new line breaks called with the \n after each property has been called in the console log
 
 // This program should log out the title, how many it serves and the ingredients.
 // Each ingredient should be logged on a new line
 // How can you fix it?
+// to fix this you would need to change "${recipe}`" to "${recipe.ingredients}`"
+// to fix each ingredient being called individually you'd use the array call on "recipe.ingredients[]\n" to call individual elements of the array and put them on a new line
 
 const recipe = {
   title: "bruschetta",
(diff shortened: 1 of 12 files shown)


Sprint-2/debug/recipe.js
```diff
@@ -1,15 +1,17 @@
...
 const recipe = {
   title: "bruschetta",
   serves: 2,
   ingredients: ["olive oil", "tomatoes", "salt", "pepper"],
 };
 
-console.log(`${recipe.title} serves ${recipe.serves}
-  ingredients:
-${recipe}`);
+console.log(`${recipe.title}\n serves ${recipe.serves}\n ingredients:\n ${recipe.ingredients[0]}\n ${recipe.ingredients[1]}\n ${recipe.ingredients[2]}\n ${recipe.ingredients[3]}`);
```
cjyuan, line 17: How would you modify this code so that it still works if we change the number of ingredients?

Sprint-2/implement/contains.test.js
```diff
@@ -20,16 +20,27 @@ as the object doesn't contains a key of 'c'
...
 // Then it should return false
+test("when given an object with non existent properties returns false", () => {
+    expect(contains({a: 1, b: 2,}, "c")).toBe(false);
+});
 
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+test("when given an invalid parameters like arrays returns false", () => {
+    expect(contains([1, 2, 3], 1)).toBe(false);
```
cjyuan, line 45: Can you check if your function can pass this test?
`expect(contains(null, "a").toBe(false);`

Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,14 @@
...
+function contains(obj, key) {
+    // check if "obj" is an object and not an array and we are working with plain objects
+    if (typeof obj !== "object" || Array.isArray(obj)) return false;
 
+    // loop over each property name in the object
+    for (let properties in obj){
+
+        // check that if each of the properties names matches what were looking for in the key and return true if matches
+        if (properties === key) return true;
+    }
```
cjyuan, line 10: **Suggestion**
You may want to check out the following two approaches for determining if an object contains a property:
```
```

Sprint-2/implement/querystring.js
```diff
@@ -6,7 +6,12 @@ function parseQueryString(queryString) {
...
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
+    // split only the first "=" to handle cases where the value contains "="
+    const [key, ...rest] = pair.split("=");
+
+    // rejoin the rest incase there are "=" in the value
+    const value = rest.join("=");
+
     queryParams[key] = value;
   }
```
cjyuan, line 16: Please note that in real querystring, both `key` and `value` are ***percent-encoded*** or ***URL encoded*** in the URL. For example, the string "5%" will be encoded as "5%25". So to get the actual value of "5%25" (whether it is a key or value in the querystring), you should call a function to decode it.

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,20 @@
...
+    const itemCount = {};
+
+    // loop over each item in the array
+    for (const item of array){
+        // if the item hasn't been counted yet treat it as 0
+        itemCount[item] = (itemCount[item] || 0) + 1; // then adds 1 to either start or increment the count for this item
+    }
+
+    return itemCount;
+}
```
cjyuan, line 18: Does the following function call returns the value you expect?
```
```

PR #699, opened 2025-07-25
ZA-ITP-May-2025 | Christian Mayamba | Module-Data-Groups | Week 2

c-mayamba:
## Learners, PR Template

Self checklist

- [ ] I have titled my PR with Region | Cohort | FirstName LastName | Sprint | Assignment Title
- [ ] My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/reviewing/style-guide/)

## Changelist

Briefly explain your PR.

## Questions


Sprint-2/implement/contains.test.js
day-lee, line 1: Once you refactor the contains function, please add a test case for it too!

Sprint-2/implement/lookup.js
```diff
@@ -1,5 +1,13 @@
-function createLookup() {
+function createLookup(nestedArray) {
   // implementation here
+  let finalObject = {};
+
+  for (let i = 0; i < nestedArray.length; i++) {
+    const key = nestedArray[i][0];
+    const value = nestedArray[i][1];
+    finalObject[key] = value;
+  }
```
day-lee, line 9: Good job! Code looks good, as a challenge, would you like to try to refactor it to use an modern array method for looping over an element?

Sprint-2/implement/contains.js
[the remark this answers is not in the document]
c-mayamba: Thanks for the feedback.

Sprint-2/implement/querystring.js

PR #714, opened 2025-07-27
LONDON | MAY 2025 | JESUS DEL MORAL | SPRINT 3 | QUOTE AUTO GENERATOR

delmorallopez:
The PR include two files: 

- Index.html
- quote.js


hkavalikas:
Part of the requirements is to update the page title to `Alarm clock app`. What do you need to change to make this happen?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,12 +1,71 @@
-function setAlarm() {}
+timeSetup = document.querySelector("#alarmSet"); // Access to alarmset
```
hkavalikas, line 1: It's very easy to accidentally create globals when we aren't using [strict mode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode), how can we avoid it and be more explicit with the variable/constant declaration?

delmorallopez: The Title in Index.html is already "Quote generator app" 
  <head>
    <meta charset="UTF-8" />
    <title>Quote generator app</title>
  </head>

Sprint-3/quote-generator/index.html
```diff
@@ -3,13 +3,57 @@
...
     <script defer src="quotes.js"></script>
   </head>
   <body>
-    <h1>hello there</h1>
     <p id="quote"></p>
     <p id="author"></p>
-    <button type="button" id="new-quote">New quote</button>
+    <button type="button" id="new-quote">New Quote</button> <br /> 
+    <button type="button" id="quote-generator">Start to Auto-Generate</button>
+    <script src="quotes.js"></script>
```
hkavalikas, line 14: Can you walk me through why we need this line?

Sprint-3/quote-generator/index.html
```diff
@@ -3,13 +3,57 @@
...
   </body>
+
+  <style>
```
hkavalikas, line 17: To keep the styling a bit cleaner (and more reusable), how can we move this out of the HTML file and into its own file?

hkavalikas:
The requirements mention `the generator should display "auto-play:ON" somewhere on the page` , where would you add it ?

hkavalikas:
Also optional but the requirements mention `the generator should pick and display a new quote every 60 seconds. (When testing you will probably want to reduce this to every 5 seconds.)` assuming we are done testing and are ready to deploy our new app, let's increase it to match the requirements when we are ready.

hkavalikas (changes requested):
Awesome job 👏 only a few minor comments

Sprint-3/quote-generator/index.html
delmorallopez: I removed that line 14 
I didn't realise I have already linked with quotes in Line 7

Sprint-3/quote-generator/index.html
delmorallopez: I have moved style code to the file style.css 
I linked Index.html to style.css 

hkavalikas:
Awesome job 👏

PR #716, opened 2025-07-28
London|May-2025|Fatma_Degirmenci|Module-Data-Groups|Sprint-2

fatmaevin:
Self checklist

## Changelist
- Added `contains`, `createLookup`, `parseQueryString`, `tally`, and `invert` functions
- Implemented corresponding test cases for each
- Fixed bugs and handled edge cases

### DIFF
--- a/Sprint-2/implement/querystring.js
+++ b/Sprint-2/implement/querystring.js
@@ -8,13 +8,14 @@ function parseQueryString(queryString) {
   for (const pair of keyValuePairs) {
     const [key, ...rest] = pair.split("=");
     const value = rest.join("=");
-    queryParams[key] = value;
+    const decodedKey = decodeURIComponent(key);
+    const decodedValue = decodeURIComponent(value);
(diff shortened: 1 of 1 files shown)


Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,17 @@
-function contains() {}
+function contains(obj, key) {
+  if (
+    typeof obj !== "object" ||
```
day-lee, line 3: In this if statement, it looks like you check the type of obj parameter. Array is an object type as well. You can check this with ` console.log(typeof obj)` 
Below test code returns false not because it's not an object but there is no key available. 
```
test("Given invalid parameters like an array", () => {
  expect(contains(["a", "b", 2, 3, 4], "b")).toEqual(false);
});
```
can you refactor the code so you can check if the obj is an array, then exit early as false?

Sprint-2/implement/querystring.js

day-lee:
Good job 💯

PR #722, opened 2025-07-29
London | 25-ITP-May | Halimatou Saddiyaa | Sprint 3 | Reading List App

Halimatou-saddiyaa:
## Learners, PR Template

Self checklist

## Changelist

I've implemented the readingList function and updated the Css file accordingly.

## Questions

### DIFF
--- a/Sprint-3/reading-list/script.js
+++ b/Sprint-3/reading-list/script.js
@@ -1,3 +1,22 @@
+function readingList(books) {
+  const bookList = document.getElementById("reading-list");
+  bookList.innerHTML = ""; // Clear any previous content
+
+  books.forEach((book) => {
+    const bookItem = document.createElement("li");
+    bookItem.classList.add("book");
+    bookItem.classList.add(book.alreadyRead ? "read" : "unread");
+
+    bookItem.innerHTML = `
+          <h3>${book.title}</h3>
+          <p>by ${book.author}</p>
+          <img src="${book.bookCoverImage}" alt="Cover of ${book.title}">
+        `;
+
+    bookList.appendChild(bookItem);
+  });
+}
+
 // for the tests, do not modify this array of books
 const books = [
   {
@@ -20,4 +39,4 @@ const books = [
     bookCoverImage: "https://blackwells.co.uk/jacket/l/9780135957059.jpg",
   },
 ];
-
+readingList(books);
--- a/Sprint-3/reading-list/script.js
+++ b/Sprint-3/reading-list/script.js
@@ -1,6 +1,6 @@
 function readingList(books) {
   const bookList = document.getElementById("reading-list");
-  bookList.innerHTML = ""; // Clear any previous content
+  bookList.innerHTML = "";
 
   books.forEach((book) => {
     const bookItem = document.createElement("li");
--- a/Sprint-3/reading-list/style.css
+++ b/Sprint-3/reading-list/style.css
@@ -1,3 +1,60 @@
+html,
+body {
+  font-family: Arial, sans-serif;
+  background-color: #f4f4f4;
+  margin: 0;
+  padding: 2rem;
+}
+
+h1 {
+  text-align: center;
+  margin-bottom: 2rem;
+}
+
+#reading-list {
+  list-style: none;
+  padding: 0;
+  display: flex;
+  flex-direction: column;
+  gap: 1.5rem;
+  justify-content: center;
+}
+
+.book {
+  width: 100%;
+  max-width: 600px;
+  padding: 1rem;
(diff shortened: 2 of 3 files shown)


Sprint-3/reading-list/script.js
day-lee, line 1: Have you run the tests? It renders as expected but doesn't pass two tests at the moment. Please refactor the code accordingly so it passes the test.

Sprint-3/reading-list/script.js
```diff
@@ -1,3 +1,22 @@
+function readingList(books) {
```
day-lee, line 1: It is a best practice to name the function in a 'verb + something' format. Can you think of a more descriptive name that tells what the function does?

Sprint-3/reading-list/style.css
```diff
@@ -1,3 +1,60 @@
...
+  border-radius: 8px;
+  color: #fff;
+  box-shadow: 0 2px 6px rgba(0, 0, 0, 0.1);
+  display: flex;
+  gap: 1rem;
+  align-items: center;
+}
+
+.read {
+  background-color: #4caf50;
```
day-lee, line 36: Using a hex code for colour is a good practice, but I guess this would be a reason why test fails... please look into the test.

Sprint-3/reading-list/script.js
```diff
@@ -1,3 +1,22 @@
...
+  const bookList = document.getElementById("reading-list");
+  bookList.innerHTML = "";
```
day-lee, line 3: Why do you need to add an empty string here? is there a specific reason why? if not, this can be removed.

Sprint-3/reading-list/script.js
```diff
@@ -1,3 +1,22 @@
...
+
+  books.forEach((book) => {
+    const bookItem = document.createElement("li");
+    bookItem.classList.add("book");
+    bookItem.classList.add(book.alreadyRead ? "read" : "unread");
+
+    bookItem.innerHTML = `
+          <h3>${book.title}</h3>
+          <p>by ${book.author}</p>
+          <img src="${book.bookCoverImage}" alt="Cover of ${book.title}">
```
day-lee, line 13: It might be confusing why the the second test: `each book in the list has an image`doesn't pass even though it does exists... try to figure out why first, and if you are not sure please let me know!

Sprint-3/reading-list/script.js
```diff
@@ -1,3 +1,22 @@
+function readingList(books) {
+  const bookList = document.getElementById("reading-list");
+  bookList.innerHTML = "";
+
+  books.forEach((book) => {
+    const bookItem = document.createElement("li");
+    bookItem.classList.add("book");
+    bookItem.classList.add(book.alreadyRead ? "read" : "unread");
+
+    bookItem.innerHTML = `
```
day-lee, line 10: Please try not to use innerHTML, it has a potential security risk, FYI, check this [link](https://medium.com/@sparklewebhelp/understanding-the-risks-of-using-inner-html-in-web-development-30d4fa67f815)

Sprint-3/reading-list/script.js
Halimatou-saddiyaa: I've refactored the code so that it passes all the tests.

Sprint-3/reading-list/script.js
Halimatou-saddiyaa: I’ve replaced the function name from readingList to createReadingList to follow the best practices of functions naming.

Sprint-3/reading-list/style.css
Halimatou-saddiyaa: I’ve changed the color to ‘red’ and ‘green’ as it is defined as a condition for the test to pass.

Sprint-3/reading-list/script.js
Halimatou-saddiyaa: It was to prevent duplicates, but as I'm only calling the function once, it is unnecessary here as you mentioned. I have removed it.

Sprint-3/reading-list/script.js
Halimatou-saddiyaa: I’ve looked at the test and realised that it was the ‘alt’ attribute that made the test failed. I’ve removed it and tested again, and the test now passes.

Sprint-3/reading-list/script.js
Halimatou-saddiyaa: I’ve read the article. It’s very interesting and I’ll make sure to not use innerHTML from now onwards . I’ve used text content to replace innerHTML

day-lee:
I've reviewed the code, and it looks great! Well done on the refactoring. Keep up the good work!

PR #755, opened 2025-08-09
ZA Cape Town | ITP-May-2025 | Dawud Vermeulen | Sprint 2 | Module-Data-Groups

Dave-Vermeulen:
## Learners, PR Template

Self checklist

## Changelist

completed the tasks within: 
- Debug
- Implement
- Interpret

## Questions

Please review my PR 🙏🏽

### DIFF
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,3 +1,10 @@
-function contains() {}
+function contains(obj, property) {
 
-module.exports = contains;
+    if(obj === null || obj === undefined){
+        return false
+    } 
+    return obj[property] !== undefined;
+}
+
+module.exports = contains; 
+// work done.
--- a/Sprint-2/implement/querystring.js
+++ b/Sprint-2/implement/querystring.js
@@ -1,14 +1,18 @@
 function parseQueryString(queryString) {
   const queryParams = {};
-  if (queryString.length === 0) {
+  if (queryString === "") { //removed the .length method
     return queryParams;
   }
   const keyValuePairs = queryString.split("&");
 
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
+    //const [key, value] = pair.split("="); //this line is making kak and its so frustrating cause it assigns value ='x' (test case 1) and leaves the rest of the equation.
+    const parts = pair.split("=");
+    const key = parts[0];
+    const value = parts.length > 1 ? parts.slice(1).join('=') : ''; //clunky but it works to split them and stitch it together again
     queryParams[key] = value;
   }
+  
 
   return queryParams;
 }
--- a/Sprint-2/implement/tally.js
+++ b/Sprint-2/implement/tally.js
@@ -1,3 +1,14 @@
-function tally() {}
+function tally(items) {
+    if (!Array.isArray(items)){
+        throw new TypeError('that aint no array bay bay'); //okay so this is the biggie smalls and its working
+    }
+    const counts = {};
+
+    for (const item of items){
+        counts[item] = (counts[item] || 0) +1; //simple sexy core logic
+    }
+    return counts;
+}
 
 module.exports = tally;
+//no updates. works well
--- a/Sprint-2/interpret/invert.js
+++ b/Sprint-2/interpret/invert.js
@@ -10,20 +10,30 @@ function invert(obj) {
   const invertedObj = {};
 
   for (const [key, value] of Object.entries(obj)) {
-    invertedObj.key = value;
+    const numKey = Number(key);
+
+    if (!isNaN(numKey) && String(numKey) === key) {
+      invertedObj[value] = numKey;
+    } else {
+      invertedObj[value] = key; //convert key to number if it is a number
+    //how do i convert it to a number if it is a number? 
+    //if the key is a number, convert it to a number, otherwise keep it as a string
+      }
   }
 
   return invertedObj;
 }
+module.exports = invert;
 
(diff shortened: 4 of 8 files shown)


Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,10 @@
...
 
-module.exports = contains;
+    if(obj === null || obj === undefined){
+        return false
+    } 
+    return obj[property] !== undefined;
+}
+
+module.exports = contains; 
+// work done.
```
cjyuan, line 10: What do you expect the from the following function calls?
```javascript
contains("ABC", "1");
contains(["A", "B", "C"], "1");
contains(123, "1");
contains(true, "1");
contains({ key: undefined }, "key");
```

Suggestion:, Look up `JS "in" operator vs Object.hasOwn`.

Sprint-2/implement/querystring.js
```diff
@@ -1,16 +1,20 @@
...
   }
   const keyValuePairs = queryString.split("&");
 
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
+    //const [key, value] = pair.split("="); //this line is making kak and its so frustrating cause it assigns value ='x' (test case 1) and leaves the rest of the equation.
+    const parts = pair.split("=");
+    const key = parts[0];
+    const value = parts.length > 1 ? parts.slice(1).join('=') : ''; //clunky but it works to split them and stitch it together again
     queryParams[key] = value;
```
cjyuan, line 13: Please note that in real querystring, both `key` and `value` are ***percent-encoded*** or ***URL encoded*** in the URL. For example, the string "5%" will be encoded as "5%25". So to get the actual value of "5%25" (whether it is a key or value in the querystring), you should call a function to decode it.

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,14 @@
...
+    if (!Array.isArray(items)){
+        throw new TypeError('that aint no array bay bay'); //okay so this is the biggie smalls and its working
+    }
+    const counts = {};
+
+    for (const item of items){
+        counts[item] = (counts[item] || 0) +1; //simple sexy core logic
+    }
+    return counts;
+}
```
cjyuan, line 11: Does the following function call returns the value you expect?
```
```

Sprint-2/interpret/invert.js
```diff
@@ -10,20 +10,30 @@ function invert(obj) {
   const invertedObj = {};
 
   for (const [key, value] of Object.entries(obj)) {
-    invertedObj.key = value;
+    const numKey = Number(key);
```
cjyuan, line 13: We don't have to convert the key to a number when it looks like a number.

Sprint-2/interpret/invert.js
```diff
@@ -10,20 +10,30 @@ function invert(obj) {
...
+module.exports = invert;
 
-// a) What is the current return value when invert is called with { a : 1 }
-
-// b) What is the current return value when invert is called with { a: 1, b: 2 }
-
+// a) What is the current return value when invert is called with { a : 1 } 
+// {'1':'a'}
+// b) What is the current return value when invert is called with { a: 1, b: 2 } 
+// { '1': 'a', '2': 'b'}
```
cjyuan, line 31: These are the expected return value from the original code, not the actual return value.

Sprint-2/interpret/invert.js
```diff
@@ -10,20 +10,30 @@ function invert(obj) {
...
 // c) What is the target return value when invert is called with {a : 1, b: 2}
-
+// {1 : a, 2 : b }
 // c) What does Object.entries return? Why is it needed in this program?
-
+// its a method that takes a obj and returns the [key, value] pairs.
 // d) Explain why the current return value is different from the target output
-
+// cause its converted to strings, the key must be a int and the value must be a string.
```
cjyuan, line 37: The question is asking, why with the original code, `invert({a: 1})` does not return `{'1': `a`}` (or `{1:`a`}`).

Note: The property name (key) is always a string. Even when we express the key as a number (e.g., `{ 1 : 'a' }`), the key will still be converted to a string `'1'`.

cjyuan (commented):
- You missed updating the `.test.js` files

Sprint-2/implement/querystring.js
Dave-Vermeulen: thats for the guidance

Sprint-2/implement/contains.test.js
```diff
@@ -1,35 +1,74 @@
...
+        expect(contains({a:1}, "")).toBe(false);
+    });
+
+    it('should handle non-object inputs gracefully', () => {
+        expect(contains(123, 'a')).toBe(false); // Number
+        expect(contains("hello", 'a')).toBe(false); // String
+        expect(contains([1, 2, 3], 'a')).toBe(false); // Array
+        expect(contains(true, 'a')).toBe(false)
+        expect(contains(NaN, 'a')).toBe(false)
+    });
```
cjyuan, line 35: `contains([1, 2, 3], "a")` could also return false because "a" is not a property (or key) of `[1, 2, 3]`.
However, "0", "1", "2" are keys of `[1, 2, 3]`, so it is better to specify the test as
`expect(contains([1, 2, 3], "1")).toBe(false);` (to ensure you are checking what you describe)

You would also need to check if the first parameter is an array in your function.

cjyuan (commented):
Have you also gone through my comments on  `Sprint-2/interpret/invert.js` in the previous review?

PR #786, opened 2025-09-13
London | 25-ITP-May | Houssam Lahlah | Sprint 2 | Sprint-2

HoussamLh:
**_Self checklist_**

**_Changelist_**

**_1. Debug_**

- Fixed issues in object property access and iteration:
- Correctly logged object property values (author object)
- Accessed specific properties instead of using incorrect indices (address object)
- Logged array items within objects properly (recipe object)
- Explanations and predictions added as comments in the code.

**_2. Implement_**

- Implemented functions with corresponding tests:
- contains(obj, prop) → checks if an object contains a specific property
- createLookup(pairs) → builds a key-value lookup from an array of pairs
- parseQueryString(queryString) → parses URL query strings, handles = in values, duplicates, and percent-decoding
- tally(items) → counts occurrences of each item in an array
- All tests cover normal cases and edge cases.

**_3. Interpret_**

- invert(obj) → swaps keys and values of an object
- Answers to questions (a → e) are included as comments explaining the original problem and the fix.

**_4. Stretch_**

- countWords(str) → counts word occurrences in a string
- Advanced: removes punctuation, ignores case, sorts results by frequency
- calculateMode(list) → returns the most frequent number
- Refactored into helper functions: buildFrequencyMap and findMode
- totalTill(till) → calculates the total value of a till in pounds
- Fixed parsing of coin strings (e.g., "50p" → 50), handles empty till
- Tests added for all functions to validate expected behavior and edge cases.

**_Questions_**

1. Why did for (const value of author) fail when iterating over an object?
2. How does contains(obj, prop) check for the existence of a property?
3. Why did invert({ a: 1, b: 2 }) originally return { key: 2 } instead of { "1": "a", "2": "b" }?
4. Why did totalTill originally return NaN when multiplying coin * quantity?

### DIFF
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,3 +1,12 @@
-function contains() {}
+function contains(obj, prop) {
+  // Check if the input is a valid object and prop is a string
+  if (obj === null || typeof obj !== "object" || Array.isArray(obj)) {
+    return false;
+  }
+
+  // Use Object.prototype.hasOwnProperty to check if property exists
+  return Object.prototype.hasOwnProperty.call(obj, prop);
+}
 
 module.exports = contains;
+
--- a/Sprint-2/implement/tally.js
+++ b/Sprint-2/implement/tally.js
@@ -1,3 +1,18 @@
-function tally() {}
+function tally(items) {
+  if (!Array.isArray(items)) {
+    throw new Error("Input must be an array");
+  }
+
+  const result = {};
+  for (const item of items) {
+    if (Object.prototype.hasOwnProperty.call(result, item)) {
+      result[item] += 1;
+    } else {
+      result[item] = 1;
+    }
+  }
+  return result;
+}
 
 module.exports = tally;
+
--- a/Sprint-2/stretch/count-words.js
+++ b/Sprint-2/stretch/count-words.js
@@ -1,28 +1,70 @@
+// Basic implementation of word counting
+
+
+// Given a string of words
+// When the string is passed to countWords
+// Then it should return an object with the count of each word
+
+// Example:
+// Given "you and me and you"
+// When countWords is called on this string
+// Then it should return { you: 2, and: 2, me: 1 }
+
+function countWords(str) {
+  const counts = {};
+
+  // ✅ Basic: split on spaces
+  const words = str.split(" ");
+
+  for (const word of words) {
+    if (!counts[word]) {
+      counts[word] = 1;
+    } else {
+      counts[word]++;
+    }
+  }
+
+  return counts;
+}
+
+module.exports = countWords;
+
+
+// Advanced challenges are implemented in countWordsAdvanced
 /*
-  Count the number of times a word appears in a given string.
+### Advanced challenges (implemented separately)
 
-  Write a function called countWords that
-    - takes a string as an argument
-    - returns an object where
-          - the keys are the words from the string and
-          - the values are the number of times the word appears in the string
+1. Remove punctuation
+   → Use regex to strip punctuation before splitting.
 
-  Example
-  If we call countWords like this:
+2. Ignore case
+   → Convert the string to lowercase before counting.
 
-  countWords("you and me and you") then the target output is { you: 2, and: 2, me: 1 }
+3. Order results
+   → Convert the counts object into an array, sort it, and return it in order.
 
-  To complete this exercise you should understand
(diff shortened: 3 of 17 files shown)


Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,12 @@
-function contains() {}
+function contains(obj, prop) {
+  // Check if the input is a valid object and prop is a string
+  if (obj === null || typeof obj !== "object" || Array.isArray(obj)) {
+    return false;
+  }
+
+  // Use Object.prototype.hasOwnProperty to check if property exists
+  return Object.prototype.hasOwnProperty.call(obj, prop);
```
cjyuan, line 8: There is another (more compact) function in `Object` we can use. Can you find out how we can rewrite line 8 with less code.

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,18 @@
-function tally() {}
+function tally(items) {
+  if (!Array.isArray(items)) {
+    throw new Error("Input must be an array");
+  }
+
+  const result = {};
+  for (const item of items) {
+    if (Object.prototype.hasOwnProperty.call(result, item)) {
```
cjyuan, line 8: There is a way to create an empty object with **no inherited properties**.

If `result` is such an object, then we can simplify line 8 as `if (result[item]) {`.

Sprint-2/stretch/count-words.js
```diff
@@ -1,28 +1,70 @@
...
+  // 1. Remove punctuation
+  const noPunctuation = str.replace(/[.,!?]/g, "");
+
+  // 2. Ignore case
+  const words = noPunctuation.toLowerCase().split(/\s+/);
+
+  for (const word of words) {
+    if (!word) continue; // skip empty strings
+    counts[word] = (counts[word] || 0) + 1;
+  }
```
cjyuan, line 60: Can you check if this function returns what you expect in the following function call?

```
```

Sprint-2/stretch/till.js
```diff
@@ -8,12 +8,17 @@ function totalTill(till) {
   let total = 0;
 
   for (const [coin, quantity] of Object.entries(till)) {
-    total += coin * quantity;
+    // coin is a string like "1p", "20p"
+    // We need to strip the "p" and convert it into a number of pennies
+    const coinValue = parseInt(coin.replace("p", ""), 10);
```
cjyuan, line 13: We could also rewrite line 13 as:
`const coinValue = parseInt(coin);`

 Can you figure out why?

cjyuan (commented):
Code is very solid! I only have a few suggestions.

cjyuan:
> Why did for (const value of author) fail when iterating over an object?

Have you try asking AI?

> How does contains(obj, prop) check for the existence of a property?

The spec was not clear how `contains()` should check for the existence of a property. When the spec is not clear, we can either ask for clarification (especially when the ambiguity can lead to serious consequence)  or make reasonable assumption and write down our assumption.

> Why did invert({ a: 1, b: 2 }) originally return { key: 2 } instead of { "1": "a", "2": "b" }?

It seems you had figured out the answer.

> Why did totalTill originally return NaN when multiplying coin * quantity?

You can also try AI for questions like this. :)

Sprint-2/implement/contains.js
HoussamLh: Hi **_cjyuan_**, 
**_Thank you_** for the feedback! 
I updated the code to use **Object.hasOwn(obj, prop)** instead of Object.prototype.hasOwnProperty.call(obj, prop). 
It’s definitely more **compact** and **readable**.

Sprint-2/implement/tally.js
HoussamLh: Hi **_cjyuan_**,
**_Thank you_** for pointing this out! 
I changed result = {} to **result = Object.create(null)**, so the object has no **inherited** properties. This allowed me to simplify the condition to if (result[item]) { as suggested.

Sprint-2/stretch/till.js
HoussamLh: Hi **_cjyuan_**,
**Thanks**! I see why — parseInt("20p", 10) returns 20 since it stops parsing at the "p". 
That makes .replace("p", "") unnecessary, so I **simplified** the code.

cjyuan:
Everything looks great. Well done!

PR #791, opened 2025-10-06
London | 25-ITP-Sep | Adnaan Abo | Sprint 1 | Sprint-1

AdnaanA:
Self checklist

Change list:

In sum.js I have implemented a function that sums only the numeric elements of an array, ignoring non-numerical values, and handles empty arrays, negatives, and decimals, with tests in sum.test.js.

In max.js I have implemented a function that finds the largest numeric element in an array while ignoring non-numeric values and supporting empty arrays, negatives, and decimals, tested in max.test.js.

In dedupe.js I have implemented a function that removes duplicate values from arrays of numbers or strings, preserving the first occurrence and handling empty arrays, with tests in dedupe.test.js.

In calculateMedian.js I have implemented a function that calculates the median of numeric elements from sorted or unsorted arrays of odd or even length, filtering out non-numeric values, tested in median.test.js.

In includes.js I have refactored the function that checks if a target value exists in an array using a for...of loop, supporting empty arrays, duplicates, and nulls, with tests in includes.test.js.


Sprint-1/fix/median.js
```diff
@@ -6,8 +6,29 @@
...
+  // Filter out non-numeric values
+  const numericList = list.filter(item => typeof item === 'number' && !isNaN(item));
+  if (numericList.length === 0) {
+    return null;
+  }
+  // Sort the numeric values
+  numericList.sort((a, b) => a - b);
+  const middleIndex = Math.floor(numericList.length / 2);
+  let median;
+  if (numericList.length % 2 === 0) {
```
ckirby19: This solution is already perfect, but just as an extension, could you show me how you could express this if...else statement using a [ternary](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_operator) operator?

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,24 @@
 function findMax(elements) {
+    if (elements.length === 0) {
```
ckirby19: You may have heard of the DRY (Don't Repeat Yourself) principle in coding, which emphasises reducing the amount of code that is repeated (such as the same `if (elements.length === 0)` code you have here and below). 
Although these if statements represent different cases (an empty array vs. one with all elements that are non-numeric), they can both be covered just with the if statement on line 8

ckirby19 (commented):
Very, very nice work! All your solutions make sense and your code is very clean. I have just added a few comments for extension

Sprint-1/fix/median.js
AdnaanA: Your right we can express the if...else statement using a ternary operator here is how our median calculation will look like using the ternary operator:

numericList.sort((a, b) => a - b);
const middleIndex = Math.floor(numericList.length / 2);

const median = numericList.length % 2 === 0 
    ? (numericList[middleIndex - 1] + numericList[middleIndex]) / 2 
    : numericList[middleIndex];

ckirby19 (approved):
Great work!

PR #795, opened 2025-10-15
Glasgow | 25-ITP-SEP | Shreef Ibrahim | Sprint 3| Alarm clock app

shreefAhmedM:
Self checklist

Here is a basic implementation of an alarm clock app using HTML and JavaScript

### DIFF
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -48,7 +48,10 @@ function startCountdown() {
 // set up alarm
 function setAlarm() {
   const val = parseInt(inputField.value);
-  if (!val || val <= 0) return;
+  if (!val || val <= 0) {
+    alert("Please enter a valid number of seconds.");
+    return;
+  }
 
   timeLeft = val;
   updateDisplay();


jennethydyrova (changes requested):
Really great start but there are couple of issues to address.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,25 +1,101 @@
...
+  }, 1000);
+}
 
-function setup() {
-  document.getElementById("set").addEventListener("click", () => {
-    setAlarm();
-  });
+// set up alarm
+function setAlarm() {
+  const val = parseInt(inputField.value);
```
jennethydyrova: What if `inputField.value` is empty? Can you try starting the alarm without entering anything into input field?

jennethydyrova (commented):
hi @shreefAhmedM! Really great work! I have just two small comments and it's good to go.

Optionally, when you have time, I would like you to think how you could rewrite this cleaner without having many variables in a global scope. You [can pass](https://stackoverflow.com/questions/256754/how-to-pass-arguments-to-addeventlistener-listener-function) parameters to event listener functions or perhaps restructure your functions in a way that allows passing variables as parameters.

Sprint-3/alarmclock/alarmclock.js
shreefAhmedM: When inputField.value is empty, the  parseInt("") returns NaN, 
So I have validated the condition if the value is NaN, Undefined , null or less than 0 , i used return  it will stop any further execution so no nothing will happen. I can throw an error message instead.

jennethydyrova: You mean you already have it in your code or you are going to push? I don't see validation against NaN, etc.

jennethydyrova: Ah okay, that happens 🙂

PR #797, opened 2025-10-19
NW | 25-ITP-Sep | TzeMing Ho | Sprint 2 |sprint-2-exercises

TzeMingHo:
## Learners, PR Template

Self checklist

## Changelist

I have tried to create tests for the functions, according to the TPP principles, followed by refactoring the functions to make them work.

### DIFF
--- a/Sprint-2/stretch/mode.js
+++ b/Sprint-2/stretch/mode.js
@@ -21,10 +21,10 @@ function trackFrequency(list) {
   return frequency;
 }
 
-function highestFrequency(frequencyArray) {
+function highestFrequency(frequencyMap) {
   let maxFreq = 0;
   let mode;
-  for (let [num, freq] of frequencyArray) {
+  for (let [num, freq] of frequencyMap) {
     if (freq > maxFreq) {
       mode = num;
       maxFreq = freq;
@@ -35,11 +35,13 @@ function highestFrequency(frequencyArray) {
 
 function calculateMode(list) {
   // track frequency of each value
-  const frequencyArray = trackFrequency(list);
+  const frequencyMap = trackFrequency(list);
 
   // Find the value with the highest frequency
 
-  return highestFrequency(frequencyArray);
+  return highestFrequency(frequencyMap);
 }
+
+calculateMode([1, 3, "2", 2, 3, null]);
 
 module.exports = calculateMode;


Sprint-2/debug/author.js
```diff
@@ -11,6 +13,9 @@ const author = {
   alive: true,
 };
 
-for (const value of author) {
-  console.log(value);
-}
+console.log(Object.values(author));
```
CameronDowner: What if we wanted to `console.log` each property at a time?

Sprint-2/debug/recipe.js
```diff
@@ -12,4 +16,7 @@ const recipe = {
 
 console.log(`${recipe.title} serves ${recipe.serves}
   ingredients:
-${recipe}`);
+${recipe.ingredients.map((ingredient) => `${ingredient}`).join("\n")}`);
```
CameronDowner: If we extract this value, is there a way we couple simplify this?

```javascript
`${recipe.ingredients.map((ingredient) => `${ingredient}`).join("\n")}`;
```

Sprint-2/implement/lookup.js
```diff
@@ -1,5 +1,16 @@
-function createLookup() {
-  // implementation here
+function createLookup(arr) {
```
CameronDowner: This is looking really good

Sprint-2/interpret/invert.test.js
```diff
@@ -0,0 +1,39 @@
...
+    expect(invert({ a: null, b: undefined, c: true })).toStrictEqual({
+      null: "a",
+      undefined: "b",
+      true: "c",
+    });
+    expect(invert({ a: [1], b: [2] })).toStrictEqual({
+      "[1]": "a",
+      "[2]": "b",
+    });
+    expect(invert({ a: { b: 1 }, c: { d: 2 } })).toStrictEqual({
```
CameronDowner: What's the difference between `toStrictEqual` and `toEqual` used earlier in `Sprint-2/implement/querystring.test.js`?

Sprint-2/implement/lookup.js
TzeMingHo: Yeah~

Sprint-2/interpret/invert.test.js
```diff
@@ -0,0 +1,39 @@
+const invert = require("./invert");
+
+describe("testing invert function", () => {
+  const errorMessage = "Input should be an object";
+  it("should throw error message when input is not an object", () => {
+    expect(() => invert("abc")).toThrow(errorMessage);
```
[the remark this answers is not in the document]
TzeMingHo: We put this invert function inside an arrow function because we expect it will fail with invalid data input. If the invert function is called as usual, like invert("abc"), the function will be fired up immediately. We don't have time to catch the error and compare it with an error message.
So, the () => arrow function is doing a referencing job, call the invert function when we expect it to fail.

Sprint-2/stretch/count-words.js
```diff
@@ -26,3 +26,23 @@
...
+    .replace(/[\.\,\!\?\"]/g, "")
+    .toLowerCase()
+    .split(" ");
+
+  let result = {};
+
+  for (element of strArray) {
+    result[element] = (result[element] || 0) + 1;
+  }
+  return Object.fromEntries(Object.entries(result).sort((a, b) => b[1] - a[1]));
```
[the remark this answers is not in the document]
TzeMingHo: The sort function can rearrange the words' order based on the frequency they appear. We use a desc sort to move the words that appear more to the top of the list.

CameronDowner:
@TzeMingHo Changes are looking good 🙂 

Could you also answer the questions I've asked in the comments? Not problem if you don't know all the answers, but would be good to see what you think

TzeMingHo:
> @TzeMingHo Changes are looking good 🙂
> 
> Could you also answer the questions I've asked in the comments? Not problem if you don't know all the answers, but would be good to see what you think

Sorry to get back to you late. I forgot to submit my replies. So you couldn't see them yet.

Sprint-2/stretch/mode.js
```diff
@@ -8,29 +8,38 @@
...
       maxFreq = freq;
     }
   }
-
   return maxFreq === 0 ? NaN : mode;
 }
 
+function calculateMode(list) {
+  // track frequency of each value
+  const frequencyArray = trackFrequency(list);
```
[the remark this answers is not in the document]
CameronDowner: You're right - in this case `Map` behaves like an object but has an iterator - so it has an order to the elements

Sprint-2/stretch/count-words.js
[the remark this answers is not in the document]
TzeMingHo: So, if we prefer order matters, we should choose an array for testing?

PR #802, opened 2025-10-26
London | 25-ITP-Sep |Imran Mohamed | Sprint 1 | Sprint-1 exercises 

i786m:
Self checklist

## Changelist

I have implemented the various functions and the associated tests for the functions.

## Questions

The instructions for sprint 1 say that the npm install should be run within the sprint 1 folder but when I attempt to use the Testing tab on vscode I am unable to run the tests from within and the output indicates the tests fail to run as it seems to be looking in the root folder for the node module for jest. I was able to run the tests using npm test however the testing functionality afforded by vscode was diminished by that. Could there be a better way of doing this?

### DIFF
--- a/Sprint-1/implement/sum.test.js
+++ b/Sprint-1/implement/sum.test.js
@@ -35,7 +35,7 @@ test("given an array with negative numbers, returns correct sum",()=>{
 // When passed to the sum function
 // Then it should return the correct total sum
 test("given an array with decimal numbers, returns correct sum",()=>{
-    expect(sum([1.5,2.5,3.5])).toBe(7.5);
+    expect(sum([1.5,2.5,3.5])).toBeCloseTo(7.5, 5);
 })
 
 // Given an array containing non-number values


Sprint-1/fix/median.js
```diff
@@ -6,9 +6,35 @@
...
+    const mid1 = filteredList.length / 2 - 1;
+    const mid2 = mid1 + 1;
+    const median = (filteredList[mid1] + filteredList[mid2]) / 2;
+    return median
+  }else { 
+    const mid = Math.floor(filteredList.length / 2);
+    const median = filteredList[mid];
+    return median;
+  };
+
```
cjyuan: To more consistently format the code, consider using this VSCode built-in feature to auto format the code: https://code.visualstudio.com/docs/languages/javascript#_formatting

i786m:
> Have you installed Jest first? To install Jest, in a terminal, run `npm install` in the `Sprint-1` folder. `npm` will look for `package.json` and then install all dependencies specified in `package.json`.
> 
> Only after that you can run Jest test.

I ran npm install in the sprint 1 folder as per instructions and have no issue using npm test via terminal, my question relates to running the tests via the testing tab in vscode and the respective functionality in the test files to run an individual test 
image

cjyuan:
I don't know how to use the Jest extension in VSCode.
In terminal, we could run jest test on a particular file (in Sprint-1 folder where jest is installed locally) as
```
 npm test <substring that match the filename>
``` 

For examples,
- `npm test fix` will run all the `.test.js` files inside subfolder `fix` (because all other filenames do not have such substring).
- `npm test median` will run `fix/median.test.js` (since that's the only file with the matching substring)

PR #806, opened 2025-10-30
London|ITP-September-2025|Alexandru Pocovnicu|Sprint 1|Implement

alexandru-pocovnicu:
## Learners, PR Template

Self checklist

## Changelist
I implemented functions to satisfy pre-written test requirements

### DIFF
--- a/Sprint-1/implement/sum.js
+++ b/Sprint-1/implement/sum.js
@@ -4,7 +4,7 @@ function sum(elements) {
   }
 
   const filteredElements = elements.filter(
-    (element) => typeof element === "number"
+    (element) => typeof element === "number" && !Number.isNaN(element)
   );
   if (filteredElements.length === 0) {
     return NaN;


Sprint-1/implement/dedupe.js
```diff
@@ -1 +1,11 @@
-function dedupe() {}
+function dedupe(arr) {
+  if (arr.length === 0) {
+    return [];
+  }
+  const arrCopy = arr.slice(0);
+  console.log(arrCopy);
```
cjyuan: Do you need this console.log() statement for the function to work?

Sprint-1/implement/dedupe.js
```diff
@@ -1 +1,11 @@
-function dedupe() {}
+function dedupe(arr) {
+  if (arr.length === 0) {
+    return [];
+  }
+  const arrCopy = arr.slice(0);
```
cjyuan: Line 7 will create a new array. We don't have to create another array here.

Sprint-1/implement/dedupe.test.js
```diff
@@ -16,12 +16,20 @@ E.g. dedupe([1, 2, 1]) target output: [1, 2]
...
+test("given an empty array, it returns an empty array", () => {
+  expect(dedupe([])).toEqual([]);
+});
 
 // Given an array with no duplicates
 // When passed to the dedupe function
 // Then it should return a copy of the original array
+test("Given an array with no duplicates,return a copy of the original array", () => {
+  expect(dedupe([1, "y", 4, 7])).toEqual([1, "y", 4, 7]);
+});
```
cjyuan: The current test code cannot check if the returned array is a **copy** of the original array because `toEqual()` compares objects (including arrays) by value. To illustrate,

```
  const A = [2, 3, 1];
    
```

In order to check if the returned array is a **copy** of the original array, we would need **additional** checks. 
Can you find out what code you need to add in order to ensure the returned value is not the original array?

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,17 @@
...
+    return elements[0];
+  }
+  if (elements.every((element) => typeof element !== "number")) {
+    return NaN;
+  }
+  let filteredElements = elements.filter(
+    (element) => typeof element === "number"
+  );
+  let orderedElements = filteredElements.sort((a, b) => a - b);
+  return orderedElements[orderedElements.length - 1];
```
cjyuan: - You could also consider treating an array that doesn’t contain any numbers as an empty array; both of them do not contain any numbers.

- Does your function return the value you expect from the following function calls?

```javascript

  findMax([ "Foo" ]);
  findMax([ "Foo", "Bar" ]);
  findMax([ NaN, 1, 2, 3, NaN ]);
```

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,41 @@ const sum = require("./sum.js");
...
+test("Given an array containing negative numbers,return the correct total sum", () => {
+  expect(sum([-6, -7, -2])).toEqual(-15);
+});
 
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
+test("Given an array with decimal/float numbers,return the correct total sum", () => {
+  expect(sum([-1.5, 1.5, 4.5])).toEqual(4.5);
+});
```
cjyuan: Decimal numbers in most programming languages (including JS) are internally represented in "floating point number" format. Floating point arithmetic is not exact. For example, the result of `46.5678 - 46 === 0.5678` is false because `46.5678 - 46` only yield a value that is very close to `0.5678`. Even changing the order in which the program add/subtract numbers can yield different values.

So the following could happen
```

```

Suggestion: Look up 

cjyuan: It has to do with how computer represent floating point numbers. In short, numbers that are sum of 2 to the power x, where x is an integer can normally be represented accurately as floating point numbers.  For examples, `0.5` = 2^-1, `1.5` = 2^0 + 2^-1 (I am using ^ to denote "to the power").

All the numbers you picked (-1.5, 1.5, 4.5) can be represented accurately. So there is no data loss when adding those numbers.

I mainly just want to raise the awareness of "possible loss of data when dealing with floating point numbers".

PR #808, opened 2025-10-31
Birmingham | 25-ITP-Sep | Joy Opachavalit | Sprint 2 | Sprint-2

enjoy15:
## Learners, PR Template

Self checklist

# Changelist

-Completed all excercises in Sprint-2

### DIFF
--- a/Sprint-2/interpret/invert.test.js
+++ b/Sprint-2/interpret/invert.test.js
@@ -4,12 +4,9 @@ test('inverts object with single key-value pair', () => {
   expect(invert({ a: 1 })).toEqual({ "1": "a" });
 });
 
-test('inverts object with multiple key-value pairs', () => {
-  expect(invert({ a: 1, b: 2 })).toEqual({ "1": "a", "2": "b" });
-});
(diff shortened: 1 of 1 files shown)


Sprint-2/implement/querystring.js
```diff
@@ -6,8 +6,17 @@ function parseQueryString(queryString) {
...
+    const key = parts[0];
+    
+    if (parts.length === 1) {
+      // No = found, set value to undefined
+      queryParams[key] = undefined;
+    } else {
+      // Join everything after the first = back together
+      const value = parts.slice(1).join("=");
+      queryParams[key] = value;
+    }
```
cjyuan: Please note that in real querystring, both `key` and `value` are ***percent-encoded*** or ***URL encoded*** in the URL. For example, the string "5%" will be encoded as "5%25". So to get the actual value of "5%25" (whether it is a key or value in the querystring), you should call a function to decode it.

cjyuan (commented):
Changes looks great. Well done!

PR #811, opened 2025-11-03
Glasgow | 25-ITP-Sep | Abraham-Habte | Sprint 1 | Sprint1-exercise

Abrsh100:
## Learners, PR Template

Self checklist

## Changelist

 I have done all the tasks required for this sprint

## Questions

No questions

### DIFF
--- a/Sprint-1/implement/dedupe.test.js
+++ b/Sprint-1/implement/dedupe.test.js
@@ -29,7 +29,8 @@ test("given an empty array, it returns an empty array", () => {
 test("given an array with no duplicates", () => {
   const input = [1, 2, 3, 4, 5];
   const result = dedupe(input);
-  expect(result).toStrictEqual([1, 2, 3, 4, 5]);
+  expect(result).not.toBe([1, 2, 3, 4, 5]);
+  expect(result).toEqual([1, 2, 3, 4, 5]);
 });
 
 // Given an array with strings or numbers
--- a/Sprint-1/implement/sum.js
+++ b/Sprint-1/implement/sum.js
@@ -1,11 +1,11 @@
 function sum(elements) {
-  return elements.reduce((accumulator, currentValue) => {
-    if (typeof currentValue === "number" && Number.isFinite(currentValue)) {
-      return accumulator + currentValue;
-    }
-    return accumulator;
+  if (!Array.isArray(elements)) return null;
+
+  return elements.reduce((accumulator, n) => {
+    return typeof n === 'number' && !isNaN(n) ? accumulator + n : accumulator;
   }, 0);
-}   
+}
+
    
 module.exports = sum;
 


cjyuan (commented):
Code is good. I only have few suggestions.

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,53 @@ const sum = require("./sum.js");
...
+});
 
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
+test("given an array with decimal number", () => {
+  const input = [0.1, 0.2, 0.3, 0.4];
+  const result = sum(input);
+  expect(result).toEqual(1);
+});
```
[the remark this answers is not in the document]
Abrsh100: OK, I get it and change to toBeCloseTo

cjyuan:
I wasn't asking you to change code. 

Can you respond to my questions?

Abrsh100:
yes I got it now sorry for not responding properly. 
at dedube js. I change the code becouse i want the  code in line 32 (dedube.test.js) is not a copy , which i was wrong so i serach and find out that to make sure that both of the input and result are copy of each ather i have to use toStrictEqual.

in sum.js I rewrote the code because I thought that if there is a value wich is nan the sum,test.js will throw an error but when i notice that the tests are writen to ignor the NAN and add the Number there fore i writern to my first code and this code ignor the NAN. by the .filter() in which only number are included in the summation

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,12 @@
...
+    if (typeof currentValue === "number" && Number.isFinite(currentValue)) {
+      return accumulator + currentValue;
+    }
+    return accumulator;
+  }, 0);
+}   
+   
 module.exports = sum;
+
+
```
Abrsh100: I rewrote the code because I thought that if there is a value wich is nan the sum,test.js will throw an error but when i notice that the tests are writen to ignor the NAN and add the Number there fore i writern to my first code and this code ignor the NAN. by the .filter() in which only number are included in the summation

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,12 @@
 function sum(elements) {
-}
-
+  return elements.reduce((accumulator, currentValue) => {
+    if (typeof currentValue === "number" && Number.isFinite(currentValue)) {
+      return accumulator + currentValue;
+    }
```
cjyuan: FYI, the condition can further simplified.

Abrsh100: yes Thank you for the information i use .reduce becouse it can handle the check inline some as .filter

Sprint-1/implement/dedupe.test.js
```diff
@@ -29,7 +29,8 @@ test("given an empty array, it returns an empty array", () => {
 test("given an array with no duplicates", () => {
   const input = [1, 2, 3, 4, 5];
   const result = dedupe(input);
-  expect(result).toStrictEqual([1, 2, 3, 4, 5]);
+  expect(result).not.toBe([1, 2, 3, 4, 5]);
```
cjyuan: The syntax `[1, 2, 3, 4, 5]` always creates a new array.   

On line 32, the statement is not checking if `result` and the array passed to `dedupe()` are different arrays.

PR #814, opened 2025-11-03
London|25-ITP-September|Alexandru Pocovnicu|Sprint 2

alexandru-pocovnicu:
## Learners, PR Template

Self checklist

## Changelist

implemented functions using TDD

### DIFF
--- a/Sprint-2/implement/querystring.js
+++ b/Sprint-2/implement/querystring.js
@@ -10,17 +10,17 @@ function parseQueryString(queryString) {
     if (!pair) {
       continue;
     }
-    const equalIndex = pair.indexOf("=");
-    if (equalIndex === -1) {
+    const index = pair.indexOf("=");
+    if (index === -1) {
       queryParams[pair] = "";
-      continue;
+    } else {
+      const key = pair.slice(0, index);
+      const value = pair.slice(index + 1);
+      queryParams[key] = value;
     }
-    const key = pair.substring(0, equalIndex);
-    const value = pair.substring(equalIndex + 1);
-    queryParams[key] = value;
   }
-
   return queryParams;
 }
+console.log(parseQueryString("equation=x=y+1&sound=none"));
 
 module.exports = parseQueryString;


Sprint-2/implement/querystring.js
```diff
@@ -3,14 +3,30 @@ function parseQueryString(queryString) {
   if (queryString.length === 0) {
```
takanocap: is there a better way to check this and exit or handle the error if its not a valid querystring?

Sprint-2/implement/querystring.js
```diff
@@ -3,14 +3,30 @@ function parseQueryString(queryString) {
...
     return queryParams;
   }
+
   const keyValuePairs = queryString.split("&");
 
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
-    queryParams[key] = value;
+    if(pair===""){
```
takanocap: can this be shorter or tested in a different way ? investigate how to test if string has a empty or falsy value

alexandru-pocovnicu: if i use array destructuring it doesn't pass the tests as some of the values have "=" in them, or at least i don't know how to do it. some pointers would be much appreciated , thank you

Sprint-2/implement/contains.test.js
```diff
@@ -16,20 +16,39 @@ as the object doesn't contains a key of 'c'
 // Given a contains function
 // When passed an object and a property name
 // Then it should return true if the object contains the property, false otherwise
+test("if the object contains the property, return 'true'", () => {
+  expect(contains({ a: 1 }, "a")).toEqual(true);
+});
+test("if the object contains the property, return 'true'", () => {
+  expect(contains({ a: 1, "a,s,3": "op" }, ["a", "s", 3])).toEqual(true);
+});
```
illicitonion: I feel like this is slightly surprising behaviour - the object doesn't really contain that array, it contains a stringification of that array. But this works, so I guess it is ok to test for...

I would probably use a different name for the test, though, to make it clear how it's different from the test on line 19.

PR #817, opened 2025-11-05
Glasgow | 25-ITP-SEP | Hanna Mykytiuk | Sprint 1 | Coursework

HannaOdud:
## Learners, PR Template

Self checklist

## Changelist

Sprint 1 exercises completed.

## Questions

No questions

### DIFF
--- a/Sprint-1/implement/dedupe.test.js
+++ b/Sprint-1/implement/dedupe.test.js
@@ -16,12 +16,29 @@ E.g. dedupe([1, 2, 1]) target output: [1, 2]
 // Given an empty array
 // When passed to the dedupe function
 // Then it should return an empty array
-test.todo("given an empty array, it returns an empty array");
+test("given an empty array, it returns an empty array", () => {
+  const currentOutput = dedupe([]);
+  const targetOutput = [];
+  expect(currentOutput).toEqual(targetOutput);
+});
 
 // Given an array with no duplicates
 // When passed to the dedupe function
 // Then it should return a copy of the original array
-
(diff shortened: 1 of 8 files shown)


Sprint-1/implement/dedupe.test.js
```diff
@@ -16,12 +16,29 @@ E.g. dedupe([1, 2, 1]) target output: [1, 2]
...
 
 // Given an array with no duplicates
 // When passed to the dedupe function
 // Then it should return a copy of the original array
-
+test("given an array with no duplicates, it should return a copy of the original array", () => {
+  const currentOutput = dedupe([5, 1, 2, 3, 8]);
+  const targetOutput = [5, 1, 2, 3, 8];
+  expect(currentOutput).toEqual(targetOutput);
+});
```

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,21 @@
...
+    let isNumbersPresent = elements.some(item => typeof item === 'number' && !isNaN(item));
+    for (let i =0; i<elements.length; i++){ //sum of elements
+        if (typeof elements[i] === 'number'){ //check if element is number
+            sum += elements[i];
+        }
+        if (isNumbersPresent == false && !isNaN(elements[i])){ //1) check if array doesnt have numbers; 2)check if element can be conversted to number
+            sum += parseFloat(elements[i]); // add converted to num element
+        }
+    }
+    return sum
```
cjyuan: Does the following function calls return the value you expected?

```
sum([NaN, 1]);
sum(["A"]);
sum(["10", 1]);
sum(["10", "1"]);
```

PR #820, opened 2025-11-07
Glasgow | 25-ITP-Sep | Abraham-Habte | Sprint 2 | Sprint2-exercise

Abrsh100:
## Learners, PR Template

Self checklist

## Changelist

I fix the errors incounterd in the function as well as in the test and creat new function to pass the tests

## Questions

no questions

### DIFF
--- a/Sprint-2/implement/querystring.js
+++ b/Sprint-2/implement/querystring.js
@@ -1,23 +1,21 @@
 function parseQueryString(queryString) {
+  const queryParams = {};
+  if (!queryString) return queryParams;
 
-    const queryParams = {};
-    if (!queryString ) return queryParams;
+  const keyValuePairs = queryString.split("&");
 
-     const keyValuePairs = queryString.split("&");
+  for (const pair of keyValuePairs) {
+    const index = pair.indexOf("=");
+    if (index === -1) {
+      queryParams[decodeURIComponent(pair)] = undefined;
+    } else {
+      const key = decodeURIComponent(pair.slice(0, index));
+      const value = decodeURIComponent(pair.slice(index + 1));
+      queryParams[key] = value;
+    }
 
-     for (const pair of keyValuePairs) {
-     const index = pair.indexOf("=");
-     if (index === -1) {
-      queryParams[decodeURIComponent(pair)] = undefined
-     } else {
-        const key = decodeURIComponent (pair.slice(0, index));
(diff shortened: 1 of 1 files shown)


Sprint-2/debug/author.js
```diff
@@ -10,7 +10,8 @@ const author = {
   age: 40,
   alive: true,
 };
+console.log(author)
```
cjyuan: This works (for displaying the properties) to certain extent.

However, I think the objective of this exercise is to practice how to **iterate through all the property values of an object**. Can you figure out how to use a loop to accomplish that?

Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,9 @@
-function contains() {}
-
+function contains(object, properityValue){ 
+    if (typeof object !== 'object' || object === null || Array.isArray(object)) {
+        return false;
+    }   
+    
+    return properityValue in object;
```
cjyuan: **Suggestion**
Consider the following two approaches for determining if an object contains a property:
```
```

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,14 @@
...
+        throw new Error("Input must be an array");
+    }
+
+   const counts = {};
+    for (let item of input) {
+        counts[item] = (counts[item] || 0) + 1;
+    }
+
+    return counts;
+}
```
```
```

cjyuan (commented):
- Code is very solid. 

- Indentation is a bit off in some of the files. Can you improve the indentation on these files?

Sprint-2/debug/author.js
Abrsh100: got it

cjyuan:
When you make changes based on review comments, please respond directly to each comment. Even a short acknowledgement is helpful. This makes the review process clearer, shows your reasoning, and ensures that nothing gets overlooked. Code review is a conversation, not just an edit pass.

Sprint-2/implement/querystring.js
```diff
@@ -1,16 +1,22 @@
...
+    } else {
+        const key = pair.slice(0, index);
+        const value = pair.slice (index + 1);
+        queryParams[key] = value;
+    }
+    return queryParams;
 
-  return queryParams;
+    }
 }
```
cjyuan: - No all keys and values are properly URL-decoded.

- The indentation is a bit off.

PR #833, opened 2025-11-12
NW | 25-ITP-Sep| Ahmad Hmedan | Sprint 1 | Sprint 1 Coursework

AhmadHmedann:
## Learners, PR Template

Self checklist

## Changelist

I have implemented the sum, max, dedupe, and median functions, and written comprehensive tests for each. I also refactored the includes function.

## Questions
 const numericList = list.filter(
    (item) => typeof item === "number" && !isNaN(item)
  );
I know how to use it and why, but I don’t know how it works. When it says it must be an array, does that mean the structure should be an array like const x = [...]? and  const x = 23 is not an array, right?

### DIFF
--- a/Sprint-1/implement/dedupe.js
+++ b/Sprint-1/implement/dedupe.js
@@ -1,17 +1,8 @@
-function dedupe(elements) {
-  const elementsClone = [...elements];
-  if (elementsClone.length === 0) return elementsClone;
 
-  for (let i = 1; i < elementsClone.length; i++) {
-    for (let j = i; j > 0; j--) {
-      if (elementsClone[i] === elementsClone[j - 1]) {
-        elementsClone.splice(i, 1);
-        i--; // because I deleted one element from my array
-        break;
-      }
-    }
-  }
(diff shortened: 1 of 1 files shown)


Sprint-1/implement/dedupe.js
```diff
@@ -1 +1,17 @@
...
+
+  for (let i = 1; i < elementsClone.length; i++) {
+    for (let j = i; j > 0; j--) {
+      if (elementsClone[i] === elementsClone[j - 1]) {
+        elementsClone.splice(i, 1);
+        i--; // because I deleted one element from my array
+        break;
+      }
+    }
+  }
```
cjyuan: You approach works. Here are some suggestions for you to consider:

- Deleting an element from an array is a relatively costly operation. A more efficient alternative is to append an element to a new array only if the element does not exist in the array.

- Most modern programming languages offer a built-in class that implements the Set data structure. It is optimized to keeps only unique items. Do look up how to use the `Set` class in JS.

PR #847, opened 2025-11-15
West Midlands | ITP-Sept-25 | Mustaf Asani | Sprint 1 | Data groups coursework/sprint 1

asaniDev:
## Learners, PR Template

Self checklist

## Changelist

Added tests to my tasks before creating functions to pass the tests, completed all tasks.

### DIFF
--- a/Sprint-1/implement/sum.js
+++ b/Sprint-1/implement/sum.js
@@ -44,14 +44,10 @@ function sum(elements) {
       return Number(totalSum.toFixed(checkIfFloat));
     }
   } else {
-    const returnValue = elements.reduce(
+    return elements.reduce(
       (accumulator, currentValue) => accumulator + currentValue
     );
-
-    return returnValue;
   }
 }
-
-sum([-34.6, 1.87, 10.67, -7.984, 15.5, -1.1, 8.0]);
 
 module.exports = sum;


Sprint-1/fix/median.js
```diff
@@ -5,10 +5,56 @@
...
+  //because we use array methods first we check we are passing an array
+  let numberArray = [];
+
+  function isNumber(value) {
+    if (typeof value === "number") {
+      return true;
+    }
+
+    return false;
+  }
```
cjyuan: We could simplify 
```
  if (condition)
    return true;
  return false;
```

into

```
  return condition;          // Note: Use !!(condition) if condition is not a Boolean-type value
```

Sprint-1/implement/dedupe.js
```diff
@@ -1 +1,50 @@
...
+        matchCount++;
+
+        if (matchCount > 1) {
+          myArray.splice(i, 1);
+        }
+      }
+    }
+  });
+
+  return myArray;
```
cjyuan: Instead of deleting duplicate items from the given array, consider adding unique items to a new array.

The function is expected to always return a new array.

Sprint-1/implement/max.test.js
```diff
@@ -16,28 +16,108 @@ const findMax = require("./max.js");
...
+    },
+    {
+      input: ["zero", "switch", "price", "tom", "max", "nick", "reach"],
+      expected: "tom",
+    },
+  ].forEach(({ input, expected }) =>
+    it(`returns the largest number ignoring non-numeric values`, () =>
+      expect(findMax(input)).toEqual(expected))
+  );
+});
```
cjyuan: How do you determine "max" value among non-number values?

What return value do you expect from the following function calls?
```
findMax( ["Ten", "one"] );
findMax( ["Ten", null] );
findMax( ["Ten", true] );
findMax( ["2", "1000"] );
findMax( [undefined, [1,2,3,4] ] );
```

Sprint-1/implement/max.js
cjyuan: This function has several bugs. 

You may want to reimplement it following this approach:
- First create a new array to keep only the values you consider "valid".
- Then find the max in the new array. 
    - When dealing with the elements in this new array, you don't have to check for invalid values.

Sprint-1/refactor/includes.js
```diff
@@ -1,12 +1,13 @@
 // Refactor the implementation of includes to use a for...of loop
 
 function includes(list, target) {
-  for (let index = 0; index < list.length; index++) {
-    const element = list[index];
+  for (item of list) {
+    const element = item;
```
cjyuan: - `item` is not delcared.
- Code on lines 4 and 5 could further be simplified.

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,53 @@
 function sum(elements) {
+  //function to check the number of decimal places
+  function numberOfFloat(number) {
+    if (Number.isInteger(number)) {
+      return 0;
+    } else {
+      return number.toString().split(".")[1].length;
+    }
+  }
```
cjyuan: This approach to figure out number of decimal places does not work for all floating point numbers.
- Some floating point numbers can also be represented in scientific format. For examples, `1e-100`, `3.12e50`.

Sprint-1/implement/sum.js
cjyuan: I have to praise you for going through a great length to implement a `sum()` function that takes into consideration the number of decimal places in the input array.

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,87 @@ const sum = require("./sum.js");
...
+  [
+    {
+      input: ["Hi ", "jim ", "what ", "are you ", "doing", "?"],
+      expected: "Hi jim what are you doing?",
+    },
+    {
+      input: ["the ", "quick ", "brown ", "fox ", "jumped"],
+      expected: "the quick brown fox jumped",
+    },
+    { input: [-1000, -500, -450, -320, -2300, -170, -260], expected: -5000 },
```
cjyuan: These values are numbers.

PR #853, opened 2025-11-16
Glasgow | 25-ITP-Sep | Abraham-Habte | Sprint 3| Quote_Generator_App

Abrsh100:
Self checklist

## Changelist

Modification of the html and js file and and writing a new css in style.css file

## Questions

No questions

### DIFF
--- a/Sprint-3/quote-generator/quotes.js
+++ b/Sprint-3/quote-generator/quotes.js
@@ -496,7 +496,7 @@ const quotes = [
 
 // call pickFromArray with the quotes array to check you get a random quote
 let randomQuote = pickFromArray(quotes);
-console.log(randomQuote); // maybe logs { quote: "...", author: "..." }  
+
 
 const quoteElement = document.getElementById("quote");
   const authorElement = document.getElementById("author"); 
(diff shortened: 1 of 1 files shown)


Sprint-3/quote-generator/quotes.js
```diff
@@ -491,3 +495,17 @@ const quotes = [
...
+console.log(randomQuote); // maybe logs { quote: "...", author: "..." }  
+
+const quoteElement = document.getElementById("quote");
+  const authorElement = document.getElementById("author"); 
+  const newquoteButton = document.getElementById("new-quote");
+ 
+  function displayRandomQuote() {
+    const randomQuote = pickFromArray(quotes);
+    quoteElement.innerText = `"${randomQuote.quote}"`;
+    authorElement.innerText = `- ${randomQuote.author}`;
```
tee4tao: Consider using `textContent `instead of `innerText `when setting the quote and author. textContent is generally preferred for performance and consistency because it doesn’t trigger layout calculations and ignores CSS visibility. innerText is useful only when you need to respect rendered styles, which isn’t necessary here since you’re just inserting plain text.

You can read more about innerText and textContent here:
https://www.geeksforgeeks.org/html/difference-between-textcontent-and-innertext/

Sprint-3/quote-generator/quotes.js
```diff
@@ -491,3 +495,17 @@ const quotes = [
...
 
 // call pickFromArray with the quotes array to check you get a random quote
+let randomQuote = pickFromArray(quotes);
+console.log(randomQuote); // maybe logs { quote: "...", author: "..." }  
+
+const quoteElement = document.getElementById("quote");
+  const authorElement = document.getElementById("author"); 
+  const newquoteButton = document.getElementById("new-quote");
+ 
+  function displayRandomQuote() {
```
tee4tao: Having a function where you can fetch a quote and update your DOM is good, as it improves code readability and quality. Well-done!

Abrsh100: got it

PR #855, opened 2025-11-16
Birmingham | 25-ITP-Sept | Khor Biel | Sprint 3 | Quote generator app

wankoak:
## Learners, PR Template

Self checklist

## Changelist

Fixed quote-generator and added styling

No question

### DIFF
--- a/Sprint-3/quote-generator/quotes.js
+++ b/Sprint-3/quote-generator/quotes.js
@@ -498,17 +498,22 @@ function displayRandomQuote() {
   const quoteEl = document.getElementById("quote");
   const authorEl = document.getElementById("author");
 
-  if (quoteEl) quoteEl.textContent = random.quote; // plain text, no extra quotes
-  if (authorEl) authorEl.textContent = random.author; // plain author text
+  if (quoteEl) quoteEl.textContent = random.quote;
+  if (authorEl) authorEl.textContent = random.author;
 }
 
-window.addEventListener("load", function () {
-  // initial display (consumes the first mocked Math.random)
-  displayRandomQuote();
+function displayRandomQuote() {
+  const random = pickFromArray(quotes);
+  const quoteEl = document.getElementById("quote");
+  const authorEl = document.getElementById("author");
 
-  // wire up the button to pick a new quote on each click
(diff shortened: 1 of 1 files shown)


Sprint-3/quote-generator/quotes.js
```diff
@@ -491,3 +491,24 @@ const quotes = [
 ];
 
 // call pickFromArray with the quotes array to check you get a random quote
+
+
+function displayRandomQuote() {
```
tee4tao: Having this function that fetch a quote and update the DOM is a good practice and helps code readability. Well-done!

Sprint-3/quote-generator/quotes.js
```diff
@@ -491,3 +491,24 @@ const quotes = [
...
+  const random = pickFromArray(quotes);
+  const quoteEl = document.getElementById("quote");
+  const authorEl = document.getElementById("author");
+
+  if (quoteEl) quoteEl.textContent = random.quote; // plain text, no extra quotes
+  if (authorEl) authorEl.textContent = random.author; // plain author text
+}
+
+window.addEventListener("load", function () {
```
tee4tao: Can you explain why you chose to use the **load** event, and why you placed the button's click event listener inside it? Do you think this was necessary, or could there have been a better approach?

tee4tao: Using **window.addEventListener("load")** is heavier than needed. **load** waits for all images, fonts,videos, etc. to finish loading. You can read more about that here:
https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event

tee4tao (changes requested):

Sprint-3/quote-generator/quotes.js
tee4tao: And if you still want to stick to your method, you can use DOMContentLoaded, instead of load. So just replace load with DOMContentLoaded.
You can also read this article to understand the difference between load and DOMContentLoaded: https://www.slingacademy.com/article/domcontentloaded-vs-load-understanding-the-difference-in-javascript/

PR #857, opened 2025-11-16
Glasgow | 25-ITP-SEP | Hanna Mykytiuk| Sprint 3| Reading-list

HannaOdud:
## Learners, PR Template

Self checklist

## Changelist

Reading-list

## Questions

No questions.

### DIFF
--- a/Sprint-3/reading-list/index.html
+++ b/Sprint-3/reading-list/index.html
@@ -4,7 +4,7 @@
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <link rel="stylesheet" href="style.css" />
-    <title>Title here</title>
+    <title>Reading list app</title>
   </head>
   <body>
     <div id="content">
--- a/Sprint-3/reading-list/script.js
+++ b/Sprint-3/reading-list/script.js
@@ -1,3 +1,36 @@
+document.addEventListener("DOMContentLoaded", () => {
+  //elements of page to interact with
+   for (let book of books){
+    const title = book.title;
+    const author = book.author;
+    const alreadyRead = book.alreadyRead;
+    const bookCoverImage = book.bookCoverImage;
+    const readingListDom = document.getElementById("reading-list")
+
+    let bookDiv = document.createElement("div");
+    bookDiv.classList.add("book");
+    if (alreadyRead == true){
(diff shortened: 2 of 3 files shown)


Sprint-3/reading-list/script.js
```diff
@@ -1,3 +1,36 @@
+document.addEventListener("DOMContentLoaded", () => {
+  //elements of page to interact with
+   for (let book of books){
+    const title = book.title;
+    const author = book.author;
+    const alreadyRead = book.alreadyRead;
+    const bookCoverImage = book.bookCoverImage;
```
AbogeJr: Good job with this. 
Just a few minor improvements you could make:
1. Seeing as the value of `book` won't change inside our `for` loop, do you think there is a better way to initialize it.
2. While this will work fine, can you find out if there is a cleaner way to unpack object properties into variables.

Sprint-3/reading-list/script.js
```diff
@@ -1,3 +1,36 @@
...
+    const readingListDom = document.getElementById("reading-list")
```
AbogeJr: I don't think we need to initialize the readingListDom object inside this block. Could you look into why that would be problematic?

Sprint-3/reading-list/script.js
```diff
@@ -1,3 +1,36 @@
...
+    bookDiv.classList.add("book");
+    if (alreadyRead == true){
+      bookDiv.classList.add("alreadyRead");
+    }
+    let titleDom = document.createElement("p");
+    titleDom.innerText = title;
+    let authorDom = document.createElement("p");
+    authorDom.innerText = author;
+    let imageDom = document.createElement("img");
+    imageDom.src = bookCoverImage;
```
AbogeJr: Nice. Although `img` tags typically include a certain attribute that you haven't added here. The attribute is meant to improve accessibility and it is also displayed in place of the image in case it does not load. Please look into it.

Sprint-3/reading-list/index.html
```diff
@@ -4,7 +4,7 @@
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <link rel="stylesheet" href="style.css" />
-    <title>Title here</title>
+    <title>Reading list app</title>
```
AbogeJr: 👌

PR #861, opened 2025-11-19
London | 25-ITP-Sep| Sophia Mohamed | Sprint 2 | Sprint 2 Coursework 

saff-coder:
## Learners, PR Template

Self checklist

## Changelist
I have implemented the required functions, and run the required tests 

## Questions
No questions, thank you

### DIFF
--- a/Sprint-2/implement/tally.test.js
+++ b/Sprint-2/implement/tally.test.js
@@ -23,7 +23,7 @@ const tally = require("./tally.js");
 // Given an empty array
 // When passed to tally
 // Then it should return an empty object
-test.todo("tally on an empty array returns an empty object");
+
 test("returns an empty object for an empty array", () => {
  const keys =Object.keys(tally([]));
   expect(keys.length).toBe(0);
@@ -31,7 +31,7 @@ test("returns an empty object for an empty array", () => {
 // Given an array with duplicate items
 // When passed to tally
 // Then it should return counts for each unique item
-test.todo("tally counts each unique item in the array");
+
 test("counts how many times each item appears", () => {
   expect(tally(["a", "b", "a", "c", "b", "a"])).toEqual({
     a: 3,
@@ -43,9 +43,14 @@ test("counts how many times each item appears", () => {
 // Given an invalid input like a string
 // When passed to tally
 // Then it should throw an error
-test.todo("tally throws an error for non-array input");
+
 test("throws an error if input is not an array", () => {
   expect(() => tally("not an array")).toThrow(TypeError);
   expect(() => tally(123)).toThrow(TypeError);
   expect(() => tally({})).toThrow(TypeError);
+});
+test("handles object prototype keys safely", () => {
+  expect(tally(["constructor", "constructor"])).toEqual({
+    constructor: 2,
+  });
 });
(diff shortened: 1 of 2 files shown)


saff-coder:
I have updated all the codes

Sprint-2/implement/contains.test.js
```diff
@@ -16,20 +16,42 @@ as the object doesn't contains a key of 'c'
...
+test("contains on object with non-existent property returns false", () => {
+  expect(contains({ a: 1, b: 2 }, "c")).toBe(false);
+});
+
 
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+test("contains on invalid parameters returns false", () => {
+  expect(contains([], "a")).toBe(false);
```
cjyuan: Which of the parameters on line 48 is "invalid"?

Sprint-2/implement/tally.test.js
```diff
@@ -24,11 +24,28 @@ const tally = require("./tally.js");
...
 // When passed to tally
 // Then it should return counts for each unique item
+test.todo("tally counts each unique item in the array");
+test("counts how many times each item appears", () => {
+  expect(tally(["a", "b", "a", "c", "b", "a"])).toEqual({
+    a: 3,
+    b: 2,
+    c: 1,
+  });
+});
```
cjyuan: Does `tally(["constructor", "constructor"])` return what you expect?

Sprint-2/implement/contains.test.js
saff-coder: You’re right — in that test, neither parameter is actually invalid.
[ ] is still an object in JavaScript, and "a" is a valid key, so the test title didn’t match the behaviour.
I’ve now updated the test to use  invalid parameters (such as strings, numbers, null, and undefined)

Sprint-2/implement/lookup.js
```diff
@@ -1,5 +1,14 @@
-function createLookup() {
-  // implementation here
+function createLookup(pairs) {
+  const lookup = {};//start with an empty object
+
+  for (const pair of pairs) {
+    const country = pair[0];// first element is country code
+    const currency = pair[1];// second element is currency code
```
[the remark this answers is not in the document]
saff-coder: Thanks for the feedback!
I’ve updated the loop to use array destructuring so the code is cleaner and easier to read.

Sprint-2/implement/querystring.js
```diff
@@ -1,16 +1,34 @@
...
-    queryParams[key] = value;
+    const key = decodeURIComponent(rawKey);
+    const value = decodeURIComponent(rawValue);
+
+    result[key] = value;
   }
 
-  return queryParams;
+  return result;
 }
```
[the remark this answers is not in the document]
saff-coder: Thank you for the suggestion!
 I have now explored the built-in URLSearchParams

Sprint-2/implement/tally.test.js
saff-coder: thank you!
tally(["constructor", "constructor"]) didn’t work properly because I was using a normal {} object, so "constructor" clashed with the built-in constructor property.
 I used **Object.create(null)** so "constructor" behaves like normal keys and doesn’t clash with JavaScript’s internal properties.

PR #874, opened 2025-11-22
Sheffield | 25-ITP-Sep | Xiayidan Abuxuukuer | Sprint 2 | Coursework

Shayida999:
## Learners, PR Template

Self checklist

## Questions

none for now. Thanks.

### DIFF
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,3 +1,20 @@
-function contains() {}
+function contains(obj, prop) {
+    if (!obj || typeof obj !== "object" || Array.isArray(obj)){
+
+        return false;
+
+      }
+}
+
+if (obj.hasOwnProperty(prop)){
+
+    return true;
+
+  } 
+  else{
+    
+    return false;
+  }
+
 
 module.exports = contains;
--- a/Sprint-2/implement/lookup.js
+++ b/Sprint-2/implement/lookup.js
@@ -1,5 +1,17 @@
-function createLookup() {
+function createLookup(pairs) {
   // implementation here
+
+  const lookup = {};
+
+  for (let i = 0; i < pairs.length; i++){
+
+    const country = pairs[i][0];
+    const country = pairs[i][1];
+    lookup[country] = currency;
+    
+}
+
+return lookup;
 }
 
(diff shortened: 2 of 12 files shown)


Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,20 @@
-function contains() {}
+function contains(obj, prop) {
```
bp7968h: This fulfils the requirements (except one unwanted check), now think of the below:
 - What will happen if we inverse the `if condition` check.
 - What does `hasOwnProperty` method return
 
 This will help make this more better and less cluttered!

Sprint-2/implement/lookup.js
```diff
@@ -1,5 +1,17 @@
...
+function createLookup(pairs) {
   // implementation here
+
+  const lookup = {};
+
+  for (let i = 0; i < pairs.length; i++){
+
+    const country = pairs[i][0];
+    const country = pairs[i][1];
+    lookup[country] = currency;
```
bp7968h: Does `currency` exist in the global and local scope?

Sprint-2/implement/lookup.js
```diff
@@ -1,5 +1,17 @@
-function createLookup() {
+function createLookup(pairs) {
   // implementation here
+
+  const lookup = {};
+
+  for (let i = 0; i < pairs.length; i++){
+
+    const country = pairs[i][0];
```
bp7968h: There is cleaner way `de-structuring` of accessing the values, if we expect the array to have fixed number of items,

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,28 @@
...
+
+        const item = arr[i];
+
+        if (counts[item]) {
+
+            counts[item]= counts[item] + 1;
+        } 
+        else {
+            counts[item]= 1; 
+        }
```
bp7968h: What if we start at `0` when the key is undefined and then increment by `1` if exists. Will shorten the code for sure.

bp7968h (changes requested):
**Reviewed**

Note: Do not have the access to change the label, have request, change tomorrow.

PR #887, opened 2025-11-26
Glasgow | 25-ITP-SEP | Fares Bakhet | Sprint 3 | Alarm Clock

Fares-Bakhet:
Learners, PR Template

Self checklist

## Changelist

Implement an alarm clock function.

## Questions

No questions.

### DIFF
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -1,11 +1,26 @@
+let intervalId = null;
+
 function setAlarm() {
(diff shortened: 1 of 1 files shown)


cjyuan (commented):
1. Code works fine if a user only clicks the "Set Alarm" button once.
Can you fix the issue?  

Before a new alarm clock is set, it is a good idea to reset the background. 
Can you think of any other "application states" that should also be reset?

2. Some unusual input values can make your app behave abnormally. Can you add code to sanitise them?

PR #901, opened 2025-11-30
West Midlands | 25-ITP-Sep | Baba Yusuf | Sprint 3 | Reading List

Baba05206:
## Learners, PR Template

Self checklist

## Changelist

Updated index.html, style.css, style.js and style.test.js files in order to complete the task
Ran the test to ensure all three tests passed
I was struggling with testing 'background color changes depending on whether book has been read' and so updated the test to expect class instead of background color.

### DIFF
--- a/Sprint-3/reading-list/index.html
+++ b/Sprint-3/reading-list/index.html
@@ -8,7 +8,16 @@
   </head>
   <body>
     <div id="content">
-      <ul id="reading-list"></ul>
+      <h1>My Reading List</h1>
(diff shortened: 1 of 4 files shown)


illicitonion (commented):
Your solution generally looks really good!

It looks like you replaced the tests with some code that isn't tests. You may want to revert that change, and explore the tests - I don't think they will quite all pass with your implementation, but it could be you want to change the tests in this instance, not your code!

PR #912, opened 2025-12-06
Sheffield | 25-ITP-Sep | Jak Rhodes-Smith | Sprint 1 | Data Groups

jakr-s:
## Learners, PR Template

Self checklist

## Changelist

Complete all mandatory tasks

### DIFF
--- a/Sprint-1/refactor/includes.js
+++ b/Sprint-1/refactor/includes.js
@@ -1,9 +1,8 @@
 // Refactor the implementation of includes to use a for...of loop
 
 function includes(list, target) {
-  for (let index = 0; index < list.length; index++) {
-    const element = list[index];
-    if (element === target) {
+  for (let item of list) {
+    if (item === target) {
       return true;
     }
   }


Sprint-1/implement/dedupe.test.js
```diff
@@ -16,12 +16,23 @@ E.g. dedupe([1, 2, 1]) target output: [1, 2]
...
+  expect(dedupe([])).toEqual([]);
+});
 
 // Given an array with no duplicates
 // When passed to the dedupe function
 // Then it should return a copy of the original array
+test("given an array with no duplicates, it returns a copy of the original array", () => {
+  expect(dedupe([1, 2, 3])).toEqual([1, 2, 3]);
+  expect(dedupe(["a", "b", "c"])).toEqual(["a", "b", "c"]);
+});
```
cjyuan: This test is expected to check also whether the returned array is a **copy** of the original array -- 
that is, whether the argument and the returned value of `dedupe(array)` are two different arrays.

The following test only checks whether both arrays contain the same elements;
it cannot check whether the arrays are actually different objects:
```
  expect(array1).isEqual(array2);
```

**Suggestion**: Find out how to test whether two arrays are different arrays in Jest.

cjyuan (commented):
Code is working great.
I only have a few suggestions.

PR #933, opened 2025-12-16
London | 25-ITP-Sept | Gislaine Della Bella | Sprint 2 | Sprint 2 -Final

Della-Bella:
Self checklist

## Changelist

Hi Volunteer,

Please ignore my first commits on this PR. I have tried to clean it up, but for some reason, it is not working, and I will go back to it.

The commits to be revised here are from 'Commits on Dec 16, 2025 - Sprint 2- Lookup.js' "d7ce32bf522dd240741bac5ce9bfc5cc6845e3ae".

Please ignore my first comments on this PR. I have tried to clean but for some reason it is not working, and I will go back to it.

The commits to be revised here are from Commits on Dec 16, 2025 - Sprint 2- Lookup.js
"d7ce32bf522dd240741bac5ce9bfc5cc6845e3ae"

thank you for yout time

### DIFF
--- a/Sprint-2/interpret/invert.js
+++ b/Sprint-2/interpret/invert.js
@@ -6,24 +6,42 @@
 
 // E.g. invert({x : 10, y : 20}), target output: {"10": "x", "20": "y"}
 
+// function invert(obj) {
+//   const invertedObj = {};
+
+//   for (const [key, value] of Object.entries(obj)) {
+//     invertedObj.key = value;
+//   }
+
+//   return invertedObj;
+// }
+
+// a) What is the current return value when invert is called with { a : 1 }
+// 1 
+
+// b) What is the current return value when invert is called with { a: 1, b: 2 }
+///2
+
+// c) What is the target return value when invert is called with {a : 1, b: 2}
+//"1": "a", "2": "b" 
+
+// c) What does Object.entries return? Why is it needed in this program?
+//Object.entries({ a: 1, b: 2 })  // Returns: [["a", 1], ["b", 2]]
+// It converts an object to an array of [key, value] pairs so you can loop through them.
+
+// d) Explain why the current return value is different from the target output
+// The bug is invertedObj.key = value — this literally creates a property named "key" 
+// instead of using the variable key. Also, it should swap: value becomes key, key becomes value.
+// e) Fix the implementation of invert (and write tests to prove it's fixed!)
+
 function invert(obj) {
   const invertedObj = {};
 
   for (const [key, value] of Object.entries(obj)) {
-    invertedObj.key = value;
+    invertedObj[value] = key; // Use [value] as key, key as value
   }
 
   return invertedObj;
 }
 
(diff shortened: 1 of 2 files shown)


Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,11 @@
-function contains() {}
+function contains(someObject, propertyName) {
+  let whatTheRobotFound = someObject[propertyName];
+  if (someObject[propertyName] === undefined) {
```
illicitonion: We can store the value `undefined` in objects. What would you expect to be returned in this case? What does your code do?

```javascript
contains({"name": undefined}, "name");

File: Sprint-2/implement/contains.test.js [hunk 2]
```diff
@@ -20,16 +20,39 @@ as the object doesn't contains a key of 'c'
 // Given an empty object
 // When passed to contains
 // Then it should return false
-test.todo("contains on empty object returns false");
+test("contains on empty object returns false", () => {
+  const inputEmpty = [];
```
illicitonion: This isn't an empty object, it's an array. It may be interesting to test what happens when passed an array, but it's not quite what the test name describes.

Sprint-2/implement/lookup.js
```diff
@@ -1,5 +1,14 @@
-function createLookup() {
-  // implementation here
+function createLookup(countryCurrencyPairs) {
+  const lookup = {};
+
+  for (let i = 0; i < countryCurrencyPairs.length; i++) {
+    const countryCode = countryCurrencyPairs[i][0];
```
illicitonion: This works, but probably isn't the most clear way of iterating across this array.

Can you think how other looping constructs, or destructuring, could help you make this more clear and easy to read?

Sprint-2/interpret/invert.js
```diff
@@ -6,24 +6,42 @@
...
-  return invertedObj;
-}
+//   return invertedObj;
+// }
 
 // a) What is the current return value when invert is called with { a : 1 }
+// 1 
 
 // b) What is the current return value when invert is called with { a: 1, b: 2 }
+///2
```
illicitonion: I don't think the function currently returns numbers. Can you talk through why you think it does?

Sprint-2/interpret/invert.js
```diff
@@ -6,24 +6,42 @@
...
 // c) What is the target return value when invert is called with {a : 1, b: 2}
+//"1": "a", "2": "b" 
 
 // c) What does Object.entries return? Why is it needed in this program?
+//Object.entries({ a: 1, b: 2 })  // Returns: [["a", 1], ["b", 2]]
+// It converts an object to an array of [key, value] pairs so you can loop through them.
 
 // d) Explain why the current return value is different from the target output
-
+// The bug is invertedObj.key = value — this literally creates a property named "key"
```
illicitonion: Good identification of the problem and the fix!

PR #934, opened 2025-12-16
West Midlands | ITP-Sept-25 | Mustaf Asani | Sprint 2 | Data groups coursework

asaniDev:
## Learners, PR Template

Self checklist

## Changelist

Completed the debug, implement and interpret tasks

### DIFF
--- a/Sprint-2/implement/tally.js
+++ b/Sprint-2/implement/tally.js
@@ -1,37 +1,13 @@
 function tally(inputArray) {
-  if (Array.isArray(inputArray)) {
-    if (inputArray.length === 0) {
-      return {};
-    }
-
-    let itemCount = 0;
-    const tallyObject = {};
-    let n = 0;
-
-    while (inputArray.length > 0) {
-      itemCount = 0;
-      const tempArray = [];
-      let i = 0;
-      let currentArrayItem = inputArray[0];
-
-      while (i < inputArray.length) {
-        if (currentArrayItem === inputArray[i]) {
-          itemCount++;
-          tempArray.push(inputArray.splice(i, 1));
-          i--;
-        }
-
-        if (itemCount > 0) {
-          tallyObject[currentArrayItem] = itemCount;
-        }
-
-        i++;
-      }
-    }
-    return tallyObject;
-  } else {
+  if (!Array.isArray(inputArray)) {
     throw new Error("error invalid input passed, please provide an array");
   }
+
+  const tallyObject = {};
+  for (const item of inputArray) {
+    tallyObject[item] = (tallyObject[item] || 0) + 1;
+  }
+  return tallyObject;
 }
 
 module.exports = tally;


Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,32 @@
...
+      for (let key of checkKeys) {
+        if (key === checkProp) {
+          isProperty = true;
+        }
+      }
+
+      if (isProperty === false) {
+        return false;
+      }
+    }
```
cjyuan: Some of the code here are redundant. Can you simplify them?

Sprint-2/implement/lookup.js
```diff
@@ -1,5 +1,20 @@
...
+    }
+
+    for (let item of inputArray) {
+      returnObject[item[0]] = item[1];
+    }
+
+    return returnObject;
+  } else {
+    throw new Error("error incorrect parameter passed please provide an array");
+  }
```
cjyuan: We could also express

```
  function() {
    if (condition) {
       statements;
       ...
    }
    else {
      throw new Error(" ... ");
    }
  }
```

as

```
  function() {
    // Handler error or invalid cases first
    if (!condition) {
      throw new Error(" ... ");
    }
    
    // From this point onward, we only need to deal with valid cases,
    // and we do not need to use "else"
    
    statements;
    ...
  }
```

Sprint-2/interpret/invert.js
```diff
@@ -9,21 +9,45 @@
 function invert(obj) {
   const invertedObj = {};
 
-  for (const [key, value] of Object.entries(obj)) {
-    invertedObj.key = value;
+  if (Object.prototype.toString.call(obj) === "[object Object]") {
```
cjyuan: What's the purpose of this check (line 12)? Why not use the approach on line 15-17 to check if `obj` is an object?

Sprint-2/interpret/invert.js
```diff
@@ -9,21 +9,45 @@
...
+    for (const [key, value] of Object.entries(obj)) {
+      if (
+        value === null ||
+        Array.isArray(value) ||
+        (typeof value === "object" && !Array.isArray(value))
+      ) {
+        throw new Error(
+          "error invalid input entered, expecting an object to have only strings as values"
+        );
```
cjyuan: This check does not reject numbers, true/false, `undefined`. If you want to ensure the value is a string, why not just write `typeof value !== 'string'` (if value is not a string)?

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,36 @@
...
+          tempArray.push(inputArray.splice(i, 1));
+          i--;
+        }
+
+        i++;
+      }
+
+      if (tempArray.length > 0) {
+        tallyObject[tempArray[0]] = tempArray.length;
+      }
```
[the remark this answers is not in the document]
asaniDev: i was not resetting the count so it keeps count of the total of all items

Sprint-2/implement/contains.test.js
```diff
@@ -20,16 +20,25 @@ as the object doesn't contains a key of 'c'
...
+test("returns false when an object with properties is passed to contains with a non-existent property name", () =>
+  expect(contains({ area: "Manchester" }, "town")).toBe(false));
 
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+test("should throw an error when when an invalid parameter like an array is passed to contains", () =>
+  expect(() => contains([1, 4, 5])).toThrow(
+    "error invalid parameter please provide an object"
+  ));
```
cjyuan: Now the second argument is `undefined`. An implementation of the function could have thrown an exception because of that.

Your function implementation is correct, but this test is not ideal. What you should specify in the second argument is a **valid key of the array**,

Sprint-2/implement/querystring.js
```diff
@@ -3,14 +3,38 @@ function parseQueryString(queryString) {
...
     return queryParams;
   }
-  const keyValuePairs = queryString.split("&");
 
-  for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
-    queryParams[key] = value;
+  if (queryString.includes("&")) {
+    const keyValuePairs = queryString.split("&");
+    console.log(keyValuePairs + "our 2 pairs");
```
cjyuan: Can you remove all debugging code in this function to keep the code clean?

Also, the indentation is a bit off. Can you improve the indentation?

Sprint-2/interpret/invert.js
```diff
@@ -9,21 +9,41 @@
...
   return invertedObj;
 }
 
+module.exports = invert;
+
 // a) What is the current return value when invert is called with { a : 1 }
+//it returns a string describing the object.
 
 // b) What is the current return value when invert is called with { a: 1, b: 2 }
+//it aso returns the type of the object.
```
cjyuan: Can you use the JS array literal `{ ... }` to show the value of the returned object (by the original function)?

Sprint-2/implement/contains.test.js
asaniDev: i have now used a value for index 0 that would be found in an array to test for an array being picked up.

PR #959, opened 2026-02-24
London|26-ITP-January|Alexandru Pocovnicu|Sprint 1|Data Groups

alexandru-pocovnicu:
## Learners, PR Template

Self checklist

## Changelist
Applied TDD, wrote and read test suites, then implemented functions to pass those tests.


Sprint-1/fix/median.js
```diff
@@ -6,9 +6,29 @@
...
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
+  if (!Array.isArray(list)) {
+    return null;
+  }
+
+  let listCopy = [...list];
+  let filteredList = listCopy.filter((x) => typeof x === "number");
+  let sortedList = filteredList.sort((a, b) => a - b);
```
cjyuan: Why not call `.filter()` on the original array? Is there a need to create a copy of the original array on line 13?

Sprint-1/implement/dedupe.test.js
```diff
@@ -16,12 +16,29 @@ E.g. dedupe([1, 2, 1]) target output: [1, 2]
...
 
 // Given an array with no duplicates
 // When passed to the dedupe function
 // Then it should return a copy of the original array
+test("given an array with no duplicates, it returns a copy of the original  array", () => {
+  const array = [1,"d",6];
+  const currentOutput = dedupe(array);
+  const targetOutput = [1,"d",6];
+  expect(currentOutput).toEqual(targetOutput);
+});
```

Sprint-1/implement/dedupe.test.js
```diff
@@ -16,12 +16,29 @@ E.g. dedupe([1, 2, 1]) target output: [1, 2]
...
 
 // Given an array with strings or numbers
 // When passed to the dedupe function
 // Then it should remove the duplicate values, preserving the first occurence of each element
+test("given an array with strings or numbers, it removes the duplicate values, preserving the first occurence of each element", () => {
+  const array = [1, "d", 6,1,5,"d",null];
+  const currentOutput = dedupe(array);
+  const targetOutput = [1, "d", 6,5,null];
+  expect(currentOutput).toEqual(targetOutput);
+});
```
cjyuan, line 45: There is a bug in your `dedupe()` function.

If you cannot find the bug, try testing more samples.

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,11 @@
 function findMax(elements) {
+  const filteredElements = elements.filter(
+    (x) => typeof x === "number" && !Number.isNaN(x)
+  );
+
+  const sortedElements = filteredElements.sort((a, b) => a - b);
+
+  return sortedElements.reduce((a, b) => Math.max(a, b), -Infinity);
```
cjyuan: If you are comparing each element in the array one by one, why sort the array on line 6?

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,13 @@
...
+  const filteredElements = elements.filter(
+    (x) => typeof x === "number" && !Number.isNaN(x)
+  );
+
+  if (filteredElements.length === 0) {
+    return 0;
+  }
+  return filteredElements.reduce((a, b) => a + b);
 }
+console.log(sum([0]));
```
cjyuan: Should remove debugging code to keep the code clean.

Sprint-1/stretch/aoc-2018-day1/solution.js
cjyuan, line 1: I think this exercise expects the data to be kept in a separate file, and the JS code should read the data from that file.

Sprint-1/fix/median.js
alexandru-pocovnicu: not really, just a vestige of my earlier tries, and I thought it can't cause any harm, I will update

Sprint-1/implement/max.js
alexandru-pocovnicu: unnecessary

Sprint-1/stretch/aoc-2018-day1/solution.js
alexandru-pocovnicu: i will skip this exercise for now

alexandru-pocovnicu:
thank you

Sprint-1/fix/median.js
```diff
@@ -6,8 +6,25 @@
...
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
+  if (!Array.isArray(list)) {
+    return null;
+  }
+
+  let filteredList = list.filter((x) => typeof x === "number");
+  let sortedList = filteredList.sort((a, b) => a - b);
```
cjyuan: These variables wont' be reassigned a new value. Best practice is to declare them with `const`.

cjyuan (commented):
Changes look good. Well done.

illicitonion:

PR #961, opened 2026-02-25
 Cape Town | 2026-ITP-Jan | Isaac Abodunrin | Sprint 1 | Grouping data: Arrays and loops

bytesandroses:
## Self checklist

## Changelist

I completed the following together with tests:

- Fixed median.js
- Implemented dedupe.js, max.js, and sum.js
- Refactored includes.js

### DIFF
--- a/Sprint-1/implement/sum.js
+++ b/Sprint-1/implement/sum.js
@@ -1,4 +1,11 @@
 function sum(elements) {
+  const filteredElements = elements.filter((element) =>
+    Number.isFinite(element)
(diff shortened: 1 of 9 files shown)


Sprint-1/implement/sum.js
JaypeeLan: This works, but it is slightly inefficient because a new array is being created when you use the filter method. You can do everything without creating a new array. It will be a little difficult, but you can do it!

illicitonion:

PR #982, opened 2026-03-07
West Midlands | 26-Jan-ITP | Fida Ali Zada | Sprint 1 | Data Groups

alizada-dev:
## Learners, PR Template

Self checklist

## Changelist

Completed all the exercises of sprint 1 from the Data Group module.

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -7,27 +7,26 @@
 
 function calculateMedian(list) {
 
-  // first of all, checking if 'list' is an array
-  if (Array.isArray(list)) {
+  // first of all, checking if 'list' isn't an array
+  if (!Array.isArray(list)) return null;
     
-    const filteredArr = list.filter(x => typeof x === 'number');
+    const numbers = list.filter(Number.isFinite);
   
-    if (filteredArr.length < 2) {
-      return null;
-    }
+    if (numbers.length === 0) return null;
+    if (numbers.length === 1) return numbers[0];
+
   
     // using 'spread operator' copies the array without mutating it
-    const sortedList = [...filteredArr].sort((a, b) => a - b);
+    numbers.sort((a, b) => a - b);
   
-    const middleIndex = Math.floor(sortedList.length / 2);
+    const middleIndex = Math.floor(numbers.length / 2);
   
-    if (sortedList.length % 2 === 0) {
-      return (sortedList[middleIndex - 1] + sortedList[middleIndex]) / 2;
+    if (numbers.length % 2 === 0) {
+      return (numbers[middleIndex - 1] + numbers[middleIndex]) / 2;
     } else {
-      return sortedList[middleIndex]
+      return numbers[middleIndex]
     }
-  }
-  return null;
+
 }
 
 module.exports = calculateMedian;
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -6,9 +6,28 @@
 // or 'list' has mixed values (the function is expected to sort only numbers).
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+
+  // first of all, checking if 'list' is an array
+  if (Array.isArray(list)) {
+    
+    const filteredArr = list.filter(x => typeof x === 'number');
+  
+    if (filteredArr.length < 2) {
+      return null;
+    }
+  
+    // using 'spread operator' copies the array without mutating it
+    const sortedList = [...filteredArr].sort((a, b) => a - b);
+  
+    const middleIndex = Math.floor(sortedList.length / 2);
+  
+    if (sortedList.length % 2 === 0) {
+      return (sortedList[middleIndex - 1] + sortedList[middleIndex]) / 2;
+    } else {
+      return sortedList[middleIndex]
+    }
+  }
+  return null;
 }
 
 module.exports = calculateMedian;
--- a/Sprint-1/implement/dedupe.js
+++ b/Sprint-1/implement/dedupe.js
@@ -1,1 +1,13 @@
-function dedupe() {}
+function dedupe(array) {
+    let arr = array.filter((item, index) => {
+
+        // indexOf() returns the first index where that value appears in the array.
+        if (array.indexOf(item) === index) {
+            return item;
+        }
+    })
+
+    return arr;
+}
+
+module.exports = dedupe;
--- a/Sprint-1/implement/dedupe.js
+++ b/Sprint-1/implement/dedupe.js
@@ -1,13 +1,15 @@
 function dedupe(array) {
-    let arr = array.filter((item, index) => {
+    // let arr = array.filter((item, index) => {
 
-        // indexOf() returns the first index where that value appears in the array.
-        if (array.indexOf(item) === index) {
-            return item;
-        }
-    })
+    //     // indexOf() returns the first index where that value appears in the array.
+    //     if (array.indexOf(item) === index) {
+    //         return item;
+    //     }
+    // })
 
-    return arr;
+    // return arr;
+    
+    return [...new Set(array)];
 }
 
 module.exports = dedupe;
--- a/Sprint-1/implement/dedupe.test.js
+++ b/Sprint-1/implement/dedupe.test.js
@@ -16,12 +16,23 @@ E.g. dedupe([1, 2, 1]) target output: [1, 2]
 // Given an empty array
 // When passed to the dedupe function
 // Then it should return an empty array
-test.todo("given an empty array, it returns an empty array");
+
+test("return an empty array for an empty array", () => {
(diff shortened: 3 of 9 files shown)


Sprint-1/fix/median.js
```diff
@@ -6,9 +6,28 @@
...
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+
+  // first of all, checking if 'list' is an array
+  if (Array.isArray(list)) {
+    
+    const filteredArr = list.filter(x => typeof x === 'number');
```
cjyuan: Do you plan to include +/- `Infinity` and `NaN` in the calculation of median?

Sprint-1/fix/median.js
```diff
@@ -6,9 +6,28 @@
...
+  
+    if (filteredArr.length < 2) {
+      return null;
+    }
```
cjyuan: What is the median of an array with one number?

Sprint-1/fix/median.js
```diff
@@ -6,9 +6,28 @@
...
+  
+    // using 'spread operator' copies the array without mutating it
+    const sortedList = [...filteredArr].sort((a, b) => a - b);
```
cjyuan: Why make a copy of `filteredArr` before sorting?

Sprint-1/fix/median.js
```diff
@@ -6,9 +6,28 @@
 // or 'list' has mixed values (the function is expected to sort only numbers).
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+
+  // first of all, checking if 'list' is an array
+  if (Array.isArray(list)) {
```
cjyuan: Could consider return `null` as soon as you discover `list` is NOT an array -- one less `{ ... }` to deal with.

Sprint-1/implement/dedupe.js
```diff
@@ -1 +1,13 @@
-function dedupe() {}
+function dedupe(array) {
+    let arr = array.filter((item, index) => {
+
+        // indexOf() returns the first index where that value appears in the array.
+        if (array.indexOf(item) === index) {
+            return item;
+        }
+    })
```
cjyuan: This work. Could explore using `Set` for better performance.

Sprint-1/implement/dedupe.test.js
```diff
@@ -16,12 +16,23 @@ E.g. dedupe([1, 2, 1]) target output: [1, 2]
...
+})
 
 // Given an array with no duplicates
 // When passed to the dedupe function
 // Then it should return a copy of the original array
 
+test("if no duplicates, return the copy of the original array", () => {
+    expect(dedupe([1, 3, 5, 7])).toEqual([1, 3, 5, 7]);
+})
+
```

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,12 @@
 function findMax(elements) {
+    const filteredArr = elements.filter(x => typeof x === 'number');
```
cjyuan: Do you plan to consider also `NaN`?

The same also applies to the function in `sum.js`.

Sprint-1/implement/max.test.js
```diff
@@ -16,28 +16,58 @@ const findMax = require("./max.js");
...
+})
+
 // Given an array with non-number values
 // When passed to the max function
 // Then it should return the max and ignore non-numeric values
 
+test("ignore the non-numeric values", () => {
+    expect(findMax(["Blue", 3, "White", "Orange", "Pink"])).toEqual(3);
+})
+
```

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,49 @@ const sum = require("./sum.js");
...
+test("return the correct total when passed negative numbers", () => {
+    expect(sum([-3, -6, -1])).toEqual(-10);
+})
+
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
+test("decimal number arrays", () => {
+    expect(sum([0.2, 0.4, 0.2])).toEqual(0.8);
+})
```
cjyuan, line 45: Decimal numbers in most programming languages (including JS) are internally represented in "floating point number" format. Floating point arithmetic is not exact. For example, the result of `46.5678 - 46 === 0.5678` is false because `46.5678 - 46` only yield a value that is very close to `0.5678`. Even changing the order in which the program add/subtract numbers can yield different values.

So the following could happen
```javascript

```

Suggestion: Look up 

Sprint-1/stretch/aoc-2018-day1/solution.js
cjyuan, line 1: This **optional** exercise requires reading numbers from a file named `input.txt`, and then sum up the numbers.

Suggestion: Look up Node.js File system (`fs`) API.

Sprint-1/fix/median.js
alizada-dev: The median is the number itself, sir.

Sprint-1/fix/median.js
alizada-dev: You are right. I used a different filtering method that excludes +/- infinity and NaN, and directly sorted the filtered array.

Sprint-1/implement/dedupe.js
alizada-dev: `function dedupe(array) {
    return [...new Set(array)];
}`

Sprint-1/implement/dedupe.test.js
alizada-dev: By implementing `return [...new Set(array)];`, the test will pass correctly, because this implementation always returns a new array.

Sprint-1/implement/max.js
alizada-dev: `const numbers = elements.filter(Number.isFinite);` I used this method.

alizada-dev: return elements.filter(Number.isFinite).reduce((a, b) => a + b, 0); Also used this implementation for sum function.

Sprint-1/implement/max.test.js
alizada-dev: Got it. `expect(findMax(["Blue", 3, "400", "2", "White", "Orange", "Pink"])).toEqual(3);` This test passes correctly.

Sprint-1/implement/sum.test.js
alizada-dev: You are right! `expect(sum([1.2, 0.6, 0.005])).toBeCloseTo(1.805);` I used `toBeCloseTo`

Sprint-1/implement/dedupe.test.js
cjyuan: The concern here is not about your function implementation -- it was correct.

The issue is, how to prepare a test to check if the function is indeed returning a **copy** of the original array.

With the test you had, a function like this could also pass the test.
```javascript
function  dedupe(array) {
  const set = new Set(array);
  if (set.size == array.length) return array;
  return [...set];
}  
```

Sprint-1/stretch/aoc-2018-day1/solution.js
alizada-dev: Got it, sir.
`const fs = require('fs');`

`const data = fs.readFileSync('input.txt', 'utf-8');`

`const numbers = data.split(/\s+/).filter(Boolean).map(Number);`

`const sum = numbers.reduce((a, b) => a + b, 0);`

Sprint-1/fix/median.js
alizada-dev: `if (!Array.isArray(list)) return null;`

Sprint-1/fix/median.js
alizada-dev: `const numbers = list.filter(Number.isFinite);`

Sprint-1/implement/dedupe.test.js
alizada-dev: `const input = [1, 3, 5, 7];`
`const result = dedupe(input);`

`expect(result).toEqual(input);`
`expect(result).not.toBe(input);`

cjyuan: With this test, there is a chance that, even though `result` has incorrect elements (for example, `[]`), 

cjyuan:
The proposed test in `dedupe.test.js` is not quite correct.

All other changes look good.

Sprint-1/implement/dedupe.test.js
alizada-dev:
```
const input = [1, 3, 5, 7];
const result = dedupe(input);

expect(result).toEqual([1, 3, 5, 7]);
expect(result).not.toBe([1, 3, 5, 7]);
```

Is it alright, this time?

cjyuan: Still not correct. Try looking up the difference between `toBe()` and `toEqual()`.

illicitonion:

PR #983, opened 2026-03-08
London | 26-ITP-January | Eugenie Ahangama | Sprint 2 | Data Groups

Eugenie-A:
## Learners, PR Template

Self checklist

## Changelist

- Fixed multiple JavaScript implementation bugs across several exercises (address, author, recipe)
- Added predictions and explanations for each problem to show understanding
- Implemented and tested a contains function
- Implemented and tested a lookup function
- Fixed parseQueryString function and wrote additional edge case tests to cover various scenarios
- Implemented and tested a tally function 
- Answered questions about the broken invert function to explain why it wasn't working
- Fixed the invert function and added module.exports which was missing
- Wrote test cases for invert to prove the fix works correctly

## Questions

What is the difference between dot notation and bracket notation?

### DIFF
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -5,8 +5,8 @@ function contains(object, propertyName) {
   if (typeof object !== "object") return false;
   if (Array.isArray(object)) return false;
 
-  // Use hasOwnProperty to check only own properties
-  return Object.prototype.hasOwnProperty.call(object, propertyName);
+  // Use Object.hasOwn to check only own properties
+  return Object.hasOwn(object, propertyName);
 }
 
 module.exports = contains;
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,3 +1,12 @@
-function contains() {}
+function contains(object, propertyName) {
+  // return false if object is null/defined or not an object
+  if (object == null) return false;
+  // typeof check + reject arrays
+  if (typeof object !== "object") return false;
+  if (Array.isArray(object)) return false;
+
+  // Use hasOwnProperty to check only own properties
+  return Object.prototype.hasOwnProperty.call(object, propertyName);
+}
 
 module.exports = contains;
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -16,20 +16,40 @@ as the object doesn't contains a key of 'c'
 // Given a contains function
 // When passed an object and a property name
 // Then it should return true if the object contains the property, false otherwise
+test("returns true when the property exists in the object", () => {
+  expect(contains({ a: 1, b: 2 }, "a")).toBe(true);
+  expect(contains({ a: 1, b: 2 }, "b")).toBe(true);
+});
 
 // Given an empty object
 // When passed to contains
 // Then it should return false
-test.todo("contains on empty object returns false");
+test("contains on empty object returns false", () => {
+  expect(contains({}, "a")).toBe(false);
+  expect(contains({}, "toString")).toBe(false);
+});
 
 // Given an object with properties
 // When passed to contains with an existing property name
 // Then it should return true
+test("returns true for existing property name", () => {
+  expect(contains({ name: "Alex" }, "name")).toBe(true);
+  expect(contains({ id: 42 }, "id")).toBe(true);
+});
 
 // Given an object with properties
 // When passed to contains with a non-existent property name
 // Then it should return false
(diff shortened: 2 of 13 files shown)


Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,12 @@
-function contains() {}
+function contains(object, propertyName) {
+  // return false if object is null/defined or not an object
+  if (object == null) return false;
+  // typeof check + reject arrays
+  if (typeof object !== "object") return false;
+  if (Array.isArray(object)) return false;
+
+  // Use hasOwnProperty to check only own properties
+  return Object.prototype.hasOwnProperty.call(object, propertyName);
```
cjyuan: Could also use `Object.hasOwn(obj, propertyName)`

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,22 @@
...
+
+  // Create an empty obkect to store our counts
+  const counts = {};
+
+  // Loop through each item in the array one by one
+  for (const item of arr) {
+    // If the item already exists in counts, add 1 to it
+    // If it doesn't exist yet (undefined), start it at 0 then add 1
+    counts[item] = (counts[item] || 0) + 1;
+  }
```
```
```

Sprint-2/interpret/invert.test.js
```diff
@@ -0,0 +1,34 @@
...
+test("invert on an empty object returns an empty object", () => {
+  expect(invert({})).toEqual({});
+});
+
+// Given an invalid input like an array
+// When passed to invert
+// Then it should throw an error
+test("invert throws an error for invalid input", () => {
+  expect(() => invert([])).toThrow("Input must be an object");
+});
```
cjyuan, line 38: Could also test a few more values of non-object types.

Sprint-2/implement/querystring.js
```diff
@@ -1,13 +1,29 @@
...
+  for (const segment of segments) {
+    // Split only on the first =, but everything after is a value
+    const eqIndex = segment.indexOf("=");
+
+    const key = eqIndex === -1 ? segment : segment.slice(0, eqIndex);
+    const value = eqIndex === -1 ? "" : segment.slice(eqIndex + 1);
+
+    if (key) {
+      queryParams[key] = value;
+    }
```
cjyuan, line 27: Note: 

Sprint-2/implement/contains.test.js
```diff
@@ -16,20 +16,40 @@ as the object doesn't contains a key of 'c'
...
+  expect(contains({ age: 54 }, "name")).toBe(false);
+});
 
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+test("handles invalid parameters like an array", () => {
+  expect(contains([], "length")).toBe(false);
+  expect(contains(123, "length")).toBe(false);
+  expect(contains("string", "length")).toBe(false);
```
cjyuan, line 54: It is a good choice to use "length" in these tests.

cjyuan (commented):
Code well written.

I only have a few suggestions.

Sprint-2/implement/contains.js
Eugenie-A: Thanks for the suggestion, I've updated it to use Object.hasOwn().

Sprint-2/implement/tally.js
Eugenie-A: I tested `tally(["toString", "toString"])` and saw the issue. With a regular {} object, `toString` already exists as an inherited property so it returned the wrong value. 

I've fixed it by using `Object.create(null)` which creates an object with no inherited properties, so it correctly returns `{ toString: 2 }`.

Sprint-2/interpret/invert.test.js
Eugenie-A: Thank you for the suggestion. I've added more invalid input types to the existing test, including strings, numbers, null and undefined.

Sprint-2/implement/querystring.js
Eugenie-A: Thank you for the suggestion. I've added decodeURIComponent to decode percent-encoded keys and values.

Sprint-2/implement/contains.test.js
Eugenie-A: Thank you, I chose "length" because it's a built-in property on arrays and strings, so it helps verify that the function correctly returns false for non-plain-objects rather than accidentally detecting inherited properties.

cjyuan:
Changes are all good, and thanks for the clear explanations. Excellent job.

illicitonion:

PR #985, opened 2026-03-09
West Midlands | 26-Jan-ITP | Fida Ali Zada | Sprint 2 | Data Group

alizada-dev:
Self checklist

## Changelist

Completed the exercises in the Sprint-2 folder

### DIFF
--- a/Sprint-2/debug/author.js
+++ b/Sprint-2/debug/author.js
@@ -13,6 +13,6 @@ const author = {
   alive: true,
 };
 
-for (const [key, value] of Object.entries(author)) {
+for (const value of Object.values(author)) {
   console.log(value);
 }
--- a/Sprint-2/debug/author.js
+++ b/Sprint-2/debug/author.js
@@ -1,4 +1,6 @@
 // Predict and explain first...
+// As object is key-value pairs of data, it isn't exactly iterable like arrays.
+// We can loop over the keys or values of an object using Object.entries()
 
 // This program attempts to log out all the property values in the object.
 // But it isn't working. Explain why first and then fix the problem
@@ -11,6 +13,6 @@ const author = {
   alive: true,
 };
 
-for (const value of author) {
+for (const [key, value] of Object.entries(author)) {
   console.log(value);
 }
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,6 +1,6 @@
 function contains(obj, x) {
     for (const key in obj) {
-        if (Object.hasOwn(obj, key) && key == x) {
+        if (key == x) {
             return true
         }
     }
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,10 +1,22 @@
+// function contains(obj, x) {
+//     for (const key in obj) {
+//         if (key == x) {
+//             return true
+//         }
+//     }
+//     return false
+// }
+
 function contains(obj, x) {
-    for (const key in obj) {
-        if (key == x) {
-            return true
-        }
+    if (
+        typeof obj !== "object" ||
+        obj === null ||
+        Array.isArray(obj)
+    ) {
+        return false;
     }
-    return false
+
+    return Object.prototype.hasOwnProperty.call(obj, x);
 }
 
 module.exports = contains;
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,3 +1,13 @@
-function contains() {}
+function contains(obj, x) {
+    for (const key in obj) {
+        if (Object.hasOwn(obj, key) && key == x) {
+            return true
+        }
+    }
+    return false
+}
 
 module.exports = contains;
+
+// Object.hasOwn(obj, key)
+// This ensures the property belongs directly to the object, not its prototype.
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -55,3 +55,10 @@ test("contains on invalid parameters returns false", () => {
     expect(contains("I am a string")).toStrictEqual(false);
     expect(contains(777)).toStrictEqual(false);
 })
+
+test("contains() handles arrays, strings, null, and undefined safely", () => {
+    expect(contains([["a", "b", "c"], "1"])).toStrictEqual(false);
+    expect(contains(["abc", "1"])).toStrictEqual(false);
+    expect(contains(null, "1")).toStrictEqual(false);
+    expect(contains(undefined, "1")).toStrictEqual(false);
+})
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -20,16 +20,38 @@ as the object doesn't contains a key of 'c'
 // Given an empty object
 // When passed to contains
 // Then it should return false
-test.todo("contains on empty object returns false");
+
+test("contains on empty object returns false", () => {
+    expect(contains({})).toStrictEqual(false)
+});
+
 
 // Given an object with properties
 // When passed to contains with an existing property name
 // Then it should return true
 
+test("contains on object with an existing property name returns true", () => {
+    expect(contains({a: 1, b: 2, c: 3}, 'c')).toStrictEqual(true);
+    expect(contains({name: "John Doe", job: "developer"}, 'job')).toStrictEqual(true)
+    expect(contains({color: "blue", size: "xxl", madeOf: "China"}, 'madeOf')).toStrictEqual(true);
+})
+
 // Given an object with properties
 // When passed to contains with a non-existent property name
 // Then it should return false
 
+test("contains on object with properties with non-existent property name return false", () => {
+    expect(contains({a: 1, b: 2, c: 3}, 'g')).toStrictEqual(false);
+    expect(contains({a: 1, b: 2, c: 3}, 'abc')).toStrictEqual(false);
+    expect(contains({name: "John Doe", age: 31}, 'job')).toStrictEqual(false);
+})
+
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+
+test("contains on invalid parameters returns false", () => {
+    expect(contains(["a", "b", "c"])).toStrictEqual(false);
+    expect(contains("I am a string")).toStrictEqual(false);
+    expect(contains(777)).toStrictEqual(false);
+})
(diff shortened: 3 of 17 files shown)


Sprint-2/debug/author.js
```diff
@@ -11,6 +13,6 @@ const author = {
   alive: true,
 };
 
-for (const value of author) {
+for (const [key, value] of Object.entries(author)) {
```
cjyuan: Since this loop does not need `key`, you could consider using the method that returns only an array of property values.

Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,13 @@
-function contains() {}
+function contains(obj, x) {
+    for (const key in obj) {
+        if (Object.hasOwn(obj, key) && key == x) {
```
cjyuan: Why check if `key` is in the same object the loop is iterating though?

Sprint-2/implement/contains.test.js
```diff
@@ -20,16 +20,38 @@ as the object doesn't contains a key of 'c'
...
+    expect(contains({a: 1, b: 2, c: 3}, 'abc')).toStrictEqual(false);
+    expect(contains({name: "John Doe", age: 31}, 'job')).toStrictEqual(false);
+})
+
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+
+test("contains on invalid parameters returns false", () => {
+    expect(contains(["a", "b", "c"])).toStrictEqual(false);
```
cjyuan, line 54: Does your function return the value you expect from the following function calls?

`contains(["a", "b", "c"], "1")`
`contains("abc", "1")`
`contains(null, "1")`
`contains(undefined, "1")`

Sprint-2/implement/querystring.js
```diff
@@ -1,12 +1,23 @@
...
+  // 2. with substring method
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
+    const index = pair.indexOf("=");
+
+    const key = pair.substring(0, index);
+    const value = pair.substring(index + 1);
+
     queryParams[key] = value;
   }
```
cjyuan, line 33: For each of the following function calls, does your function return the value you expect?
```
parseQueryString("a=b&=&c=d")
parseQueryString("a=")
parseQueryString("=b")
parseQueryString("a=b&&c=d")
```

---
Note: 

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,26 @@
...
+                countObj[item] = countObj[item] + 1;
+            } else {
+                countObj[item] = 1;
+            }
+        }
+        
+        return countObj;
+    }
+
+    else throw new Error("");
```
cjyuan: Can you give a more descriptive error message?

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,26 @@
...
+        let countObj = {};
+    
+        for (const item of arr) {
+    
+            if (countObj[item]) {
+                countObj[item] = countObj[item] + 1;
+            } else {
+                countObj[item] = 1;
+            }
+        }
```
```
```

Sprint-2/interpret/invert.js
```diff
@@ -10,20 +10,29 @@ function invert(obj) {
...
 }
 
 // a) What is the current return value when invert is called with { a : 1 }
+      // { key: 1 }
 
 // b) What is the current return value when invert is called with { a: 1, b: 2 }
+      // { key: 2 }
 
 // c) What is the target return value when invert is called with {a : 1, b: 2}
+      // { key: 2 }
```
cjyuan: The "target value" refers to the expected value when the function is working correctly.

Sprint-2/stretch/count-words.js
```diff
@@ -26,3 +26,26 @@
...
+  }
+
+  // `Object.entries()` converts object to array
+  const sortedArray = Object.entries(countObj).sort((a, b) => b[1] - a[1]);
+
+  // `Object.fromEntries()` coverts the sorted array to object
+  return Object.fromEntries(sortedArray);
+}
+
+console.log(countWords("you? and! me, and you. me me me me hello Me"));
```
cjyuan, line 50: Does your function return what you expect in the following function calls?

```
```

Sprint-2/stretch/till.js
```diff
@@ -22,10 +22,15 @@ const till = {
 };
 const totalAmount = totalTill(till);
 
+module.exports = totalTill;
+
 // a) What is the target output when totalTill is called with the till object
+      // £NaN
```
cjyuan: I think the question is asking for the "expected output", not the actual output.

Sprint-2/debug/author.js
alizada-dev: `for (const value of Object.values(author)) {
  console.log(value);
}`

Sprint-2/implement/contains.js
alizada-dev: `for (const key in obj) {
        if (key == x) {
            return true
        }
    }`

Sprint-2/implement/contains.test.js
alizada-dev: Now, it handles correctly.

Sprint-2/implement/tally.js
alizada-dev: `else throw new Error("The input should be an array!");`

Sprint-2/implement/tally.js
alizada-dev: `function tally(arr) {
    if (!Array.isArray(arr)) {
        throw new Error("The input should be an array!");
    }
        
    const countObj = Object.create(null);

    for (const item of arr) {

        if (countObj[item]) {
            countObj[item] = countObj[item] + 1;
        } else {
            countObj[item] = 1;
        }
    }
    
    return countObj;
}`

Sprint-2/interpret/invert.js
alizada-dev: `{1 : 'a', 2: 'b'}`

alizada-dev:
I am sorry, sir, for responding late. I got a job because it was so stressful not working.

Sprint-2/implement/querystring.js
alizada-dev: `function parseQueryString(queryString) {
  const queryParams = {};

  if (!queryString) return queryParams;

    // skip empty pairs (e.g. &&)
    if (pair === "") continue;

    const index = pair.indexOf("=");

    let key, value;

    if (key === -1) {
      key = pair;
      value = "";
    } else {
      key = pair.substring(0, index);
      value = pair.substring(index + 1);
    }

    // decoding URL
    try {
      key = decodeURIComponent(key);
      value = decodeURIComponent(value);
    } catch (error) {
      
    }

    queryParams[key] = value;
  }

  return queryParams;
}`

cjyuan:
The proposed changes look good. Have you applied the changes and pushed the changes to GitHub? I don't see any new commits on this PR branch.

alizada-dev:
I had forgotten to push the changes. I just pushed them.

cjyuan:
Still nothing new on this branch. Did you push the changes on the right branch?

cjyuan:
Those commits were done on your local computer. I cannot see them unless you sync/push them to GitHub.

It is a good practice to double check what you have committed to this PR branch on GitHub by clicking the "Files changed" tab to review them.

image

alizada-dev:
Dear Sir, I am confused. I have committed and pushed all the changes to this PR branch on GitHub. If you check my latest commits, they are shown in the "Files changed" tab as well.

Sprint-2/implement/contains.test.js
cjyuan: This test cannot not yet confirm that the function correctly returns false when the first argument is an array. 
This is because `contains(["a", "b", "c"])` is the executed as `contains("[a", "b", "c"], undefined)`, which could also return `false` simply because `"undefined"` is not a key of the array.

**Arrays are objects, with their indices acting as keys**. A proper test should use a **valid
key** to ensure the function returns `false` specifically because the input is an array, not because the key is missing.

Currently your function will return `true` in these two function calls.
- `contains(["a", "b", "c"], "1")`
- `contains("abc", "1")`

cjyuan:
Sorry. I noticed no changes were made to `till.js` and `count-words.js`, and thought you didn't push your commits to GitHub.

I have now marked the addressed comments as "Resolved conversation". 

Can you addressed the remaining three unresolved conversation?

Sprint-2/implement/contains.test.js
alizada-dev: Now, I understand what you are telling, sir. I changed the function first:
`function contains(obj, x) {
    if (
        typeof obj !== "object" ||
        obj === null ||
        Array.isArray(obj)
    ) {
        return false;
    }

    return Object.prototype.hasOwnProperty.call(obj, x);
}

module.exports = contains;`

And the test:
`test("contains() handles arrays, strings, null, and undefined safely", () => {
    expect(contains([["a", "b", "c"], "1"])).toBe(false);
    expect(contains(["abc", "1"])).toBe(false);
})`

Please check if it is okay; otherwise, I will change.

Sprint-2/stretch/count-words.js
alizada-dev: Thank you for pointing out the bugs. Now, this should work for all cases.

`function countWords(str) {
  const words = str
    .toLowerCase()
    .replace(/[^a-zA-Z0-9\s]/g, '')   // replace punctuation with space
    .trim()
    .split(/\s+/);    // split on any white space

  const countObj = Object.create(null);

  for (const word of words) {
    countObj[word] = (countObj[word] || 0) + 1;
  }

  // `Object.entries()` converts object to array
  const sortedArray = Object.entries(countObj).sort((a, b) => b[1] - a[1]);

  // `Object.fromEntries()` coverts the sorted array to object
  return Object.fromEntries(sortedArray);
}`

Sprint-2/stretch/till.js
alizada-dev: // a) What is the target output when totalTill is called with the till object
      // £4.40

alizada-dev:
It is all right, sir. Thank you for your time and comments. I brought some changes. If the changes look good, I will push them to GitHub.

cjyuan:
The proposed changes look good. I will mark this PR as completed.

Note: Sharing code as a "code block" in PR comment can make the code easier to read.

illicitonion:

PR #986, opened 2026-03-09
Sheffield | 26-Jan-ITP | Daniel Aderibigbe | Sprint 1 | Coursework/sprint 1

Dan2Clouted:
## Learners, PR Template

Self checklist

## Changelist

Completed all exercises.

### DIFF
--- a/Sprint-1/implement/dedupe.js
+++ b/Sprint-1/implement/dedupe.js
@@ -6,8 +6,6 @@ function dedupe(array) {
       result.push(item);
     }
   }
-
-  return result;
 }
 
 module.exports = dedupe;
--- a/Sprint-1/implement/dedupe.js
+++ b/Sprint-1/implement/dedupe.js
@@ -6,6 +6,7 @@ function dedupe(array) {
       result.push(item);
     }
   }
+  return result;
 }
 
 module.exports = dedupe;
--- a/Sprint-1/implement/dedupe.js
+++ b/Sprint-1/implement/dedupe.js
@@ -6,6 +6,8 @@ function dedupe(array) {
       result.push(item);
     }
   }
+
+  return result;
 }
 
 module.exports = dedupe;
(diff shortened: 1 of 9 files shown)


Sprint-1/implement/dedupe.js
```diff
@@ -1 +1,12 @@
-function dedupe() {}
+function dedupe(array) {
+  const result = [];
+
+  for (let i = 0; i < array.length; i++)
+    if (!result.includes(array[i])) {
+      result.push(array[i]);
+    }
+
```
cjyuan: - Better practice is to always use `{ ... }` even if the for loop is controlling only one statement.

- When the code in a for-loop does not need array index, a for-of loop is is usually more readable.

Dan2Clouted:
Thanks for the detailed feedback — really helpful.

I’ve gone through all the comments and made the following updates:

* Refactored loops (e.g. using `for...of` where appropriate and adding braces for clarity)
* Improved tests to ensure correctness, including checking that returned arrays are new copies rather than the original reference
* Handled edge cases in functions like `max` and `sum`, including ignoring `NaN` and non-finite values
* Updated decimal tests to use `toBeCloseTo` to avoid floating point precision issues
* Removed duplicate tests and improved test coverage where needed

cjyuan (commented):
Changes look good, but you introduced a bug into one of the files. Can you fix it?

Dan2Clouted:
The bug issues has been resolved.

cjyuan:
I can still see the bug. If you run jest test, you should notice some failed tests.

Dan2Clouted:
I think now it should have been fixed.

cjyuan:
No new commits pushed to GitHub, and label not yet changed.

Dan2Clouted:
Sorry for all this, But i think it should be fixed now. Thank you for your time.

cjyuan:
One of the commit fixed the bug but it wasn't pushed to GitHub before. However, a newer commit reintroduced the bug.

So the bug remains.

Dan2Clouted:
It should be fixed now. Thank you for your time and patience.

illicitonion:

PR #987, opened 2026-03-10
London | 26-ITP-January | Mouawia Elkhalifa | Sprint 1 | Module Data Groups

MouawiaElkhalifa:
Self checklist

## Changelist

Completed Sprint 1 exercises:

- Implemented sum function
- Implemented max function
- Implemented dedupe function
- Fixed median function
- Refactored includes function

All tests are passing (45/45).

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -6,9 +6,18 @@
 // or 'list' has mixed values (the function is expected to sort only numbers).
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  if(!Array.isArray(list)){return null;}
+  const number  = list.filter(n => typeof n === "number");
+  if(number.length  === 0) {return null;}
+  number.sort((a,b) => a-b)
+  const middleIndex = Math.floor(number.length / 2);
+
+
+  if(number.length%2===0){return (number[middleIndex-1] + number[middleIndex])/2;
+  }
+  else return number[middleIndex];
 }
 
-module.exports = calculateMedian;
+
+
+module.exports = calculateMedian
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -6,18 +6,26 @@
 // or 'list' has mixed values (the function is expected to sort only numbers).
 
 function calculateMedian(list) {
-  if(!Array.isArray(list)){return null;}
-  const number  = list.filter(n => typeof n === "number");
-  if(number.length  === 0) {return null;}
-  number.sort((a,b) => a-b)
-  const middleIndex = Math.floor(number.length / 2);
-
-
-  if(number.length%2===0){return (number[middleIndex-1] + number[middleIndex])/2;
+  if (!Array.isArray(list)) {
+    return null;
   }
-  else return number[middleIndex];
+
+  const numbers = list.filter(
+    (item) => typeof item === "number" && Number.isFinite(item)
+  );
+
+  if (numbers.length === 0) {
+    return null;
+  }
+
+  const sortedNumbers = [...numbers].sort((a, b) => a - b);
+  const middleIndex = Math.floor(sortedNumbers.length / 2);
+
+  if (sortedNumbers.length % 2 === 0) {
+    return (sortedNumbers[middleIndex - 1] + sortedNumbers[middleIndex]) / 2;
+  }
+
+  return sortedNumbers[middleIndex];
 }
 
-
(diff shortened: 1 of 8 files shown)


Sprint-1/fix/median.js
```diff
@@ -6,9 +6,18 @@
 // or 'list' has mixed values (the function is expected to sort only numbers).
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  if(!Array.isArray(list)){return null;}
+  const number  = list.filter(n => typeof n === "number");
```
cjyuan: - Do you plan to consider also `-Infinity`, `Infinity`, and `NaN` in the median calculation (and in the functions in `implement/max.js` and `implement/sum.js`)?

- Naming convention for variables storing an array is: 
  - Append a suffix `Array` or `List` to the name (e.g., `numberArray`)

Sprint-1/fix/median.js
```diff
@@ -6,9 +6,18 @@
...
+  if(number.length  === 0) {return null;}
+  number.sort((a,b) => a-b)
+  const middleIndex = Math.floor(number.length / 2);
+
+
+  if(number.length%2===0){return (number[middleIndex-1] + number[middleIndex])/2;
+  }
+  else return number[middleIndex];
```
cjyuan: Indentation is off and some of the code is not consistently formatted.

?

Sprint-1/implement/dedupe.test.js
```diff
@@ -16,12 +16,28 @@ E.g. dedupe([1, 2, 1]) target output: [1, 2]
...
 // Given an array with no duplicates
 // When passed to the dedupe function
 // Then it should return a copy of the original array
-
+test("given an array with no duplication, it copy of the original array ", () => {
+  const array = [1, 2, 3];
+  const currentOutput = dedupe(array);
+  const targetOutput = [1, 2, 3];
+  expect(currentOutput).toEqual(targetOutput);
+});
```

Sprint-1/implement/max.test.js
```diff
@@ -16,28 +16,65 @@ const findMax = require("./max.js");
...
+});
 // Given an array with only non-number values
 // When passed to the max function
 // Then it should return the least surprising value given how it behaves for all other inputs
+test("given an array with only non-number values returns -Infinity", () => {
+  const maxElement = ["history", "nation"];
+  const currentOutput = findMax(maxElement);
+  const targetOutput = -Infinity;
+  expect(currentOutput).toEqual(targetOutput);
+});
```
cjyuan, line 89: When a string representing a valid numeric literal (for example, `"300"`) is compared to a number, 

To test if the function can correctly ignore non-numeric values, consider including a string such as `"300"` in the relevant test cases.

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,59 @@ const sum = require("./sum.js");
...
 
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
+test("given an array with decimal/float numbers [1.5, -2.4, 4.5], returns the correct total sum", () => {
+  const list = [1.5, -2.4, 4.5];
+  const currentOutput = sum(list);
+  const targetOutput = 3.6;
+  expect(currentOutput).toEqual(targetOutput);
+});
```

So the following could happen
```javascript

```

Suggestion: Look up 

Sprint-1/fix/median.js
MouawiaElkhalifa: Thanks, I improved the implementation to handle non-numeric values correctly.

Sprint-1/fix/median.js
MouawiaElkhalifa: Thanks! I fixed the indentation

Sprint-1/implement/dedupe.test.js
MouawiaElkhalifa: Thanks! I added an additional check using not.toBe to ensure the function returns a new array instead of the original reference.

Sprint-1/implement/max.test.js
MouawiaElkhalifa: Thanks! I added a new test case to ensure numeric strings like "300" are ignored and only real numbers are considered.

Sprint-1/implement/sum.test.js
MouawiaElkhalifa: Thanks! I updated the test to use toBeCloseTo to correctly handle floating point precision.

cjyuan:
Changes look good. Well done.

illicitonion:

PR #993, opened 2026-03-11
London | 26-ITP-Jan | Damian Dunkley | Sprint 2 | Data-Groups

DamianDL:
## Learners, PR Template

Self checklist

## Changelist

All Sprint 2 exercises updated and code changed
## Questions

Please review. If Invert is not correct- changing the function to make it work and putting in naive consol.log tests, please provide some details of what is required? Thanks

### DIFF
--- a/Sprint-2/debug/recipe.js
+++ b/Sprint-2/debug/recipe.js
@@ -10,6 +10,8 @@ const recipe = {
   ingredients: ["olive oil", "tomatoes", "salt", "pepper"],
 };
 
-console.log(`${recipe.title} serves ${recipe.serves}
-  ingredients:
-${recipe}`);
+console.log(`${recipe.title} serves ${recipe.serves}`);
+for (const ingredient of recipe.ingredients) {
+  console.log(ingredient);
+}
+// The problem is that the code is trying to log out the ingredients using a for...of loop, but it is not using the correct syntax to access the ingredients. The recipe.ingredients is an array, so we need to use a for...of loop to iterate over that array and log each ingredient.
--- a/Sprint-2/debug/recipe.js
+++ b/Sprint-2/debug/recipe.js
@@ -9,9 +9,14 @@ const recipe = {
   serves: 2,
   ingredients: ["olive oil", "tomatoes", "salt", "pepper"],
 };
-
+/* Replacing the for loop with Array.prototype.join() to log the ingredients in a single line, separated by new lines. 
 console.log(`${recipe.title} serves ${recipe.serves}`);
 for (const ingredient of recipe.ingredients) {
   console.log(ingredient);
 }
+*/
+console.log(`${recipe.title} serves ${recipe.serves}`);
+console.log (recipe.ingredients.join("\n"));
+
+
 // The problem is that the code is trying to log out the ingredients using a for...of loop, but it is not using the correct syntax to access the ingredients. The recipe.ingredients is an array, so we need to use a for...of loop to iterate over that array and log each ingredient.
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -3,7 +3,7 @@ function contains(obj, prop) {
    !!obj && // confirm that obj is not null or undefined. If undefined or null, return false.
    typeof obj === "object" && // confirm that obj is an object. If not an object, return false.
    !Array.isArray(obj) && // confirm that obj is not an array. If an array, return false.
-   Object.prototype.hasOwnProperty.call(obj, prop)// check if the object has the property. If it does, return true; otherwise, return false.
+   Object.hasOwnProperty.call(obj, prop)// check if the object has the property. If it does, return true; otherwise, return false.
  );
 }
 
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,9 +1,9 @@
 function contains(obj, prop) {
  return (
-   !!obj &&
-   typeof obj === "object" &&
-   !Array.isArray(obj) &&
-   Object.prototype.hasOwnProperty.call(obj, prop)
+   !!obj && // confirm that obj is not null or undefined. If undefined or null, return false.
+   typeof obj === "object" && // confirm that obj is an object. If not an object, return false.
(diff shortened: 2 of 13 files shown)


Sprint-2/debug/recipe.js
```diff
@@ -10,6 +10,8 @@ const recipe = {
   ingredients: ["olive oil", "tomatoes", "salt", "pepper"],
 };
 
-console.log(`${recipe.title} serves ${recipe.serves}
-  ingredients:
-${recipe}`);
+console.log(`${recipe.title} serves ${recipe.serves}`);
+for (const ingredient of recipe.ingredients) {
+  console.log(ingredient);
+}
```
cjyuan, line 16: Could also explore using `Array.prototype.join()`.

Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,10 @@
-function contains() {}
+function contains(obj, prop) {
+ return (
+   !!obj && // confirm that obj is not null or undefined. If undefined or null, return false.
+   typeof obj === "object" && // confirm that obj is an object. If not an object, return false.
+   !Array.isArray(obj) && // confirm that obj is not an array. If an array, return false.
+   Object.prototype.hasOwnProperty.call(obj, prop)// check if the object has the property. If it does, return true; otherwise, return false.
```
cjyuan: Could also use `Object.hasOwn()`.

Sprint-2/implement/contains.test.js
```diff
@@ -16,20 +16,48 @@ as the object doesn't contains a key of 'c'
...
+  expect(contains(obj3, "age")).toBe(true);
+  expect(contains(obj3, "gender")).toBe(false);
+});
 
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+test("contains returns false for invalid input types", () => {
+  const arr = [1, 2, 3];
+  expect(contains(arr, "0")).toBe(false); // Arrays do not have properties like objects, so it should return false
```
cjyuan, line 60: Note: An array is a kind of objects and its indexes are its keys (properties).

So for an implementation that does not check if the given value is an array using `Array.isArray()`, `contains([1, 2, 3], "0")` could have returned `true`.

Sprint-2/implement/lookup.js
```diff
@@ -1,5 +1,23 @@
...
+    if (!Array.isArray(pair) || pair.length !== 2) {// check if the current pair is an array of length 2. If not, throw an error.
+      throw new Error("Each pair must be an array of two elements");
+    } 
+
+    const key = pair[0];// extract the first element of the pair as the key
+    const value = pair[1];// extract the second element of the pair as the value
+    lookup[key] = value;// add the key-value pair to the lookup object
+  }
+  return lookup;
+};
```
cjyuan: Indentation is off.

as recommended in 
?

Sprint-2/implement/querystring.js
```diff
@@ -1,13 +1,15 @@
...
+  const keyValuePairs = queryString.split("&");// Split the query string into an array of key-value pairs using '&' as the delimiter
 
-  for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
-    queryParams[key] = value;
+  for (const pair of keyValuePairs) {// Iterate through each key-value pair in the array
+    const separatorIndex = pair.indexOf("=");// Find the index of the '=' character in the current pair
+    const key = separatorIndex === -1 ? pair : pair.slice(0, separatorIndex);// Extract the key from the pair. If there is no '=', the entire pair is the key; otherwise, it's the substring before '='
+    const value = separatorIndex === -1 ? "" : pair.slice(separatorIndex + 1);// Extract the value from the pair. If there is no '=', the value is an empty string; otherwise, it's the substring after '='
+    queryParams[key] = value;// Add the key and value to the queryParams object
```
cjyuan, line 12: No change required.

Note: 

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,12 @@
-function tally() {}
+function tally(items) {
+  if (!Array.isArray(items)) {
+    throw new TypeError("Input must be an array");
+  }
+
+  return items.reduce((counts, item) => {
+    counts[item] = (counts[item] || 0) + 1;
+    return counts;
+  }, {});
```
```
```

cjyuan (commented):
You did a good job in rejecting invalid parameters in all functions.
I only have some suggestions.

Sprint-2/implement/contains.js
DamianDL: Removed the for loop, replaced with a console.log... .join construction.

DamianDL: Removed the for loop, replaced with a console.log... .join construction.

Changed Object.prototype.hasOwnProperty() with just Object.hasOwnProperty() to simplify code

Sprint-2/implement/contains.test.js
DamianDL: I think that you're saying that my use of !Array.isArray(obj) as a check only works because I have included the property and if I had !Array.isArray(), this would allow 0 to be found in an array of 1, 2, 3 because 1 has a property of "0".  Have I understood that correctly or are you expecting me to make changes? Thanks

Sprint-2/implement/lookup.js
DamianDL: I do have Prettier installed and format on save enabled. Resaved file to trigger formatting.

Sprint-2/implement/tally.js
DamianDL: Updated tally.js logic and included tests for strings in test script.

Sprint-2/implement/contains.test.js
cjyuan: Sorry that my comment wasn't clear enough.

I meant say the description given in the comment is not quite correct.
>> // Arrays do not have properties like objects, so it should return false

The function should return false because the first argument is an array (and not because it does not have properties).

Sprint-2/implement/contains.js
cjyuan: I meant, we can also use `Object.hasOwn(obj, prop)`.

cjyuan:
Changes look good. Well done.

illicitonion:

PR #997, opened 2026-03-11
Glasgow | Jan-26-ITP | Alasdair MacDonald | Sprint 2 | Exercises

MacDonald91:
## Learners, PR Template

Self checklist

## Changelist

Completed Sprint 1 exercises.

## Questions

n/a

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -1,14 +1,30 @@
-// Fix this implementation
-// Start by running the tests for this function
-// If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
-
-// Hint: Please consider scenarios when 'list' doesn't have numbers (the function is expected to return null)
-// or 'list' has mixed values (the function is expected to sort only numbers).
-
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+
+  // must be an array
+  if (!Array.isArray(list)) {
+    return null;
+  }
+
+  // keep only numeric values
+  const numbers = list.filter(value => typeof value === "number");
+
+  // if no numbers exist return null
+  if (numbers.length === 0) {
+    return null;
+  }
+
+  // sort numbers without modifying original list
+  const sorted = [...numbers].sort((a, b) => a - b);
+
+  const middle = Math.floor(sorted.length / 2);
+
+  // even length
+  if (sorted.length % 2 === 0) {
+    return (sorted[middle - 1] + sorted[middle]) / 2;
+  }
+
+  // odd length
+  return sorted[middle];
 }
 
 module.exports = calculateMedian;
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -1,12 +1,13 @@
 function calculateMedian(list) {
-
   // must be an array
   if (!Array.isArray(list)) {
     return null;
   }
 
-  // keep only numeric values
-  const numbers = list.filter(value => typeof value === "number");
+  // keep only VALID numeric values (exclude NaN, Infinity)
+  const numbers = list.filter(
+    (value) => typeof value === "number" && Number.isFinite(value)
+  );
 
   // if no numbers exist return null
   if (numbers.length === 0) {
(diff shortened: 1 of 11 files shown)


Sprint-1/fix/median.js
```diff
@@ -1,14 +1,30 @@
...
+  if (!Array.isArray(list)) {
+    return null;
+  }
 
-function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  // keep only numeric values
+  const numbers = list.filter(value => typeof value === "number");
```

Sprint-1/fix/median.js
```diff
@@ -1,14 +1,30 @@
...
+
+  // if no numbers exist return null
+  if (numbers.length === 0) {
+    return null;
+  }
+
+  // sort numbers without modifying original list
+  const sorted = [...numbers].sort((a, b) => a - b);
```
cjyuan, line 18: Is `numbers` the original array?

Sprint-1/implement/dedupe.test.js
```diff
@@ -1,27 +1,34 @@
...
+test("given an empty array, it returns an empty array", () => {
+  expect(dedupe([])).toEqual([]);
+});
 
 // Given an array with no duplicates
 // When passed to the dedupe function
 // Then it should return a copy of the original array
+test("given an array with no duplicates, it returns the same array", () => {
+  expect(dedupe([1, 2, 3])).toEqual([1, 2, 3]);
+});
```
cjyuan, line 23: This test should fail if the function returns the original array (instead of a **copy** of the original array).

Sprint-1/implement/max.test.js
```diff
@@ -1,43 +1,50 @@
...
+  expect(findMax(["hey", 10, "hi", 60, 10])).toBe(60);
+});
 
 // Given an array with only non-number values
 // When passed to the max function
-// Then it should return the least surprising value given how it behaves for all other inputs
+// Then it should return the least surprising value
+test("only non-number values returns -Infinity", () => {
+  expect(findMax(["a", "b", "c"])).toBe(-Infinity);
+});
```
cjyuan, line 50: When a string representing a valid numeric literal (for example, `"300"`) is compared to a number, 

---
Why return a string (instead of a value of type "number") when the given array contains only non-number values?

When a function has a dual return type, it becomes unclear what the caller should expect. Developers would need to look at the implementation or documentation to understand the behavior.

Sprint-1/implement/sum.test.js
```diff
@@ -1,36 +1,43 @@
...
+test("given negative numbers", () => {
+  expect(sum([10, -5, -5])).toBe(0);
+});
 
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
+test("given decimal numbers", () => {
+  expect(sum([1.5, 2.5, 3])).toBe(7);
+});
```
cjyuan, line 29: Decimal numbers in most programming languages (including JS) are internally represented in "floating point number" format. Floating point arithmetic is not exact. For example, the result of `46.5678 - 46 === 0.5678` is false because `46.5678 - 46` only yield a value that is very close to `0.5678`. Even changing the order in which the program add/subtract numbers can yield different values.

So the following could happen
```javascript

```

Suggestion: Look up 

cjyuan (commented):
You missed updating one file: 
`Sprint-1/refactor/includes.js`

Sprint-1/implement/sum.test.js
MacDonald91: **Thanks for the feedback!

I’ve refactored the includes function to use a for...of loop as required. 

I also fixed the test case for the empty array to ensure it correctly passes a target value.

All tests are now passing locally.**

Sprint-1/implement/dedupe.test.js
cjyuan: Can you address this comment?

Sprint-1/implement/max.test.js
cjyuan: Can you address this comment?

Sprint-1/implement/sum.test.js
cjyuan: This response does not address my request.

Note: This PR is not needed to complete Step 3. You don't have to complete this by the deadline.

cjyuan:
Still have not yet addressed all comments.

PR #998, opened 2026-03-11
London | 26-ITP-January | Eugenie Ahangama | Sprint 1 | Data Groups

Eugenie-A:
## Learners, PR Template

Self checklist

## Changelist

- Fixed `calculateMedian` implementation in `median.js`
- Implemented `dedupe.js` and wrote tests in `dedupe.test.js`
- Implemented `findMax` in `max.js` and wrote tests in `max.test.js`
- Implemented `sum.js` and wrote tests in `sum.test.js`
- Refactored `includes.js` from a `for` loop to a `for...of loop` and added comments to both files

## Questions

What makes a good test? What makes a bad test?

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -6,9 +6,19 @@
 // or 'list' has mixed values (the function is expected to sort only numbers).
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  // Return null for non arrays
+  if (!Array.isArray(list)) return null;
+  // Filter to only numeric values
+  const nums = list.filter((x) => typeof x === "number" && isFinite(x));
+  // Return null if no valid numbers remain
+  if (nums.length === 0) return null;
+
+  const sorted = [...nums].sort((a, b) => a - b);
+  const middleIndex = Math.floor(sorted.length / 2);
+
+  return sorted.length % 2 !== 0
+    ? sorted[middleIndex]
+    : (sorted[middleIndex - 1] + sorted[middleIndex]) / 2;
 }
 
 module.exports = calculateMedian;
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -12,10 +12,13 @@ function calculateMedian(list) {
   const nums = list.filter((x) => typeof x === "number" && isFinite(x));
   // Return null if no valid numbers remain
   if (nums.length === 0) return null;
-
-  const sorted = [...nums].sort((a, b) => a - b);
+  // Sort the numbers in ascending order
+  // filter already returns a new array so need to spread before spreading
+  const sorted = nums.sort((a, b) => a - b);
+  // Find the middle index of the sorted array
   const middleIndex = Math.floor(sorted.length / 2);
-
+  // If odd length, return the middle index
+  // If even length, return the average of the two middle elements
   return sorted.length % 2 !== 0
     ? sorted[middleIndex]
     : (sorted[middleIndex - 1] + sorted[middleIndex]) / 2;
--- a/Sprint-1/implement/dedupe.test.js
+++ b/Sprint-1/implement/dedupe.test.js
@@ -23,9 +23,13 @@ test("given an empty array, it returns an empty array", () => {
 // Given an array with no duplicates
 // When passed to the dedupe function
 // Then it should return a copy of the original array
-test("given an araay with no duplicates, it returns a copy of the original array", () => {
-  expect(dedupe([1, 2, 3])).toEqual([1, 2, 3]);
-  expect(dedupe(["a", "b", "c"])).toEqual(["a", "b", "c"]);
+test("given an array with no duplicates, it returns a copy of the original array", () => {
+  const input = [1, 2, 3];
+  const result = dedupe(input);
+  // Check the contents are the same
+  expect(result).toEqual(input);
+  // Check it's a different array, not the same reference
+  expect(result).not.toBe(input);
 });
 
 // Given an array with strings or numbers
--- a/Sprint-1/implement/dedupe.test.js
+++ b/Sprint-1/implement/dedupe.test.js
@@ -26,12 +26,10 @@ test("given an empty array, it returns an empty array", () => {
 test("given an array with no duplicates, it returns a copy of the original array", () => {
   const input = [1, 2, 3];
   const result = dedupe(input);
-  // Check the contents are the same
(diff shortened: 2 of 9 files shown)


Sprint-1/fix/median.js
```diff
@@ -6,9 +6,19 @@
...
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  // Return null for non arrays
+  if (!Array.isArray(list)) return null;
+  // Filter to only numeric values
+  const nums = list.filter((x) => typeof x === "number" && isFinite(x));
+  // Return null if no valid numbers remain
+  if (nums.length === 0) return null;
+
+  const sorted = [...nums].sort((a, b) => a - b);
```
cjyuan: Why make a copy of `nums` before sorting?

Sprint-1/implement/dedupe.test.js
```diff
@@ -16,12 +16,23 @@ E.g. dedupe([1, 2, 1]) target output: [1, 2]
...
+  expect(dedupe([])).toEqual([]);
+});
 
 // Given an array with no duplicates
 // When passed to the dedupe function
 // Then it should return a copy of the original array
+test("given an araay with no duplicates, it returns a copy of the original array", () => {
+  expect(dedupe([1, 2, 3])).toEqual([1, 2, 3]);
+  expect(dedupe(["a", "b", "c"])).toEqual(["a", "b", "c"]);
+});
```
cjyuan, line 33: This test should fail if the function returns the original array (instead of a **copy** of the original array).

Sprint-1/implement/max.test.js
```diff
@@ -16,28 +16,49 @@ const findMax = require("./max.js");
...
+test("given an array with non-numeric values, returns the max and ignores non-numerics", () => {
+  expect(findMax(["hey", 10, "hi", 60, 10])).toEqual(60);
+});
 
 // Given an array with only non-number values
 // When passed to the max function
 // Then it should return the least surprising value given how it behaves for all other inputs
+test("given an array with only non-numeric values, returns -Infinity", () => {
+  expect(findMax(["hey", "hi", null, undefined])).toEqual(-Infinity);
+});
```

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,42 @@ const sum = require("./sum.js");
...
+  expect(sum([-10, -20, -30])).toEqual(-60);
+  expect(sum([-5, -3, -2])).toEqual(-10);
+});
 
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
+test("given an array with decimal numbers, returns the correct sum", () => {
+  expect(sum([1.5, 2.5, 3.0])).toEqual(7);
+});
```
cjyuan, line 42: Decimal numbers in most programming languages (including JS) are internally represented in "floating point number" format. Floating point arithmetic is not exact. For example, the result of `46.5678 - 46 === 0.5678` is false because `46.5678 - 46` only yield a value that is very close to `0.5678`. Even changing the order in which the program add/subtract numbers can yield different values.

So the following could happen
```javascript

```

Suggestion: Look up 

Sprint-1/fix/median.js
Eugenie-A: I used `[...nums]` because I wanted to avoid mutating the input array `list`. However, since `nums` is already created by `filter`, it's a brand new array and not a reference to `list`. So sorting `nums` directly won't affect the original input, making the spread unnecessary. I'll remove it.

Sprint-1/implement/dedupe.test.js
Eugenie-A: The additional check is `expect(result).not.toBe(input)` which verifies that the returned array is a different reference in memory, not the same array. I've also fixed the typo in the test name.

Sprint-1/implement/max.test.js
Eugenie-A: Thank you for the suggestion.

I've added `"300"` as a test case to verify that numeric strings are correctly ignored. The `typeof x === 'number'` check in `max.js` already handles this correctly since `"300"` is a string type, not a number.

Sprint-1/implement/sum.test.js
Eugenie-A: I wasn't aware of floating point precision issues in JavaScript. I've added a test case using `toBeCloseTo` instead of `toEqual` for floating point calculations, which checks that the result is close enough to the expected value rather than requiring an exact match.

cjyuan (commented):
Other changes are good.

Sprint-1/implement/dedupe.test.js
```diff
@@ -23,9 +23,13 @@ test("given an empty array, it returns an empty array", () => {
...
-  expect(dedupe([1, 2, 3])).toEqual([1, 2, 3]);
-  expect(dedupe(["a", "b", "c"])).toEqual(["a", "b", "c"]);
+test("given an array with no duplicates, it returns a copy of the original array", () => {
+  const input = [1, 2, 3];
+  const result = dedupe(input);
+  // Check the contents are the same
+  expect(result).toEqual(input);
+  // Check it's a different array, not the same reference
+  expect(result).not.toBe(input);
 });
```
cjyuan, line 33: Your function is correct, but currently there is a chance that an incorrectly implemented function can pass this test. Can you figure out why?

Eugenie-A: A function that just returns a copy without deduplicating (e.g. `return [...arr]`) would still pass since `[1, 2, 3]` has no duplicates. 

I've added an extra check using an array with duplicates `[1, 1, 2, 3]` to make the test more robust.

Sprint-1/implement/dedupe.test.js
```diff
@@ -30,6 +30,8 @@ test("given an array with no duplicates, it returns a copy of the original array
   expect(result).toEqual(input);
   // Check it's a different array, not the same reference
   expect(result).not.toBe(input);
+  // Check that a duplicate input is correctly deduped in the result
+  expect(dedupe([1, 1, 2, 3])).toEqual([1, 2, 3]);
```
cjyuan: The case of "array with duplicates" is checked else where.

I was referring to the specific case of "array with no duplicates".
Hint: Is it possible that `result` and `input` can have the same **wrong** content?

Eugenie-A: I see, if dedupe mutated `input`, then `toEqual(input)` would still pass since both would have the same wrong content. I've fixed it by comparing `result` against a hardcoded `[1, 2, 3]` instead.

cjyuan:
All good now. Well done!

illicitonion:

PR #1005, opened 2026-03-13
GLASGOW | Jan-26-ITP | Prakash Dcosta | Sprint 1 | Data Groups

dcostaprakash:
Self checklist

- [X] My changes follow the [style guide]

## Changelist

In this PR, i have implemented functions according to requrements and built test cases

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -6,9 +6,29 @@
 // or 'list' has mixed values (the function is expected to sort only numbers).
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  // It must be an array
+  if (!Array.isArray(list)) {
+    return null;
+  }
+  // Keep only numeric values
+  const numbers = list.filter(
+    (value) => typeof value === "number" && !isNaN(value)
+  );
+
+  // If there are no numbers existing then return null
+  if (numbers.length === 0) return null;
+
+  // Sort numbers without modifying original list
+  const sorted = [...numbers].sort((a, b) => a - b);
+  const middle = Math.floor(sorted.length / 2);
+
+  // For even length
+  if (sorted.length % 2 === 0) {
+    return (sorted[middle - 1] + sorted[middle]) / 2;
+  }
+
+  // For odd length
+  return sorted[middle];
 }
 
 module.exports = calculateMedian;
--- a/Sprint-1/implement/dedupe.test.js
+++ b/Sprint-1/implement/dedupe.test.js
@@ -23,9 +23,11 @@ test("given an empty array, it returns an empty array", () => {
 // Given an array with no duplicates
 // When passed to the dedupe function
 // Then it should return a copy of the original array
-test("returns original array if there are no duplicates", () => {
-  expect(dedupe([10, 45, 85, 20])).toEqual([10, 45, 85, 20]);
-  expect(dedupe(["hello", "a", "hi", "b"])).toEqual(["hello", "a", "hi", "b"]);
+test("returns copy of the original array if there are no duplicates", () => {
+  const input = [10, 45, 85, 20];
+  const result = dedupe(input);
+  expect(result).toEqual(input);
+  expect(result).not.toBe(input);
 });
 
 // Given an array with strings or numbers
--- a/Sprint-1/implement/dedupe.test.js
+++ b/Sprint-1/implement/dedupe.test.js
@@ -27,6 +27,7 @@ test("returns copy of the original array if there are no duplicates", () => {
   const input = [10, 45, 85, 20];
   const result = dedupe(input);
   expect(result).toEqual(input);
+  expect(input).toEqual([10, 45, 85, 20]);
   expect(result).not.toBe(input);
 });
 
--- a/Sprint-1/implement/dedupe.test.js
+++ b/Sprint-1/implement/dedupe.test.js
@@ -16,12 +16,26 @@ E.g. dedupe([1, 2, 1]) target output: [1, 2]
 // Given an empty array
 // When passed to the dedupe function
 // Then it should return an empty array
-test.todo("given an empty array, it returns an empty array");
+test("given an empty array, it returns an empty array", () => {
+  expect(dedupe([])).toEqual([]);
+});
 
 // Given an array with no duplicates
 // When passed to the dedupe function
 // Then it should return a copy of the original array
+test("returns original array if there are no duplicates", () => {
+  expect(dedupe([10, 45, 85, 20])).toEqual([10, 45, 85, 20]);
+  expect(dedupe(["hello", "a", "hi", "b"])).toEqual(["hello", "a", "hi", "b"]);
+});
 
 // Given an array with strings or numbers
 // When passed to the dedupe function
 // Then it should remove the duplicate values, preserving the first occurence of each element
+test("returns first occurance for array of strings or numbers", () => {
+  expect(dedupe([10, 10, 20, 20, 30, 40])).toEqual([10, 20, 30, 40]);
+  expect(dedupe(["hello", "hello", "hi", "a", "hi"])).toEqual([
+    "hello",
+    "hi",
+    "a",
+  ]);
+});
--- a/Sprint-1/implement/max.test.js
+++ b/Sprint-1/implement/max.test.js
@@ -16,28 +16,48 @@ const findMax = require("./max.js");
 // When passed to the max function
 // Then it should return -Infinity
 // Delete this test.todo and replace it with a test.
-test.todo("given an empty array, returns -Infinity");
+test("given an empty array, returns -Infinity", () => {
+  expect(findMax([])).toBe(-Infinity);
+});
 
 // Given an array with one number
 // When passed to the max function
 // Then it should return that number
+test("given an array with one number, returns that number", () => {
+  expect(findMax([5])).toBe(5);
+});
 
 // Given an array with both positive and negative numbers
 // When passed to the max function
 // Then it should return the largest number overall
+test("returns the largest number overall", () => {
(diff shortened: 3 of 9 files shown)


Sprint-1/fix/median.js
```diff
@@ -6,9 +6,29 @@
...
+  // Keep only numeric values
+  const numbers = list.filter(
+    (value) => typeof value === "number" && !isNaN(value)
+  );
+
+  // If there are no numbers existing then return null
+  if (numbers.length === 0) return null;
+
+  // Sort numbers without modifying original list
+  const sorted = [...numbers].sort((a, b) => a - b);
```
cjyuan, line 22: Is `numbers` the original list?

Sprint-1/implement/dedupe.test.js
```diff
@@ -16,12 +16,26 @@ E.g. dedupe([1, 2, 1]) target output: [1, 2]
...
+  expect(dedupe([])).toEqual([]);
+});
 
 // Given an array with no duplicates
 // When passed to the dedupe function
 // Then it should return a copy of the original array
+test("returns original array if there are no duplicates", () => {
+  expect(dedupe([10, 45, 85, 20])).toEqual([10, 45, 85, 20]);
+  expect(dedupe(["hello", "a", "hi", "b"])).toEqual(["hello", "a", "hi", "b"]);
+});
```
cjyuan, line 32: This test should fail if the function returns the original array (instead of a **copy** of the original array).

Sprint-1/implement/max.test.js
```diff
@@ -16,28 +16,48 @@ const findMax = require("./max.js");
...
+test("returnes max and ignores non numeric values", () => {
+  expect(findMax(["stay", 3, "quiet", -32, 5])).toBe(5);
+});
 
 // Given an array with only non-number values
 // When passed to the max function
 // Then it should return the least surprising value given how it behaves for all other inputs
+test("returns infinity in non number values", () => {
+  expect(findMax(["hi", "sally", "a", "alpha"])).toBe(-Infinity);
+});
```

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,11 @@
-function sum(elements) {
+function sum(arr) {
+  let total = 0;
+  for (let i = 0; i < arr.length; i++) {
+    if (typeof arr[i] === "number") {
```
cjyuan: Do you plan to include also `-Infinity`, `Infinity`, and `NaN` in the calculation?

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,39 @@ const sum = require("./sum.js");
...
+test("given an array with negative numbers, returns correct sum", () => {
+  expect(sum([-50, -20, 10])).toBe(-60);
+});
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
-
+test("given an array with decimal/float numbers, returns correct sum", () => {
+  expect(sum([2.8, 4.2, 7.7])).toBe(14.7);
+});
```
cjyuan, line 38: Decimal numbers in most programming languages (including JS) are internally represented in "floating point number" format. Floating point arithmetic is not exact. For example, the result of `46.5678 - 46 === 0.5678` is false because `46.5678 - 46` only yield a value that is very close to `0.5678`. Even changing the order in which the program add/subtract numbers can yield different values.

So the following could happen
```javascript

```

Suggestion: Look up 

Sprint-1/fix/median.js
dcostaprakash: No numbers is not the original list. ...numbers creates a copy. The sort() with sort the copy not the original. Should i put a remark there?

Sprint-1/implement/dedupe.test.js
dcostaprakash: Yes, understood, i have modified it accordingly. Tks

Sprint-1/implement/max.test.js
dcostaprakash: Got it. I have made the change. Thank you

Sprint-1/implement/sum.js
dcostaprakash: Yes, they will ger included since they are type of "number". Do i need to exclude it?

Sprint-1/fix/median.js
cjyuan: I meant to point out, "is it necessary to make a copy of `numbers` before sorting"?

Sprint-1/implement/sum.js
cjyuan: What do you expect from the following function calls?

```
sum([NaN, 1]);
sum([Infinity, -Infinity]);
```

No change required if they meet your expectation.

Sprint-1/implement/sum.test.js
dcostaprakash: Thank you for this. I looked it up and made the change accordingly.

dcostaprakash:
Made changes as per the review. Thank you

Sprint-1/fix/median.js
cjyuan, line 22: Is it necessary to make a copy of `numbers` before sorting?

Sprint-1/implement/dedupe.test.js
```diff
@@ -16,12 +16,28 @@ E.g. dedupe([1, 2, 1]) target output: [1, 2]
...
 
 // Given an array with no duplicates
 // When passed to the dedupe function
 // Then it should return a copy of the original array
+test("returns copy of the original array if there are no duplicates", () => {
+  const input = [10, 45, 85, 20];
+  const result = dedupe(input);
+  expect(result).toEqual(input);
+  expect(result).not.toBe(input);
```
cjyuan, line 31: Currently, there is a chance that, even though `result` has incorrect elements (for example, `[]`), 

Sprint-1/implement/max.test.js
```diff
@@ -16,28 +16,48 @@ const findMax = require("./max.js");
...
+test("returns max and ignores non number values", () => {
+  expect(findMax(["stay", "300", "quiet", -32, 5])).toBe(5);
+});
 
 // Given an array with only non-number values
 // When passed to the max function
 // Then it should return the least surprising value given how it behaves for all other inputs
+test("returns infinity in non number values", () => {
+  expect(findMax(["hi", "sally", "a", "alpha"])).toBe(-Infinity);
+});
```
cjyuan, line 63: Why not include some other types of value in this test? For example, `null`, `undefined`, numbers.

Sprint-1/fix/median.js
dcostaprakash: No, it is not necessary to make a copy of numbers before sorting, however, it is a good practice.

dcostaprakash: No, it is not necessary to make a copy of numbers before sorting, however, it is a good practice.

cjyuan: Can you elaborate how it is a good practice in this scenario?

Sprint-1/implement/max.test.js
dcostaprakash: Done. Have added them the test cases. They failed and i then checked the function and accordingly made a change there for Infinity. Thank you

Sprint-1/implement/sum.js
dcostaprakash: I checked. For NaN and Number and for Infinity & -Infinity, the output was coming to NaN. I have made the change in the code to exclude NaN and sum it up. I have also made a change to the test cases also. Thank you

Sprint-1/implement/dedupe.test.js
dcostaprakash: I have made tha change to the test. Pl confirm if its fine. Thank you

dcostaprakash:
Have made the changes as per the feedback. Thank you

cjyuan:
Can you address this follow-up comment: https://github.com/CodeYourFuture/Module-Data-Groups/pull/1005#discussion_r2981920684

image


Other changes are good.

Sprint-1/fix/median.js
dcostaprakash: In this scenario it is not needed. I just tried to make it cleaner by removing non numeric values if any. Do you want me to remove it?

cjyuan: "... to make it cleaner by removing non numeric values" -- You have already removed non numeric values using `.filter()`. 

You can keep the code unchanged. I am more interested in knowing **why** you think  "making a copy of `numbers` before sorting" is a good practice. Can you explain?

dcostaprakash: Ok. I got your point. According to me, making a copy before sorting is a good practice for the following reasons:
1) We could accidentally modify it for example if we use the median function directly, the new file would be sorted and it would replace the original file
2) Incase we need to refer the original file later in the code, we will not be able to since the index of the original file could get sorted
I hope this answers. If i have missed anything, pl add to it. Thank you

cjyuan: - By "file", do you mean "array"?

- It's true that `numbers.sort()` can mutate the `numbers` array. If we need to use `numbers` after sorting, then it is considered a safe practice to not mutate the array. In your implementation, is `numbers` being used after sorting?

dcostaprakash: • Yes by file I meant array.
• No, in my current implementation code, the numbers array is not used again after the sorted variable is created.

cjyuan: "immutability" do has its advantage. The reasons you gave do not quite fit the `numbers` array as the array is clearly not needed after sorting.

Thought you might be interested in this ChatGPT explaination:
https://chatgpt.com/share/69cf697a-5ea8-8331-bb33-47ccc8b02a17

dcostaprakash: Ok. I will look at it in detail. Thank you

illicitonion:

PR #1006, opened 2026-03-13
London | 26-ITP-January | Eugenie Ahangama | Sprint 3 | Todo List App

Eugenie-A:
## Learners, PR Template

Self checklist

## Changelist

- Added the "Delete completed tasks" button in `index.html`
- Implemented deleteCompleted() in `todos.mjs`
- Wrote Jest tests for deleteCompleted() in `todos.test.mjs`
- Added event listener for delete completed button in `script.mjs`
- Implemented deleteCompletedTodos callback in `script.mjs`

## Questions

Why is render without parentheses a bug?

### DIFF
--- a/Sprint-3/todo-list/index.html
+++ b/Sprint-3/todo-list/index.html
@@ -1,40 +1,50 @@
-<!DOCTYPE html>
+<!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8" />
-  <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
-  <title>ToDo List</title>
-  <link rel="stylesheet" href="style.css" />
-  <link href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/6.4.0/css/all.min.css" rel="stylesheet">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>ToDo List</title>
+    <link rel="stylesheet" href="style.css" />
+    <link
+      href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/6.4.0/css/all.min.css"
+      rel="stylesheet"
+    />
 
-  <script type="module" src="script.mjs"></script>
-</head>
-<body>
-  <div class="todo-container">
-    <h1>My ToDo List</h1>
+    <script type="module" src="script.mjs"></script>
+  </head>
+  <body>
+    <div class="todo-container">
+      <h1>My ToDo List</h1>
 
-    <div class="todo-input">
-      <input type="text" id="new-task-input" placeholder="Enter a new task..." />
-      <button id="add-task-btn">Add</button>
-    </div>
+      <div class="todo-input">
+        <input
+          type="text"
+          id="new-task-input"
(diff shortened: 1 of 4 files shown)


Sprint-3/todo-list/index.html
```diff
@@ -1,40 +1,50 @@
-<!DOCTYPE html>
+<!doctype html>
```
kyle-tightest, line 1: this is not case sensitive so it doesn't matter but it is more standard to capitalise this.  
did your formatter do this automatically?

Sprint-3/todo-list/todos.mjs
```diff
@@ -26,4 +26,10 @@ export function toggleCompletedOnTask(todos, taskIndex) {
...
   }
-}
\ No newline at end of file
+}
+
+// Delete all completed tasks from todos[]
+export function deleteCompleted(todos) {
+  const completed = todos.filter((todo) => todo.completed);
+  completed.forEach((todo) => todos.splice(todos.indexOf(todo), 1));
+}
```
kyle-tightest, line 34: While this code works, it is inefficient as you are iterating over the entire array when you do `filter` and then iterating over each of the filtered array items when you do `forEach`.  
The `.filter` method returns an array filtered by the condition you give it.  
Try and see if you can use the `.filter` method differently to achieve the same outcome, without needing `forEach`?

Sprint-3/todo-list/index.html
Eugenie-A: Yes, that was my Prettier formatter in VS Code doing it automatically on save. I know it's more standard to keep it uppercase but Prettier lowercases it by default. Prettier formatter is the standard formatter we were asked to use in our work.

Sprint-3/todo-list/todos.mjs
Eugenie-A: Thank you for the feedback. I've updated the function to use `filter` once to get the incomplete todos, then `splice` to update the original array in place.

Sprint-3/todo-list/index.html
kyle-tightest: Okay no problem

Sprint-3/todo-list/todos.mjs
kyle-tightest: Thanks, this works as well. 
But again, when you do `splice` in that second step you're iterating over each of the filtered array items.  
This is not necessary.  
Try to look at the function again and see if you can do it in 1 line, instead of 2?

Eugenie-A: I've combined both lines into one, the `filter` now happens directly inside the `splice` call so it's done in a single step.

kyle-tightest: Almost there! you can remove the `splice` and just use filter

Eugenie-A: If I just use `filter` and `return` the result, won't the tests fail since they check the original todos array after calling the function?

kyle-tightest: Ah sorry you're 100% correct, the function is built to change the array in place instead of returning a new array as a result. Okay nevermind then this is perfect

illicitonion:

PR #1007, opened 2026-03-14
London | 26-ITP-JAN | Xuanming Hu | Sprint 1 | Data Groups

Samual-Hu:
## Learners, PR Template

Self checklist

## Changelist

Completed all exercises in the Sprint 1 directory.

## Questions

N/A

### DIFF
--- a/Sprint-1/implement/max.js
+++ b/Sprint-1/implement/max.js
@@ -1,4 +1,15 @@
 function findMax(elements) {
+    if (!Array.isArray(elements)) {
+    return null;
+  }
+
+  const numericElements = elements.filter(item => typeof item === "number");
+
+  if (numericElements.length === 0) {
+    return -Infinity;
+  }
+
+  return Math.max(...numericElements);
 }
 
 module.exports = findMax;
--- a/Sprint-1/implement/max.js
+++ b/Sprint-1/implement/max.js
@@ -1,9 +1,9 @@
 function findMax(elements) {
-    if (!Array.isArray(elements)) {
+  if (!Array.isArray(elements)) {
     return null;
   }
 
-  const numericElements = elements.filter(item => typeof item === "number");
+  const numericElements = elements.filter((item) => typeof item === "number");
 
   if (numericElements.length === 0) {
     return -Infinity;
--- a/Sprint-1/implement/max.test.js
+++ b/Sprint-1/implement/max.test.js
@@ -48,11 +48,21 @@ test("given an array with decimal numbers, returns the largest decimal number", 
 // When passed to the max function
 // Then it should return the max and ignore non-numeric values
 test("given an array with non-number values, returns the max and ignores non-numeric values", () => {
-  expect(findMax(["apple", "banana", 10, "cherry", 20])).toEqual(20);
+  expect(findMax(["apple", "banana", 10, "300", 20])).toEqual(20);
 });
 // Given an array with only non-number values
 // When passed to the max function
 // Then it should return the least surprising value given how it behaves for all other inputs
 test("given an array with only non-number values, returns -Infinity", () => {
   expect(findMax(["apple", "banana", "cherry"])).toEqual(-Infinity);
+});
+
+//test NaN
+test("NaN", () => {
+  expect(findMax([NaN])).toEqual(NaN);
+});
+
+//test with variables
+test("variables", () => {
+  expect(findMax([0, NaN, 1])).toEqual(NaN);
 });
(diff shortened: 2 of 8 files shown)


Sprint-1/implement/max.js
```diff
@@ -1,4 +1,15 @@
 function findMax(elements) {
+    if (!Array.isArray(elements)) {
+    return null;
+  }
```
cjyuan: Indentation is off.

?

Sprint-1/implement/max.test.js
```diff
@@ -16,28 +16,43 @@ const findMax = require("./max.js");
...
-
+test("given an array with non-number values, returns the max and ignores non-numeric values", () => {
+  expect(findMax(["apple", "banana", 10, "cherry", 20])).toEqual(20);
+});
 // Given an array with only non-number values
 // When passed to the max function
 // Then it should return the least surprising value given how it behaves for all other inputs
+test("given an array with only non-number values, returns -Infinity", () => {
+  expect(findMax(["apple", "banana", "cherry"])).toEqual(-Infinity);
+});
```

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,15 @@
...
+
+  const numericElements = elements.filter(item => typeof item === "number");
+
+  if (numericElements.length === 0) {
+    return -Infinity;
+  }
+
+  return Math.max(...numericElements);
```
cjyuan: What do you expect from the following function calls (on extreme cases)?

```
findMax([NaN])
findMax([0, NaN, 1])
```

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,17 @@
...
+
+  let total = 0;
+
+  for (let i = 0; i < elements.length; i++) {
+    if (typeof elements[i] === "number") {
+      total += elements[i];
+    }
+  }
+
+  return total;
```
cjyuan, line 14: What do you expect from the following function calls (on extreme cases)?

```
sum([NaN, 1]);
sum([Infinity, -Infinity]);
```

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,17 @@
 function sum(elements) {
+    if (!Array.isArray(elements)) {
+    return null;
+  }
```
cjyuan: Indentation is off.

Sprint-1/implement/max.js
Samual-Hu: I had installed prettier previously, but it wasn't working because I didn't check "Format On Save" in the settings. I've resolved this now.

Sprint-1/implement/sum.js
Samual-Hu: I've resolved this now.

Sprint-1/implement/max.test.js
Samual-Hu: I have tested `"300"`, and the function can correctly ignore non-numeric values.

Sprint-1/implement/max.js
Samual-Hu: For `findMax([NaN])`, I expected `-Infinity`, but the actual result is `NaN`.
For `findMax([0, NaN, 1])`, I expected 1, but the actual result is also `NaN`.
This is because the type of `NaN` is number.

Sprint-1/implement/sum.js
Samual-Hu: For `sum([NaN, 1])`, I expected 1, but the actual result is `NaN`.
For `sum([Infinity, -Infinity])`, I expected 0, but the actual result is `NaN`.
This is also because the type of `NaN` is number.

Samual-Hu:
Hi @cjyuan , Thank you for your comments. I have resolved the indentation issues and conducted additional tests.

Sprint-1/implement/max.test.js
cjyuan: Your function is good. We prepare test not just to test our own implementation, but also to ensure future modification (by someone else) won't unexpectedly change the behavior of the function.

Sprint-1/implement/max.js
cjyuan: If the function does not behave the way you expected, why not update the function so that it matches your expectation?

cjyuan:
Thanks for the explanations.

Updating the check in `max.js` and `sum.js` to filter out `NaN` should be straight forward. 

So all good.

illicitonion:

PR #1014, opened 2026-03-15
Sheffield | 26-ITP-Jan | Mahmoud Shaabo | Sprint 1 | Module-Data-Groups

mahmoudshaabo1984:
Completed all mandatory exercises for Sprint 1 in the following directories: fix, implement, and refactor.

### Progress Checklist:
- [x] **fix/median.js:** Fixed median calculation, added guard clause for non-array inputs, and handled `NaN` edge cases after reviewing JavaScript documentation.
- [x] **implement/max.js:** Implemented logic using `Math.max` and spread syntax with `NaN` filtering to ensure robust numerical operations.
- [x] **implement/sum.js:** Built the summation logic using `.reduce()` while ignoring non-numeric values and ensuring `NaN` safety.
- [x] **implement/dedupe.js:** Cleaned up duplicates using the modern `Set` object.
- [x] **refactor/includes.js:** Refactored traditional for-loops into modern `for...of` syntax for better readability.
- [x] **Testing:** Verified that all 24 tests passed successfully using `npm test`.

---

### Personal Note for CJ:
Hi CJ, 
I have ensured that all checkboxes follow the `[x]` syntax. During this Sprint, I spent some time researching JavaScript documentation regarding numerical types, which led me to implement additional checks for `NaN` values to make my functions more robust. Navigating the test results with NVDA helped me identify a few syntax errors early on. Everything is now passing and aligned with the required best practices. Looking forward to your feedback!

### DIFF
--- a/Sprint-1/implement/dedupe.test.js
+++ b/Sprint-1/implement/dedupe.test.js
@@ -1,27 +1,36 @@
 const dedupe = require("./dedupe.js");
-/*
-Dedupe Array
-
-📖 Dedupe means **deduplicate**
-
-In this kata, you will need to deduplicate the elements of an array
-
-E.g. dedupe(['a','a','a','b','b','c']) target output: ['a','b','c']
-E.g. dedupe([5, 1, 1, 2, 3, 2, 5, 8]) target output: [5, 1, 2, 3, 8]
-E.g. dedupe([1, 2, 1]) target output: [1, 2]
-*/
-
-// Acceptance Criteria:
 
 // Given an empty array
 // When passed to the dedupe function
 // Then it should return an empty array
-test.todo("given an empty array, it returns an empty array");
+test("given an empty array, it returns an empty array", () => {
+  expect(dedupe([])).toEqual([]);
+});
 
 // Given an array with no duplicates
 // When passed to the dedupe function
 // Then it should return a copy of the original array
+test("given an array with no duplicates, returns a copy of the array", () => {
+  expect(dedupe([1, 2, 3])).toEqual([1, 2, 3]);
+});
 
-// Given an array with strings or numbers
+// Given an array with duplicate strings
 // When passed to the dedupe function
-// Then it should remove the duplicate values, preserving the first occurence of each element
+// Then it should remove duplicates, preserving first occurrence
+test("removes duplicate strings", () => {
+  expect(dedupe(["a", "a", "a", "b", "b", "c"])).toEqual(["a", "b", "c"]);
+});
+
+// Given an array with duplicate numbers
+// When passed to the dedupe function
+// Then it should remove duplicates, preserving first occurrence
+test("removes duplicate numbers", () => {
+  expect(dedupe([5, 1, 1, 2, 3, 2, 5, 8])).toEqual([5, 1, 2, 3, 8]);
+});
+
+// Given an array with mixed duplicates
+// When passed to the dedupe function
+// Then it should remove duplicates preserving first occurrence
+test("removes duplicates in mixed array", () => {
+  expect(dedupe([1, 2, 1])).toEqual([1, 2]);
+});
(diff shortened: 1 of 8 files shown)


Sprint-1/implement/dedupe.test.js
```diff
@@ -1,27 +1,36 @@
...
+  expect(dedupe([])).toEqual([]);
+});
 
 // Given an array with no duplicates
 // When passed to the dedupe function
 // Then it should return a copy of the original array
+test("given an array with no duplicates, returns a copy of the array", () => {
+  expect(dedupe([1, 2, 3])).toEqual([1, 2, 3]);
+});
+
```
cjyuan, line 21: This test should fail if the function returns the original array (instead of a **copy** of the original array).

Sprint-1/implement/dedupe.test.js
```diff
@@ -1,27 +1,36 @@
...
+});
 
-// Given an array with strings or numbers
+// Given an array with mixed duplicates
 // When passed to the dedupe function
-// Then it should remove the duplicate values, preserving the first occurence of each element
+// Then it should remove duplicates preserving first occurrence
+test("removes duplicates in mixed array", () => {
+  expect(dedupe([1, 2, 1])).toEqual([1, 2]);
+});
```
cjyuan: I think "mixed array" means the array contains both numbers and strings.

Sprint-1/implement/max.test.js
```diff
@@ -1,43 +1,50 @@
...
+test("returns the largest decimal number", () => {
+  expect(findMax([1.1, 2.5, 0.9, 2.4])).toBe(2.5);
+});
 
 // Given an array with non-number values
 // When passed to the max function
 // Then it should return the max and ignore non-numeric values
+test("ignores non-numeric values and returns the max number", () => {
+  expect(findMax(["hey", 10, "hi", 60, 10])).toBe(60);
+});
```
cjyuan, line 48: Note: When a string representing a valid numeric literal (for example, `"300"`) is compared to a number, 

Sprint-1/implement/sum.test.js
```diff
@@ -1,36 +1,43 @@
...
+test("correctly sums arrays containing negative numbers", () => {
+  expect(sum([10, -3, 5])).toBe(12);
+});
 
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
+test("correctly sums decimal numbers", () => {
+  expect(sum([1.5, 2.5, 3.0])).toBe(7);
+});
```
cjyuan, line 31: Decimal numbers in most programming languages (including JS) are internally represented in "floating point number" format. Floating point arithmetic is not exact. For example, the result of `46.5678 - 46 === 0.5678` is false because `46.5678 - 46` only yield a value that is very close to `0.5678`. Even changing the order in which the program add/subtract numbers can yield different values.

So the following could happen
```javascript

```

Suggestion: Look up 

cjyuan (commented):
Code looks good. I only have a few suggestions.

mahmoudshaabo1984:
Hi CJ,
I have updated the test files in the sprint-1/implement folder based on your feedback. Specifically, I have made the following changes:
. 1
sum.test.js: Switched to toBeCloseTo() to handle floating-point precision correctly.
. 2
dedupe.test.js: Added a test case for mixed-type arrays and included an immutability check to ensure the function returns a new array reference.
. 3
max.test.js: Added a test case to ensure numeric strings (like "300") are ignored and not coerced into numbers.
I also fixed the Markdown syntax for the checkboxes in the PR description as requested. All 47 tests are now passing. Ready for your review!
Best regards,
Mahmoud

cjyuan:
Changes look good.

illicitonion:

PR #1016, opened 2026-03-16
Manchester | 26-ITP-Jan | Liban Jama | Sprint 1 | Data Groups 

libanj0161:
## Learners, PR Template

Self checklist

## Changelist

Completed Fix under data groups module.

## Questions

N/A.

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -11,16 +11,10 @@ function calculateMedian(list) {
     return null;
   }
 
-  // create a variable called numbers and store an empty array.
-  let numbers = [];
-
-  // loop over each item in list.
-  for (const item of list) {
-    // if item is a number, push it into the numbers array.
-    if (typeof item === "number") {
-      numbers.push(item);
-    }
-  }
+  // filter the list to include only valid numbers (excludes Infinity, -Infinity, and NaN)
+  const numbers = list.filter(
+    (item) => typeof item === "number" && Number.isFinite(item)
+  );
 
   // if numbers array is empty return null.
   if (numbers.length === 0) {
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -6,9 +6,39 @@
 // or 'list' has mixed values (the function is expected to sort only numbers).
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  // if list is not an array return null.
+  if (!Array.isArray(list)) {
+    return null;
+  }
+
+  // create a variable called numbers and store an empty array.
+  let numbers = [];
+
+  // loop over each item in list.
+  for (const item of list) {
+    // if item is a number, push it into the numbers array.
+    if (typeof item === "number") {
+      numbers.push(item);
+    }
+  }
+
+  // if numbers array is empty return null.
+  if (numbers.length === 0) {
+    return null;
+  }
+
+  // sort the numbers in the number array in ascending order.
+  numbers.sort((a, b) => a - b);
+
+  // find the middle index and store it in a variable.
+  const middleIndex = Math.floor(numbers.length / 2);
+
+  // if its an even array there are two middle indexes, so add them together and divide by two.
+  if (numbers.length % 2 === 0) {
+    return (numbers[middleIndex] + numbers[middleIndex - 1]) / 2;
+  }
+  // for odd arrays return middle index.
+  return numbers[middleIndex];
 }
 
 module.exports = calculateMedian;
--- a/Sprint-1/implement/dedupe.test.js
+++ b/Sprint-1/implement/dedupe.test.js
@@ -32,21 +32,15 @@ test("given an array with no duplicates, it should return a copy of the original
   expect(result).not.toBe(input);
 });
 
-// Given an array of strings
+// Given an array with strings
 // When passed to the dedupe function
-// Then it should not modify the original array
-test("given an array with strings, it should not modify the original array", () => {
-  const input = ["a", "b", "c"];
-  const original = ["a", "b", "c"];
+// Then it should remove duplicates and not modify the original array
+test("given an array with strings, it should remove duplicates and preserve the original array", () => {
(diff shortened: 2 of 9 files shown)


Sprint-1/fix/median.js
```diff
@@ -6,9 +6,39 @@
...
+  // create a variable called numbers and store an empty array.
+  let numbers = [];
+
+  // loop over each item in list.
+  for (const item of list) {
+    // if item is a number, push it into the numbers array.
+    if (typeof item === "number") {
+      numbers.push(item);
+    }
+  }
```
cjyuan: - Could consider using array's `.filter()` method to simplify the code on lines 15-23.

- Do you plan to consider `-Infinity`, `Infinity`, and `NaN` in the median calculation (and also in the functions in `implement/max.js` and `implement/sum.js`)?

Sprint-1/implement/dedupe.test.js
cjyuan: There are two more tests to be implemented on this file.

Sprint-1/implement/max.test.js
```diff
@@ -16,28 +16,54 @@ const findMax = require("./max.js");
...
+  expect(findMax([3, "dogs", 1, "cat", null])).toBe(3);
+});
+
 // Given an array with only non-number values
 // When passed to the max function
 // Then it should return the least surprising value given how it behaves for all other inputs
+
+test("given an array with only non-number values, returns -Infinity", () => {
+  expect(findMax(["a", null, true, undefined])).toBe(-Infinity);
+});
```
cjyuan, line 69: When a string representing a valid numeric literal (for example, `"300"`) is compared to a number, 

libanj0161: Added tests for numeric string value.

Sprint-1/implement/dedupe.test.js
libanj0161: Added two tests.

Sprint-1/fix/median.js
libanj0161: used .filter to simplify code, and isFinite.

Sprint-1/implement/dedupe.test.js
```diff
@@ -16,12 +16,20 @@ E.g. dedupe([1, 2, 1]) target output: [1, 2]
...
+test("given an empty array, it returns an empty array", () => {
+  expect(dedupe([])).toEqual([]);
+});
 
 // Given an array with no duplicates
 // When passed to the dedupe function
 // Then it should return a copy of the original array
+test("given an array with no duplicates, it should return a copy of the original array", () => {
+  expect(dedupe(["a", "b", "c"])).toEqual(["a", "b", "c"]);
+});
```
cjyuan, line 33: Your function implementation is correct. However, this test could be improved to better ensure
that any future changes continue to align with the expected behavior described on line 25:

Sprint-1/implement/max.test.js
```diff
@@ -16,28 +16,68 @@ const findMax = require("./max.js");
...
+// When passed to the max function
+// Then it should return the correct total
+test("given an array including Nan return correct total", () => {
+  expect(findMax([11, NaN, 5])).toBe(11);
+});
+
+// Given an array with numeric string value
+// When passed to the max function
+// Then it should ignore the numeric string and return correct total
+test("given an array with numeric string value, it should ignore the numeric string and return correct total", () => {
```
cjyuan: The description of the expected return value are not quite correct.

Sprint-1/implement/dedupe.test.js
libanj0161: Added additional check expect(result).not.toBe(input) which verifies that the returned array is a different array in memory, ensuring it’s a true copy rather than the original.

Sprint-1/implement/max.test.js
libanj0161: Updated test description

Sprint-1/implement/dedupe.test.js
```diff
@@ -24,7 +24,11 @@ test("given an empty array, it returns an empty array", () => {
 // When passed to the dedupe function
 // Then it should return a copy of the original array
 test("given an array with no duplicates, it should return a copy of the original array", () => {
-  expect(dedupe(["a", "b", "c"])).toEqual(["a", "b", "c"]);
+  const input = ["a", "b", "c"];
+  const result = dedupe(input);
+
+  expect(result).toEqual(input);
+  expect(result).not.toBe(input);
```
cjyuan, line 32: There is a chance that, even though `result` has incorrect elements (for example, `[]`), 

libanj0161: Added .toHaveLength to ensure the returned array has the same number of elements as the original.

cjyuan: This is better but it is still not 100% bullet proof. What if result is `["a", "b", "x"]`?

cjyuan: Can you update the test to ensure `expected` has the expected elements?
And optionally add another test to ensure `input` remains unchanged after the function call.

libanj0161: From my research .toEqual already ensures the returned array has the exact expected elements. I also added an additional test to verify that calling the function does not change the original array.

Sprint-1/implement/dedupe.test.js
```diff
@@ -32,6 +32,18 @@ test("given an array with no duplicates, it should return a copy of the original
...
+// Given an array of strings
+// When passed to the dedupe function
+// Then it should not modify the original array
+test("given an array with strings, it should not modify the original array", () => {
+  const input = ["a", "b", "c"];
+  const original = ["a", "b", "c"];
+
+  dedupe(input);
+
+  expect(input).toEqual(original);
```
cjyuan, line 45: You could combine this with the previous test.

cjyuan (commented):
Changes look good.

Sprint-1/implement/dedupe.test.js
libanj0161: I have now, thanks

illicitonion:

PR #1019, opened 2026-03-16
Birmingham | ITP-Jan | Roman Sanaye | Sprint 2 | Exercises

RomanSanaye:
## Learners, PR Template

Self checklist

## Changelist

- Completed all exercises in this sprint, including the following folders:
  - **Debugging**: fixed errors and improved code logic.
  - **Implement**: implemented required functions and features as instructed.
  - **Interpret**: interpreted given code snippets and exercises.
  - **Stretch**: completed advanced challenges, including the `totalTill` function.
- Added Jest tests for key functions where required, including the `totalTill` function.
- Ensured all functions are beginner-friendly, tested, and formatted according to the style guide.

## Questions

No questions at this time.

### DIFF
--- a/Sprint-2/debug/recipe.js
+++ b/Sprint-2/debug/recipe.js
@@ -1,8 +1,10 @@
 // Predict and explain first...
+// ==> this code is not correct as the ingredients act as string not an object key therefore we wont have the value.
 
 // This program should log out the title, how many it serves and the ingredients.
 // Each ingredient should be logged on a new line
 // How can you fix it?
+// ==> to fix the issue and print each ingredients in new line we need to loop through the array inside our recipe object.
 
 const recipe = {
   title: "bruschetta",
@@ -11,5 +13,7 @@ const recipe = {
 };
 
 console.log(`${recipe.title} serves ${recipe.serves}
-  ingredients:
-${recipe}`);
+ingredients:`);
+for (let fruits of recipe.ingredients) {
+  console.log(fruits);
+}
--- a/Sprint-2/debug/recipe.js
+++ b/Sprint-2/debug/recipe.js
@@ -13,7 +13,5 @@ const recipe = {
 };
 
 console.log(`${recipe.title} serves ${recipe.serves}
-ingredients:`);
-for (let fruits of recipe.ingredients) {
-  console.log(fruits);
-}
+ingredients: ${recipe.ingredients.join("\n")}`);
+// this approach looks clean and advanced
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -58,9 +58,9 @@ test("It returns false when the object does NOT have the specified property", ()
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
-test("It return false when an array is passed instead of object", () => {
-  let object = [1, 2, 3];
-  let key = "name";
+test("It returns false even if array contains the key (arrays are invalid)", () => {
+  const object = [1, 2, 3];
+  const key = 0;
   expect(contains(object, key)).toEqual(false);
 });
 
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -63,3 +63,35 @@ test("It return false when an array is passed instead of object", () => {
   let key = "name";
   expect(contains(object, key)).toEqual(false);
 });
+
+// New testes after receiving feedback ==>
+
+test("It return false when null is passed instead of object", () => {
+  let object = null;
+  let key = "name";
+  expect(contains(object, key)).toEqual(false);
+});
+
+test("It return false when undefined is passed instead of object", () => {
+  let object = undefined;
+  let key = "name";
+  expect(contains(object, key)).toEqual(false);
+});
+
+test("It return false when a number is passed instead of object", () => {
+  let object = 123;
+  let key = "name";
+  expect(contains(object, key)).toEqual(false);
+});
+
+test("It return false when a string is passed instead of object", () => {
+  let object = "hello";
+  let key = "name";
+  expect(contains(object, key)).toEqual(false);
+});
+
+test("It return false when a boolean is passed instead of object", () => {
+  let object = true;
+  let key = "name";
+  expect(contains(object, key)).toEqual(false);
+});
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -15,21 +15,51 @@ as the object doesn't contains a key of 'c'
 
 // Given a contains function
 // When passed an object and a property name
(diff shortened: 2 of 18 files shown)


Sprint-2/debug/recipe.js
```diff
@@ -11,5 +13,7 @@ const recipe = {
 };
 
 console.log(`${recipe.title} serves ${recipe.serves}
-  ingredients:
-${recipe}`);
+ingredients:`);
+for (let fruits of recipe.ingredients) {
+  console.log(fruits);
+}
```
cjyuan: This works.

Here's another approach you can consider:
Since ingredient values are separated by '\n' in the output, we could also use `Array.prototype.join()` to construct the equivalent string and then output the resulting string.

Sprint-2/implement/contains.test.js
```diff
@@ -15,21 +15,51 @@ as the object doesn't contains a key of 'c'
...
+});
 
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+test("It return false when an array is passed instead of object", () => {
+  let object = [1, 2, 3];
+  let key = "name";
+  expect(contains(object, key)).toEqual(false);
+});
```
cjyuan, line 65: Array is a kind of object in JS, and "0", "1", "2" are keys of `[1, 2, 3]`. So it is better to express the test as `expect(contains([1, 2, 3], "1")).toBe(false);` to ensure the function can truly reject array.

You can also test other types of invalid parameters (e. g., number, `undefined`, `null`, string, boolean).

Note: Currently your function cannot yet properly handle invalid parameters.

Sprint-2/implement/lookup.js
```diff
@@ -1,5 +1,15 @@
-function createLookup() {
-  // implementation here
+// start your function;
+function createLookup(input) {
+  let output = {};
+  for (let pair of input) {
+    let key = pair[0];
+    let value = pair[1];
+    output[key] = value;
```
cjyuan: `output` and `pair` are not going to be reassigned a value, so a better practice is to declare them using `const`.

We can still modify the array or object assigned to a constant in JS, we just cannot reassign another value to the constant.

Sprint-2/implement/querystring.js
```diff
@@ -1,16 +1,24 @@
...
-
-  for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
-    queryParams[key] = value;
+  // URLSearchParams is a built-in JavaScript object that can read query strings.
+  const params = new URLSearchParams(queryString); // params stores the query string internally as key-value pairs.
+  for (let pair of params) {
+    let key = pair[0];
+    let value = pair[1];
+    queryParams[key] = value.replace(/ /g, "+");
```
cjyuan, line 12: This approach would convert a value like `"a%20%20b"` to `"a++b"` instead of `"a  b"`. 
No change requierd.
I think the given test case is incorrect because a `+` in a real query string should be treated as space.

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,15 @@
...
+  let output = {};
+  if (!Array.isArray(arr)) {
+    throw new Error("Invalid input");
+  } else if (arr.length === 0) {
+    return output;
+  }
+  for (let item of arr) {
+    let count = arr.filter((x) => x === item).length;
+    output[item] = count;
+  }
```
cjyuan: This code work but it is not very efficient. This is because for every item (including items that have already been counted), the expression on line 9 needs to process the whole array (i.e. all elements in the array).

An approach similar to what you used in `stretch/count-words.js` would be more efficient.

Sprint-2/stretch/count-words.js
```diff
@@ -26,3 +28,16 @@
...
+  let cleanStr = str.split(" ");
+  let objWords = {};
+  for (let word of cleanStr) {
+    if (word !== "") continue;
+    if (objWords[word]) {
+      objWords[word]++;
+    } else {
+      objWords[word] = 1;
+    }
+  }
```
cjyuan: Can you check if your function returns what you expect in the following function call?

```
```

Sprint-2/stretch/mode.js
```diff
@@ -9,28 +9,29 @@
 // into smaller functions using the stages above
 
 function calculateMode(list) {
-  // track frequency of each value
```
cjyuan, line 12: The purpose of this exercise is to practice refactoring code in a "large" function into smaller functions.
You were supposed to separate the original code into two functions:
Function 1 -- when given an array, returns the frequency of the element in a Map object.
Function 2 -- when given a Map object representing the frequency of of elements, return the mode.

Then, rewrite the original function as a composition of these two functions.

Sprint-2/debug/recipe.js
RomanSanaye: Hi @cjyuan ,
Thanks for your time and useful feedbacks.
I have updated the function and addressed your point.

Sprint-2/implement/contains.test.js
RomanSanaye: I have updated the function and tests are added as you addressed.

Sprint-2/implement/lookup.js
RomanSanaye: let  was replaced with const, and it is a safer option.

Sprint-2/implement/tally.js
RomanSanaye: tally function been updated and refactored.

Sprint-2/stretch/count-words.js
RomanSanaye: I have updated the function and used Object.create(null) which has no prototype or not having inherited properties.

Sprint-2/stretch/mode.js
RomanSanaye: I have updated and refactored the function.

Sprint-2/implement/contains.test.js
```diff
@@ -15,21 +15,83 @@ as the object doesn't contains a key of 'c'
...
```
cjyuan, line 65: In practice, we prepare tests not just to check our own implementation, but also to guard against future modification. 

Currently, this incorrectly implemented function could also pass this test:
```
function contains(object, key) {
  if (object === null || typeof object !== "object") {
    return false;
  }
  return Object.prototype.hasOwnProperty.call(object, key);
}
```

cjyuan (commented):
Other changes look good.

Sprint-2/implement/contains.test.js
RomanSanaye: Thanks for the feedback! I’ve added a test to ensure arrays are rejected even when the key exists, so incorrect implementations will now fail.

cjyuan:
All good now. Well done.

illicitonion:

PR #1020, opened 2026-03-16
Glasgow | Jan-26 | Elisabeth Matulian | Sprint 3 | Alarm clock app

Elisabeth-Matulian:
## Learners, PR Template

Self checklist

## Changelist

Updated the <title> tag in index.html

Implemented the setAlarm() function

Added a countdown timer with the title updated every second

Added clearInterval to pauseAlarm() to stop the timer

## Questions

N/Q

### DIFF
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -1,4 +1,42 @@
-function setAlarm() {}
+const timeRemaining = document.querySelector("#timeRemaining");
+
+let intervalId;
+
+
+
+function setAlarm() {
+  // I will use the function showTime twice, one for showing the time just after the user clicks the set button and another one for updating the time every second in countdown function
+  function showTime() {
+  // getting te mins and seconds from the inputValue in appropriate format
+      const minutes = Math.floor(inputValue / 60)
+        .toString()
+        .padStart(2, "0");
+      const seconds = (inputValue % 60).toString().padStart(2, "0");
+      //Gathering everything together and outputing remaining time
+      timeRemaining.innerText = `Time Remaining: ${minutes}:${seconds}`;
+};
+  // I added the next line in case if we want to reset the time for alarm
+  clearInterval(intervalId);
+  // getting the input value in number format
+  let inputValue = Number(document.querySelector("#alarmSet").value);
+  showTime();
+  intervalId = setInterval(countdown, 1000);
+
+  function countdown() {
+    if (inputValue > 0) {
+      //decrease the entered number by 1
+      inputValue--;
+      showTime();
+    }
+    // When the inputValue is 0, play the alarm and clear interval so that the function stops initialising 
+    else {
+      playAlarm();
+      clearInterval(intervalId);
+    }
+  }
+}
+
+
 
 // DO NOT EDIT BELOW HERE
 
@@ -20,6 +58,8 @@ function playAlarm() {
 
 function pauseAlarm() {
   audio.pause();
+  //I know it's said to not edit this code, I wonder if it's ok to add this line below in order to pause the time before the time runs out. 
+  clearInterval(intervalId);
 }
 
 window.onload = setup;
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -1,42 +1,44 @@
 const timeRemaining = document.querySelector("#timeRemaining");
 
-let intervalId;
-
-
+let intervalId; //declared here so that I can use it in line-17, before asigning on line 21. 
 
 function setAlarm() {
   // I will use the function showTime twice, one for showing the time just after the user clicks the set button and another one for updating the time every second in countdown function
-  function showTime() {
-  // getting te mins and seconds from the inputValue in appropriate format
+  function showTime(inputValue) {
+    if (inputValue >= 0) {
+  // getting the mins & sec from the inputValue(line19) in appropriate format
       const minutes = Math.floor(inputValue / 60)
         .toString()
         .padStart(2, "0");
       const seconds = (inputValue % 60).toString().padStart(2, "0");
       //Gathering everything together and outputing remaining time
       timeRemaining.innerText = `Time Remaining: ${minutes}:${seconds}`;
-};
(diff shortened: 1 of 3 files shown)


Sprint-3/alarmclock/alarmclock.js
```diff
@@ -20,6 +58,8 @@ function playAlarm() {
 
 function pauseAlarm() {
   audio.pause();
+  //I know it's said to not edit this code, I wonder if it's ok to add this line below in order to pause the time before the time runs out. 
+  clearInterval(intervalId);
 }
```
hkavalikas, line 69: Excellent catch, and the approach is correct, as a bonus task, though, how can we make this change and add the `clearInterval` call, without editing anything below the `// DO NOT EDIT BELOW HERE` line?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,42 @@
...
+
+
+
+function setAlarm() {
+  // I will use the function showTime twice, one for showing the time just after the user clicks the set button and another one for updating the time every second in countdown function
+  function showTime() {
+  // getting te mins and seconds from the inputValue in appropriate format
+      const minutes = Math.floor(inputValue / 60)
+        .toString()
+        .padStart(2, "0");
```
hkavalikas: This works even though `inputValue` is declared after the function declaration, due to the variable `inputValue` getting accessed at call time instead of definition time. If we wanted to simplify this ever so slightly, how can we get the variables that will be used in the function before the actual function declaration?

Sprint-3/alarmclock/alarmclock.js
hkavalikas, line 1: Let's handle a few more cases here. What will happen if we add -5 and 0 to the countdown?

hkavalikas (changes requested):

hkavalikas:
Finally, let's try to add a few more details to the Changelist. The purpose of the changelist is to mention individual updates/additions/changes made for this PR, not so much to add a headline (like the title) of the changes.

Sprint-3/alarmclock/alarmclock.js
Elisabeth-Matulian: Could you give me a hint?

Sprint-3/alarmclock/alarmclock.js
Elisabeth-Matulian: I added inputValue as a parameter for showTime function

Sprint-3/alarmclock/alarmclock.js
Elisabeth-Matulian: updated

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,44 @@
-function setAlarm() {}
+const timeRemaining = document.querySelector("#timeRemaining");
+
+let intervalId; //declared here so that I can use it in line-17, before asigning on line 21. 
+
+function setAlarm() {
+  // I will use the function showTime twice, one for showing the time just after the user clicks the set button and another one for updating the time every second in countdown function
+  function showTime(inputValue) {
+    if (inputValue >= 0) {
```
hkavalikas: This will cause the alarm to play 2 ticks later, how can we fix that?

Sprint-3/alarmclock/alarmclock.js
hkavalikas: Sure thing, what if I told you that you can have multiple event listeners for the same element?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,44 @@
...
+  // getting the mins & sec from the inputValue(line19) in appropriate format
+      const minutes = Math.floor(inputValue / 60)
+        .toString()
+        .padStart(2, "0");
+      const seconds = (inputValue % 60).toString().padStart(2, "0");
+      //Gathering everything together and outputing remaining time
+      timeRemaining.innerText = `Time Remaining: ${minutes}:${seconds}`;
+}}
```
hkavalikas: We could add each `}` to a separate line but not too important.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,44 @@
...
+  //the next line in case if we want to reset the time for alarm
+  clearInterval(intervalId);
+  // getting user's input value in number format
+  let inputValue = Number(document.querySelector("#alarmSet").value);
```
hkavalikas: We are querying `document.querySelector("#alarmSet")` three times, how can we do it once and reuse it?

hkavalikas (changes requested):

Sprint-3/alarmclock/alarmclock.js
Elisabeth-Matulian: I created a variable to store it

Sprint-3/alarmclock/alarmclock.js
Elisabeth-Matulian: formating used

Sprint-3/alarmclock/alarmclock.js
Elisabeth-Matulian: i removed if condition

Sprint-3/alarmclock/alarmclock.js
Elisabeth-Matulian: I still couldn't figure out how to implement it :face_with_spiral_eyes:

hkavalikas: You have actually done it, but the code is commented out on lines 43-45. You can indeed register multiple event listeners to the same element as you did; they won't override each other. For what is worth, adding them in the pause alarm did make sense too, but I hope it was an interesting extra step. Let's comment it back it and I will approve :)

Elisabeth-Matulian: Oh, so that’s how it works))) 

I didn’t think it would work, I just left it as a comment so you could see the way I think and point me in the right direction. That’s surely something I’ll remember))

hkavalikas:
Awesome work

illicitonion:

PR #1021, opened 2026-03-16
Birmingham |  26 JAN ITP | Merve Reis | Sprint 1 | Data Groups

mervereis:
## Learners, PR Template

Self checklist

## Changelist

Functions created as requested and test cases wroten for

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -24,7 +24,7 @@ function calculateMedian(list) {
     return null;
   }
 
-  const sorted = [...numbers].sort((a, b) => a - b);
+  const sorted = numbers.sort((a, b) => a - b);
 
   const middle = Math.floor(sorted.length / 2);
 
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -24,7 +24,6 @@ function calculateMedian(list) {
     return null;
   }
 
-  // Copy before sort (extra safety, though filter already creates new array)
   const sorted = [...numbers].sort((a, b) => a - b);
 
   const middle = Math.floor(sorted.length / 2);
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -5,10 +5,34 @@
 // Hint: Please consider scenarios when 'list' doesn't have numbers (the function is expected to return null)
 // or 'list' has mixed values (the function is expected to sort only numbers).
 
+// function calculateMedian(list) {
+//   const middleIndex = Math.floor(list.length / 2);
+//   const median = list.splice(middleIndex, 1)[0];
+//   return median;
+// }
+
+// module.exports = calculateMedian;
+
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  if (!Array.isArray(list)) {
+    return null;
+  }
+
+  const numbers = list.filter((item) => typeof item === "number");
+
+  if (numbers.length === 0) {
+    return null;
+  }
+
+  numbers.sort((a, b) => a - b);
+
+  const middle = Math.floor(numbers.length / 2);
+
+  if (numbers.length % 2 === 0) {
+    return (numbers[middle - 1] + numbers[middle]) / 2;
+  }
+
+  return numbers[middle];
 }
 
 module.exports = calculateMedian;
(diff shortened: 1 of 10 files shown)


Sprint-1/fix/median.js
```diff
@@ -5,10 +5,34 @@
...
+
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  if (!Array.isArray(list)) {
+    return null;
+  }
+
+  const numbers = list.filter((item) => typeof item === "number");
```
cjyuan: Do you plan to allow `-Infinity`, `Infinity`, and `NaN` in the median calculation (and also in the functions in `implement/max.js` and `implement/sum.js`)?

Sprint-1/implement/max.test.js
```diff
@@ -1,43 +1,52 @@
...
+    expect(findMax(["hey", 10, "hi", 60, 10])).toBe(60);
+  });
+
+  // Given an array with only non-number values
+  // When passed to the max function
+  // Then it should return the least surprising value given how it behaves for all other inputs
+  test("returns -Infinity when array has only non-number values", () => {
+    expect(findMax(["a", "b", "c"])).toBe(-Infinity);
+  });
+});
```

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,12 @@
 function sum(elements) {
+  const numbers = elements.filter((el) => typeof el === "number");
+
+  if (numbers.length === 0) return 0;
+
+  if (numbers.length === 1) return numbers[0];
```
cjyuan: Why introduce if statements for these two cases? The code on lines 8-9 can also handle these two cases. That is, even without lines 4 and 6, the function would still behave the same.

Sprint-1/implement/sum.test.js
```diff
@@ -1,36 +1,45 @@
...
+  test("sums negative numbers correctly", () => {
+    expect(sum([-5, -10, 15])).toBe(0);
+  });
+
+  // Given an array with decimal/float numbers
+  // When passed to the sum function
+  // Then it should return the correct total sum
+  test("sums decimal numbers correctly", () => {
+    expect(sum([1.5, 2.5, 3.5])).toBe(7.5);
+  });
```
cjyuan, line 30: Decimal numbers in most programming languages (including JS) are internally represented in "floating point number" format. Floating point arithmetic is not exact. For example, the result of `46.5678 - 46 === 0.5678` is false because `46.5678 - 46` only yield a value that is very close to `0.5678`. Even changing the order in which the program add/subtract numbers can yield different values.

So the following could happen
```javascript

```

Suggestion: Look up 

mervereis: @cjyuan I updated the test cases and added new one for increase coverage on floating numbers

Sprint-1/implement/sum.js
mervereis: @cjyuan I updated the function

Sprint-1/implement/max.test.js
mervereis: @cjyuan I updated the test cases and added new one for increase coverage on string numbers conversation and comparison

Sprint-1/fix/median.js
mervereis: @cjyuan I updated the test cases and added new one for increase coverage on Infinity and NaN controls

Sprint-1/fix/median.js
```diff
@@ -5,10 +5,35 @@
...
+  }
+
+  const numbers = list.filter((item) => Number.isFinite(item));
+
+  if (numbers.length === 0) {
+    return null;
+  }
+
+  // Copy before sort (extra safety, though filter already creates new array)
+  const sorted = [...numbers].sort((a, b) => a - b);
```
cjyuan: Why do you need this "extra safety" to make a copy of numbers before sorting? Safety from what?

cjyuan (commented):
Other changes look good.

Sprint-1/fix/median.js
mervereis: @cjyuan removed unnecessary comments

cjyuan: Why make a copy of `numbers` before sorting?

mervereis: @cjyuan I reverted to spreading numbers before sorting, now just using numbers directly instead of creating a copy of it.

cjyuan:
Changes look good.

illicitonion:

PR #1022, opened 2026-03-16
London | 26-ITP-January | Laura C | Sprint 3 | Alarm Clock

lauracs24:
### Self checklist

* [x] I have titled my PR with Region | Cohort | FirstName LastName | Sprint | Assignment Title
* [x] My changes meet the requirements of the task
* [x] I have tested my changes
* [x] My changes follow the style guide

### Implement

* Updated the page title in `index.html` to **Alarm clock app**
* Implemented the alarm countdown logic in `alarmclock.js`
* Updated the countdown every second using a timer
* Played the alarm sound when the timer reached `00:00

### Testing

* Ran the provided tests using `npm test`
* Confirmed that all tests pass

### DIFF
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -1,4 +1,35 @@
-function setAlarm() {}
+let secondsRemaining = 0;
+let timer = null;
+
+function setAlarm() {
+  const input = document.getElementById("alarmSet");
+  const heading = document.getElementById("timeRemaining");
+
+  secondsRemaining = Number(input.value);
+
+  updateHeading(heading, secondsRemaining);
+
+  timer = setInterval(() => {
+    secondsRemaining -= 1;
+
+    updateHeading(heading, secondsRemaining);
+
+    if (secondsRemaining === 0) {
+      clearInterval(timer);
+      playAlarm();
+    }
+  }, 1000);
+}
+
+function updateHeading(heading, totalSeconds) {
+  const minutes = Math.floor(totalSeconds / 60);
+  const seconds = totalSeconds % 60;
+
+  const formattedMinutes = String(minutes).padStart(2, "0");
+  const formattedSeconds = String(seconds).padStart(2, "0");
+
+  heading.innerText = `Time Remaining: ${formattedMinutes}:${formattedSeconds}`;
+}
 
 // DO NOT EDIT BELOW HERE
 
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -7,6 +7,17 @@ function setAlarm() {
 
   secondsRemaining = Number(input.value);
 
+  // Handle invalid values (0 or negative)
+  if (secondsRemaining <= 0) {
+    updateHeading(heading, 0);
+    return;
+  }
+
+  // Prevent multiple timers running
+  if (timer !== null) {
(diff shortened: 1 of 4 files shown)


Sprint-3/alarmclock/alarmclock.js
hkavalikas, line 1: What will happen if we add -5 to the timer, or 0?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,35 @@
-function setAlarm() {}
+let secondsRemaining = 0;
+let timer = null;
+
+function setAlarm() {
```
hkavalikas, line 4: What will happen if we add a value like 20 and click on "Set Alarm" 5 times? How can we improve that?

Sprint-3/alarmclock/alarmclock.js
hkavalikas, line 1: Completely optional, but if you want a bonus task, how can we stop the timer if the user clicks on "Stop Alarm"? e.g. if we count from 5 -> then on 3 we click on "Stop alarm", it should freeze the timer.

hkavalikas (changes requested):

Sprint-3/alarmclock/alarmclock.js
lauracs24: Hmm I see, I’ve updated the implementation to handle invalid inputs (0 and negative values), prevent multiple timers when clicking "Set Alarm multiple times, and improved the Stop Alarm button so it clears the timer as well as stopping the sound. Hopefully that works!

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -18,8 +61,14 @@ function playAlarm() {
...
   audio.pause();
+
+  if (timer !== null) {
+    clearInterval(timer);
+    timer = null;
+  }
 }
 
-window.onload = setup;
+window.onload = setup;
```
hkavalikas: Can you think of a way of doing the pause alarm timer top without editing below the `DO NOT EDIT BELOW HERE` line?

Sprint-3/alarmclock/alarmclock.js
hkavalikas, line 1: What will happen if we add 1.5 in the timer? How can we fix it?

hkavalikas (changes requested):

lauracs24:
Hi @hkavalikas. Can you please review my request?

Sprint-3/alarmclock/alarmclock.js
lauracs24: I moved the timer-stopping logic above the DO NOT EDIT BELOW HERE line by creating a separate stopTimer() function and attaching it to the Stop Alarm button, while leaving the existing pauseAlarm() unchanged. Feel free to add the final label if theres nothing mandatory left for me to do 😅

hkavalikas:
You might have missed this:
https://github.com/CodeYourFuture/Module-Data-Groups/pull/1022#discussion_r2994630131

Sprint-3/alarmclock/alarmclock.js
lauracs24: Sorry I missed this!

I’ve now updated the implementation to ensure the timer only uses whole seconds by converting the input to an integer and added validation to handle invalid inputs @hkavalikas

hkavalikas:
Awesome work

illicitonion:

PR #1023, opened 2026-03-17
London |ITP-Jan-2016 | Ping Wang | Sprint 2 |Coursework

pathywang:
Self checklist

## Changelist

i did coursework with node or npm jest to check my work to make sure that my code works as expected after i completed my prep


Sprint-2/debug/recipe.js
```diff
@@ -11,5 +12,8 @@ const recipe = {
 };
 
 console.log(`${recipe.title} serves ${recipe.serves}
-  ingredients:
-${recipe}`);
+ingredients:`);
+
+for (const ingredient of recipe.ingredients) {
+  console.log(ingredient);
+}
```
cjyuan: This work.

Alternate approach to explore:
Since ingredient values are separated by '\n' in the output, we could also use `Array.prototype.join()` to construct the equivalent string and then output the resulting string.

Sprint-2/implement/contains.test.js
```diff
@@ -20,16 +20,30 @@ as the object doesn't contains a key of 'c'
...
+test.todo("contains no nice property  in object return false")
+
+
 
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+object={name:'jin', age:13}
+console.log (contains(object, '[1,2,5]'))
+test.todo("contains no array property  in object return false")
```
cjyuan: `test.todo(" ... ")` is just a placeholder to serve as a reminder for "what needs to be done". You were supposed to implement the corresponding Jest tests described in the comments on this file.

Can you implement all the tests in all the `.test.js` files?

Sprint-2/implement/lookup.js
```diff
@@ -1,5 +1,12 @@
-function createLookup() {
-  // implementation here
+function createLookup(countryCurrencyPairs) {
+  const countryCodeCurrency={}
+  for (const[country, currency] of countryCurrencyPairs){
+   countryCodeCurrency[country]=currency
+  }
+
+  return countryCodeCurrency
 }
```
cjyuan, line 9: The spacing around the operators is not quite consistent.

as recommended in 
?

Sprint-2/implement/lookup.test.js
```diff
@@ -33,3 +35,12 @@ It should return:
...
 */
+
+// create single country currency in countryCurrencyPairs
+const result1=createLookup([['GB','GBP']])
+console.log(result1)
+test.todo('creates a single county code currency ')
+
+//given an empty country currency pair/array
+console.log(createLookup([[]]))
+test.todo('return an empty object with empty pair given')
```
cjyuan: - Should also implement these tests

- Better practice to end each statement by a semicolon (`;`) to reduce chance of errors.

Sprint-2/implement/querystring.js
```diff
@@ -1,16 +1,24 @@
...
+
     queryParams[key] = value;
   }
 
   return queryParams;
 }
 
+console.log(parseQueryString("color=blue&quality=good"))
+console.log(parseQueryString("equation=x=y+1"))
+console.log(parseQueryString(""))
```
cjyuan, line 28: For each of the following function calls, does your function return the value you expect?
```
parseQueryString("a=b&=&c=d")
parseQueryString("a=b&&c=d")
```
---
Note: 
- (Implementing this is optional) In real query string, both `key` and `value` are ***percent-encoded*** or ***URL encoded*** in the URL.

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,27 @@
...
+  for(item of array){
+    if (count[item]){
+        count[item] += 1
+    }
+    else{
+        count[item] = 1
+    }
+  }
+  return count
+  }
```
cjyuan: - Can you improve the indentation?

- Does the following function call returns the value you expect?
```
```

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,28 @@
...
+  for (const item of array) {
+    if (count[item]) {
+      count[item] += 1;
+    } else {
+      count[item] = 1;
+    }
+  }
+
+  return count;
+}
```
cjyuan: Code is not yet properly indented. Can you fix the identation?

cjyuan (commented):
Please note that in CYF courses, the recommended way to inform the reviewer of your changes is to do both of the following:
  - **Reply to their feedback.**
    - In the responses, clarify how each piece of feedback was addressed to demonstrate that you've carefully reviewed the suggestions.
      - You may find the suggestions in this [PR Guide](https://programming.codeyourfuture.io/guides/reviewing/trainee-pr-guide/#addressing-reviewer-feedback) useful.
    - Your response may trigger a notification (depending on the reviewer's settings), helping ensure they’re aware of the updates you’ve made. 
  - **Replace the "Reviewed" label by a "Needs review" label** (which you have done -- great!)
    - Without this label, the reviewer would not know if your changes is ready to be reviewed.

Sprint-2/implement/tally.js
pathywang: sorry sometimes i forgot about indention syntax because i pay more attention on node running, the rule is about 2 standard space for indentation and allign the parallel function

Sprint-2/debug/recipe.js
pathywang: yes, good suggestion,thanks. while recipe.ingredients.join("\n") turns the array into a single string and each ingredient is separated by a newline (\n) which makes code shorter and cleaner.

Sprint-2/implement/tally.js
pathywang: Thank you for the suggestion. It is the first time i hear the empty object with no inherited properties. i went to AI help who explained to me about the advantage of no inherited properties. i am trying to practice with my code from now on.

Sprint-2/implement/tally.js
pathywang: i would put comment on after i correct my code according PR guide which give me another chance to review correction. What is more, i know what is the reason for correction later on.

cjyuan: If you have successfully enabled "Format on save" and assign "Prettier" as your JS code formatter on VSCode, you won't have to worry about indentation and code format anymore.

cjyuan:
Changes look good. Well done.

pathywang:
Thank you for your suggestion and correction, really appreciate it!

kind regards

ping

illicitonion:

PR #1024, opened 2026-03-17
 London |ITP- Jan-2026| Ping Wang|Sprint 3 |Alarm clock

pathywang:
Self checklist

## Changelist

i do feel hard at the beginning because there is communication between index.html and JavaScript.js code which i never had this experience in previous study.  I need get element from html to JavaScript then use script to make html function properly with JS code. How do i do on my own without any help, it is quite overwhelming  for me even now. My understanding is like proper project in real life which seemed far away from my current study but it is not. i need get on with it. To be honest, i am still not confident at the moment. I think i will get better with constant review and more practice

.Study prep twice, any questions to ask chatGpt mainly 
. Read readme.md carefully in order to understand requirement for this PR 
. Create my own branch Sprint-3/alarm clock from main 
. Update title alarm clock App in index.html then commit to github
. complete code on alarmclock.js file then commit change to github
. did test on alarmclock.test.js file

### DIFF
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -1,4 +1,34 @@
-function setAlarm() {}
+let countdownInterval = null;
+
+function setAlarm() {
+  const input = document.getElementById("alarmSet");
+  let totalSeconds = parseInt(input.value);
+
+
+  if (countdownInterval) clearInterval(countdownInterval);
+
+  const heading = document.getElementById("timeRemaining");
+
+  function updateHeading() {
+    const minutes = Math.floor(totalSeconds / 60);
+    const seconds = totalSeconds % 60;
+    heading.textContent = `Time Remaining: ${String(minutes).padStart(2, "0")}:${String(seconds).padStart(2, "0")}`;
+  }
+
+  updateHeading();
+
+  countdownInterval = setInterval(() => {
+    totalSeconds--;
+
+    if (totalSeconds <= 0) {
+      clearInterval(countdownInterval);
+      playAlarm();
+      totalSeconds = 0; 
+    }
+
+    updateHeading();
+  }, 1000);
+}
 
 // DO NOT EDIT BELOW HERE
 
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -2,12 +2,17 @@ let countdownInterval = null;
 
 function setAlarm() {
   const input = document.getElementById("alarmSet");
+  const heading = document.getElementById("timeRemaining");
+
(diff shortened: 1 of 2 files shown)


Sprint-3/alarmclock/alarmclock.js
hkavalikas, line 1: What will happen if we add -5 or not add anything to the timer?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,34 @@
...
+
+    if (totalSeconds <= 0) {
+      clearInterval(countdownInterval);
+      playAlarm();
+      totalSeconds = 0; 
+    }
+
+    updateHeading();
+  }, 1000);
+}
```
hkavalikas, line 41: The alarm will play before the seconds remaining are set to exactly 0, how can we improve that?

hkavalikas (changes requested):

hkavalikas:
Finally, let's try to add a few more details on the Changelist. The purpose of the changelist is to mention individual updates/additions/changes made for this PR, not so much to add a headline (like the title) of the changes.

Sprint-3/alarmclock/alarmclock.js
pathywang: I think a negative value will trigger the alarm right away (or nearly immediately) and empty input breaks the timer unless handled. so i correct my code if (isNaN(totalSeconds) || totalSeconds <= 0) {
    alert("Please enter a valid positive number.");
    return;
  } to make sure the code is running smoothly

Sprint-3/alarmclock/alarmclock.js
pathywang: I think the alarm is triggered before the UI clearly shows 00:00, so it feels early. I amend my code in order to update the UI before triggering the final action

hkavalikas:
I just want to follow up on the `Changelist` comment, and I will review the other changes a bit later. 

For the changelist, we are usually adding bullet points or some notes on the changes made in this PR, for example, I added X for Y, made updates X, used Y... and so on. Literally giving us a short overview of the changes made in this PR. 

But I just wanted to quickly acknowledge your current changelist/concerns. The first few times can definitely be overwhelming, so don't get discouraged, you will be repeating these steps so many times you won't even think about them in the future. The beginning is always hardest but you have done an excellent job 👏 .

hkavalikas:
PR looks great, only the changelog request is left to address. 
Ref: https://github.com/CodeYourFuture/Module-Data-Groups/pull/1024#issuecomment-4110742846

pathywang:
Thanks Harry for encouragement while i am a bit lack of confidence at the moment. What is more, i update changelist under your instruction.

illicitonion:

PR #1027, opened 2026-03-17
Birmingham | ITP Jan 2026 | Arunkumar Akilan | Sprint 2 | Data Groups

arunkumarakilan:
### Self checklist

---

### 🐞 Debug

- Fixed bugs in `address.js`
- Fixed bugs in `author.js`
- Fixed bugs in `recipe.js`
- Ensured all debug tests pass

### ⚙️ Implementation

- Implemented `tally` function to count occurrences in an array  
- Implemented `contains` function to check if object has a property  
- Implemented `createLookup` function to convert key-value pairs into an object  
- Implemented `parseQueryString` with edge case handling  
- Implemented `invert` function to swap keys and values  

### 🧪 Testing

- Added unit tests for all functions  
- Covered edge cases (empty input, invalid input, multiple values, etc.)  
- All tests are passing  

---

## Notes

- Used methods like `Object.keys`, `Object.entries`, and array iteration  
- Improved understanding of objects and key-value handling  

---

## Questions

No questions at the moment

### DIFF
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -11,6 +11,38 @@ E.g. contains({a: 1, b: 2}, 'c') // returns false
 as the object doesn't contains a key of 'c'
 */
 
+test("An empty object should return false",()=>{
+    const object = {};
+    const propertyName = "a";
+    expect(contains(object, propertyName)).toEqual(false);
+});
+  
+
+test("returns true when the object contains an existing property name", () => {
+  const object = { a: 1, b: 2, c: 3 };
+  const propertyName = "a";
+  expect(contains(object, propertyName)).toEqual(true);
+});
+test("returns false when the object not contains an existing property name", () => {
+  const object = { a: 1, b: 2, c: 3 };
+  const propertyName = "d";
+  expect(contains(object, propertyName)).toEqual(false);
+});
+test("Should throw an error for an invalid parameters", () => {
+  const object = [1,2,3,4,5,6,7];
+  const propertyName = "a";
+  expect(() => contains(object, propertyName)).toThrow();
+});
+test("Should throw an error for an invalid parameters", () => {
+  const object = null;
+  const propertyName = "a";
+  expect(() => contains(object, propertyName)).toThrow();
+});
+test("Should throw an error for an invalid parameters", () => {
+  const object = "hi";
+  const propertyName = "a";
+  expect(() => contains(object, propertyName)).toThrow();
+});
 // Acceptance criteria:
 
 // Given a contains function
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -11,12 +11,11 @@ E.g. contains({a: 1, b: 2}, 'c') // returns false
 as the object doesn't contains a key of 'c'
 */
 
-test("An empty object should return false",()=>{
-    const object = {};
-    const propertyName = "a";
-    expect(contains(object, propertyName)).toEqual(false);
+test("An empty object should return false", () => {
+  const object = {};
+  const propertyName = "a";
+  expect(contains(object, propertyName)).toEqual(false);
 });
-  
 
 test("returns true when the object contains an existing property name", () => {
   const object = { a: 1, b: 2, c: 3 };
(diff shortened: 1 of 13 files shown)


Sprint-2/implement/contains.test.js
```diff
@@ -11,6 +11,38 @@ E.g. contains({a: 1, b: 2}, 'c') // returns false
...
+test("returns false when the object not contains an existing property name", () => {
+  const object = { a: 1, b: 2, c: 3 };
+  const propertyName = "d";
+  expect(contains(object, propertyName)).toEqual(false);
+});
+test("Should throw an error for an invalid parameters", () => {
+  const object = [1,2,3,4,5,6,7];
+  const propertyName = "a";
+  expect(() => contains(object, propertyName)).toThrow();
+});
```
cjyuan, line 34: Array is also an object where it indexes are its keys.
`Object.hasOwn([1, 2, 3], "a")` evaluates to `false` because `"a"` is not a key (index) of the array.
However, `Object.hasOwn([1, 2, 3], "0")` evaluates to `true` because `"0"` is a key (index) of the array.

To test if the implemented function can correctly return `false` when the first argument is an array,
we should specify an actual key of the array.

Sprint-2/implement/querystring.js
```diff
@@ -5,8 +5,10 @@ function parseQueryString(queryString) {
...
   const keyValuePairs = queryString.split("&");
 
-  for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
+  for (const item of keyValuePairs) {
+    const pair = item.split("=");
+    const key = pair.shift();
+    const value = pair.join("=");
     queryParams[key] = value;
   }
```
cjyuan, line 14: For each of the following function calls, does your function return the value you expect?
```
parseQueryString("a=b&=&c=d")
parseQueryString("a=")
parseQueryString("=b")
parseQueryString("a=b&&c=d")
```

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,18 @@
-function tally() {}
+function tally(arr) {
+    const object = {};
```
```
```

Sprint-2/interpret/invert.js
```diff
@@ -9,21 +9,28 @@
 function invert(obj) {
   const invertedObj = {};
 
-  for (const [key, value] of Object.entries(obj)) {
-    invertedObj.key = value;
+  for (const [value, key] of Object.entries(obj)) {
+    invertedObj[key] = value;
   }
```
cjyuan: This work but in terms of readability,
- `[value, key] of ...` can mislead people to think, "the code incorrectly de-structure the key-value pair".
- Whereas, `invertedObj[value] = key` reads more natural like "the code is using `value` as key, and `key` as value".

cjyuan (commented):
Can you ensure all the code is consistently formatted?
If you had successfully enabled "Format on save" and specified "Prettier" as your default JS formatter, you would not have this issue.

Sprint-2/implement/contains.test.js
arunkumarakilan: Thanks for the clarification!

I understand that arrays are objects and that
Object.hasOwn([1, 2, 3], "0") returns true because "0" is a valid index.

However, in my implementation I explicitly treat arrays as invalid inputs:

if (typeof object !== "object" || Array.isArray(object) || object === null)

So the function throws an error before reaching Object.keys().
That’s why the test passes — it never evaluates array indexes like "0".

I’ll update either the test or the implementation depending on the expected behaviour

Sprint-2/interpret/invert.js
arunkumarakilan: I’ve updated the destructuring to [key, value] and used invertedObj[value] = key for better readability and consistency.

Sprint-2/implement/tally.js
arunkumarakilan: Thanks for pointing this out!

I understand the issue with inherited properties like toString.
I’ll update the implementation to use Object.create(null) so the object has no prototype and avoids this problem.

Sprint-2/implement/querystring.js
arunkumarakilan: My function currently includes empty keys for malformed query pairs. I can improve it by skipping invalid entries.

Sprint-2/implement/contains.test.js
cjyuan: It's not about your function, it is correctly implemented.

In practice, we prepare tests not just to check our implementation, but also to serve as a guard for future modification. In TDD, the tests will also serve as the spec that define how the function should be implemented.
That's why I keep emphasising the importance of preparing the tests correctly.

Sprint-2/implement/querystring.js
cjyuan: This one is good.

cjyuan:
Changes look good.

illicitonion:

PR #1029, opened 2026-03-17
Sheffield | 26-ITP-Jan | Karim M'hamdi | Sprint 1 | Module-Data-Groups

KKtech06:
## Learners, PR Template

Self checklist

## Questions

N/A

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -10,7 +10,7 @@ function calculateMedian(list) {
     return null;
   }
   const numbersOnly = list.filter(item => typeof item === 'number');
-  const sorted = [...numbersOnly].sort((a, b) => a - b);
+  const sorted = numbersOnly.sort((a, b) => a - b);
 
   if (sorted.length === 0) {
     return null;
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -6,9 +6,24 @@
 // or 'list' has mixed values (the function is expected to sort only numbers).
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  if (!Array.isArray(list)) {
+    return null;
+  }
+  const numbersOnly = list.filter(item => typeof item === 'number');
+  const sorted = [...numbersOnly].sort((a, b) => a - b);
+
+  if (sorted.length === 0) {
+    return null;
+  } else if (sorted.length % 2 === 0) {
+    const mid1 = sorted.length / 2 - 1;
+    const mid2 = sorted.length / 2;
+    const median = (sorted[mid1] + sorted[mid2]) / 2;
+    return median;
+  } else {
+    const middleIndex = Math.floor(sorted.length / 2);
+    const median = sorted[middleIndex];
+    return median;
+  }
 }
 
 module.exports = calculateMedian;
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -9,7 +9,7 @@ function calculateMedian(list) {
   if (!Array.isArray(list)) {
     return null;
   }
-  const numbersOnly = list.filter(item => typeof item === 'number');
+  const numbersOnly = list.filter((item) => Number.isFinite(item));
   const sorted = numbersOnly.sort((a, b) => a - b);
 
   if (sorted.length === 0) {
--- a/Sprint-1/implement/max.js
+++ b/Sprint-1/implement/max.js
@@ -1,12 +1,22 @@
-function findMax(elements){
-    let max = -Infinity;
-   for (let i = 0; i < elements.length; i++) {
-     if (typeof elements[i] === "number")
-       if (elements[i] > max) {
-         max = elements[i];
-       }
-   }
-    return max;
+function findMax(elements) {
+  let max = -Infinity;
+  let foundNumber = false;
+
+  for (let i = 0; i < elements.length; i++) {
+    if (Number.isFinite(elements[i])) {
+      foundNumber = true;
+
+      if (elements[i] > max) {
+        max = elements[i];
+      }
+    }
+  }
+
+  if (!foundNumber) {
+    return null;
+  }
+
+  return max;
 }
 
 module.exports = findMax;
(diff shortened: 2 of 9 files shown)


Sprint-1/fix/median.js
```diff
@@ -6,9 +6,24 @@
...
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  if (!Array.isArray(list)) {
+    return null;
+  }
+  const numbersOnly = list.filter(item => typeof item === 'number');
+  const sorted = [...numbersOnly].sort((a, b) => a - b);
```
cjyuan: - `-Infinity`, `Infinity`, and `NaN` are also values of type "number". Do you plan to include them in the calculation of median?

- On line 13, why make another copy of `numbersOnly` before sorting?

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,12 @@
...
+function findMax(elements){
+    let max = -Infinity;
+   for (let i = 0; i < elements.length; i++) {
+     if (typeof elements[i] === "number")
+       if (elements[i] > max) {
+         max = elements[i];
+       }
+   }
+    return max;
 }
```
cjyuan: What do you expect from the following function calls (on extreme cases)?

```
findMax([NaN])
findMax([0, NaN, 1])
```

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,14 @@
-function sum(elements) {
+function sum(elements)
+{    let sum = 0;
+    for (let i = 0; i < elements.length; i++) {
+        if (typeof elements[i] === "number"){
+            sum += elements[i];
+        }
+    }
```
cjyuan: - Indentation is a bit off

- Do you plan to also include `NaN` in the sum?

cjyuan (commented):
In the "implement" folder, you missed updating the `.test.js` files accordingly.

Sprint-1/fix/median.js
KKtech06: Thanks for the feedback. I updated the filter to use Number.isFinite() so NaN, Infinity, and -Infinity are excluded from the median calculation. I also removed the extra copy before sorting since filter() already returns a new array.

Sprint-1/implement/max.js
KKtech06: I tested the edge cases. For [NaN], it was returning -Infinity, so I fixed it by using Number.isFinite() and returning null when no valid numbers are found. For [0, NaN, 1], it correctly returns 1.

Sprint-1/implement/sum.js
KKtech06: Thanks for the feedback. I fixed the indentation and updated the condition to use Number.isFinite() so NaN and Infinity are excluded from the sum.

cjyuan:
Function implementations are ok, but the `.test.js` files are barely modified. Can you implement the Jest tests in those three files?

KKtech06:
Hello CJ i have modified the test.js files and ran them thought Jest and got a Pass for all of the test.js files 

Thank you

Sprint-1/implement/max.test.js
```diff
@@ -16,29 +16,48 @@ const findMax = require("./max.js");
...
+test("given decimal numbers, returns the largest decimal number", () => {
+  expect(findMax([1.2, 3.7, 2.5])).toBe(3.7);
+});
 
 // Given an array with non-number values
 // When passed to the max function
 // Then it should return the max and ignore non-numeric values
+test("ignores non-number values and returns the max", () => {
+  expect(findMax(["hey", 10, "hi", 60, 10])).toBe(60);
+});
```
cjyuan, line 56: When a string representing a valid numeric literal (for example, `"300"`) is compared to a number, 
As a result, the expression `60 < "300"` evaluates to `true`.

Sprint-1/implement/max.test.js
```diff
@@ -16,29 +16,48 @@ const findMax = require("./max.js");
...
 
 // Given an array with only non-number values
 // When passed to the max function
 // Then it should return the least surprising value given how it behaves for all other inputs
-
+test("given only non-number values, returns null", () => {
+  expect(findMax(["hey", "hi", null, undefined, NaN])).toBe(null);
+});
```
cjyuan, line 67: The test on line 62 make it looks like the function considers `null` as the max value. 

You can also consider treating an array with only non-number values the same as an array with no numbers, and give all arrays containing no numbers the same treatment?

Sprint-1/implement/sum.test.js
```diff
@@ -13,25 +13,43 @@ const sum = require("./sum.js");
...
+test("given negative numbers, returns the correct total", () => {
+  expect(sum([-10, -3, -2])).toBe(-15);
+});
 
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
+test("given decimal numbers, returns the correct total", () => {
+  expect(sum([1.5, 2.5, 3])).toBe(7);
+});
```

So the following could happen
```javascript

```

Suggestion: Look up 

KKtech06: Hi CJ I have  updated the test to use toBeCloseTo() because it works better with decimal numbers.

Thank You

Sprint-1/implement/max.test.js
KKtech06: Hi CJ I have updated the test so arrays with no numeric values return null, the same as an empty array. This keeps the behaviour consistent.

Thank You

Sprint-1/implement/max.test.js
KKtech06: Hi CJ I have added a test with "300" to make sure numeric strings are ignored..

Thank You

Sprint-1/implement/max.test.js
```diff
@@ -16,28 +16,52 @@ const findMax = require("./max.js");
 // When passed to the max function
 // Then it should return -Infinity
 // Delete this test.todo and replace it with a test.
-test.todo("given an empty array, returns -Infinity");
+test("given an empty array, returns null", () => {
+  expect(findMax([])).toBe(null);
+});
```
cjyuan, line 21: Consistency is good, but return `null` would violate the original spec. 

Also, it is best practice is to design a function to return only one type of value.

KKtech06: Hello CJ  I have  changed it from null to -Infinity to follow the spec and keep the return type consistent.

Thank you

cjyuan:
Changes look good.

illicitonion:

PR #1030, opened 2026-03-17
Glasgow | 26-ITP-Jan| Martin McLean |Sprint 3|Quote Generator

mjm-git185:
Changelist

I have titled my PR with Region | Cohort | FirstName LastName | Sprint | Assignment Title
My changes meet the requirements of the task
I have tested my changes
My changes follow the style guide
my coursework

### DIFF
--- a/Sprint-3/quote-generator/QuoteGeneratorApp.html
+++ b/Sprint-3/quote-generator/QuoteGeneratorApp.html
@@ -1,0 +1,16 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Title here</title>
+    <script defer src="quotes.js"></script>
+  </head>
+  <body onload="pickNewQuoteToDisplay()">
+    <h1>hello there</h1>
+
+    <p id="quote"></p>
+    <p id="author"></p>
+    <button type="button" class="button" id="new-quote">New quote</button>
+  </body>
+</html>
--- a/Sprint-3/quote-generator/QuoteGeneratorApp.html
+++ b/Sprint-3/quote-generator/QuoteGeneratorApp.html
@@ -3,14 +3,18 @@
   <head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <title>Title here</title>
+    <link href="style.css" rel="stylesheet" />
+    <title>Quote of the day page</title>
     <script defer src="quotes.js"></script>
   </head>
-  <body onload="pickNewQuoteToDisplay()">
-    <h1>hello there</h1>
 
-    <p id="quote"></p>
-    <p id="author"></p>
-    <button type="button" class="button" id="new-quote">New quote</button>
+  <body onload="pickNewQuoteToDisplay()">
+    <h1>quote for today</h1>
+    <div class="contaner">
+      <div class="item1"><p id="quote"></p></div>
+      <div class="item2"><p id="author"></p></div>
+
+      <button type="button" class="item3" id="new-quote">New quote</button>
+    </div>
   </body>
 </html>
--- a/Sprint-3/quote-generator/QuoteGeneratorApp.html
+++ b/Sprint-3/quote-generator/QuoteGeneratorApp.html
@@ -8,7 +8,7 @@
     <script defer src="quotes.js"></script>
   </head>
 
-  <body onload="pickNewQuoteToDisplay()">
+  <body>
     <h1>quote for today</h1>
     <div class="contaner">
       <div class="item1"><p id="quote"></p></div>
--- a/Sprint-3/quote-generator/quotes.js
+++ b/Sprint-3/quote-generator/quotes.js
@@ -1,18 +1,16 @@
-const botton = document
+document
   .getElementById("new-quote")
   .addEventListener("click", pickNewQuoteToDisplay);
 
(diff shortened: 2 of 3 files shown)


Sprint-3/quote-generator/QuoteGeneratorApp.html
```diff
@@ -0,0 +1,20 @@
...
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <link href="style.css" rel="stylesheet" />
+    <title>Quote of the day page</title>
+    <script defer src="quotes.js"></script>
+  </head>
+
+  <body onload="pickNewQuoteToDisplay()">
```
cjyuan: Suggestion: Look up 

> Pros and cons between these two approaches to call a function, `callback()`, on page load?
- `<body onload="callback()">` in HTML code
- `windows.addEventListener("load", callback)` in JS code

Sprint-3/quote-generator/quotes.js
```diff
@@ -1,3 +1,19 @@
+const botton = document
+  .getElementById("new-quote")
+  .addEventListener("click", pickNewQuoteToDisplay);
```
cjyuan: - Is `botton` assigned the value you expect?  What's the return value of `addEventListener()`?

- What is the use of this constant?

Sprint-3/quote-generator/quotes.js
```diff
@@ -490,4 +506,4 @@ const quotes = [
   },
 ];
 
-// call pickFromArray with the quotes array to check you get a random quote
+console.log(pickFromArray(quotes));
```
cjyuan: If this code necessary for the app to function? If not, we should delete it to keep our code clean.

Sprint-3/quote-generator/quotes.js
```diff
@@ -1,3 +1,19 @@
...
+  //gets the quote object and brakes it down to the key and value
+  let quoteAndAuther = pickFromArray(quotes);
+  let quoteForDisplay = quoteAndAuther.quote;
+  let authorForDisplay = quoteAndAuther.author;
+  // prints the values to the page
+  document.getElementById("quote").innerHTML = "'" + quoteForDisplay + "'";
+  document.getElementById("author").innerHTML = ":-" + authorForDisplay;
+}
+
+//function randomQuote()
```
cjyuan: What is this comment about?

Sprint-3/quote-generator/quotes.js
```diff
@@ -1,3 +1,19 @@
+const botton = document
+  .getElementById("new-quote")
+  .addEventListener("click", pickNewQuoteToDisplay);
+
+function pickNewQuoteToDisplay(quote) {
+  //gets the quote object and brakes it down to the key and value
+  let quoteAndAuther = pickFromArray(quotes);
+  let quoteForDisplay = quoteAndAuther.quote;
+  let authorForDisplay = quoteAndAuther.author;
```
cjyuan: `const` instead of `let`?

Sprint-3/quote-generator/quotes.js
```diff
@@ -1,3 +1,19 @@
+const botton = document
+  .getElementById("new-quote")
+  .addEventListener("click", pickNewQuoteToDisplay);
+
+function pickNewQuoteToDisplay(quote) {
```
cjyuan: Does this function need a parameter? `quote` is not used anywhere in the function.

Sprint-3/quote-generator/quotes.js
```diff
@@ -1,3 +1,19 @@
...
+  //gets the quote object and brakes it down to the key and value
+  let quoteAndAuther = pickFromArray(quotes);
+  let quoteForDisplay = quoteAndAuther.quote;
+  let authorForDisplay = quoteAndAuther.author;
+  // prints the values to the page
+  document.getElementById("quote").innerHTML = "'" + quoteForDisplay + "'";
+  document.getElementById("author").innerHTML = ":-" + authorForDisplay;
```
cjyuan, line 17: Suggestion: Look up
> The differences between using `.innerHTML`, `innerText`, and `textContent` to set text content of an HTML element.

Sprint-3/quote-generator/quotes.js
mjm-git185: is it better to think of const as the default and use let only when nessasery

cjyuan: It's up the individual's practice.

I usually focus on making the code work correctly first, then improve the code during the refactoring stage.

Sprint-3/quote-generator/quotes.js
mjm-git185: i was under the impression that i needed to assign a variable for the document.get....  how ever i had a careful look at the botton with the dev tools and it does nothing, so i got rid of it.
thanks

cjyuan (commented):
Instead of placing all the "run on load' code in multiple places:
(1) This function call:
`<body onload="pickNewQuoteToDisplay()">`

and
(2) Code that setup the event listener:
```document
  .getElementById("new-quote")
  .addEventListener("click", pickNewQuoteToDisplay);
```

can you practice placing all the "run on load" code inside a function?
Doing so can make it clearer that "this is what runs when the page loads."
For examples,

```javascript
function setup() {
}

```
or
```javascript
});
```

mjm-git185:
i have made the change to the setUp on load it seems far more readable like this thanks

cjyuan:
Changes look good. Well done.

illicitonion:

PR #1031, opened 2026-03-17
London | 26-ITP-JAN | Asha Ahmed | Sprint 1 | Data Groups

ashaahmed7:
## Learners, PR Template

Self checklist

## Changelist

Completed the exercises in each directory on Sprint 1

## Questions

N/A

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -6,8 +6,17 @@
 // or 'list' has mixed values (the function is expected to sort only numbers).
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
+  if (!Array.isArray(list)) return null;
+  const num = list.filter((l) => typeof l === "number").sort((a, b) => a - b);
+  if (num.length < 1) return null;
+  if (num.length === 1) return num[0];
+
+  const middleIndex = Math.floor(num.length / 2);
+
+  if (num.length % 2 === 0) {
+    return (num[middleIndex] + num[middleIndex - 1]) / 2;
+  }
+  const median = num[middleIndex];
   return median;
 }
 
--- a/Sprint-1/implement/sum.js
+++ b/Sprint-1/implement/sum.js
@@ -1,4 +1,22 @@
 function sum(elements) {
+  if (!Array.isArray(elements)) {
+    return null;
+  }
+  // Check if array is empty
+  if (elements.length === 0) {
+    return 0;
+  }
+  // Filter only numeric values
+  const numericElements = elements.filter((item) => typeof item === "number");
+  if (numericElements.length === 0) {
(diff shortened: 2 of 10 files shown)


Sprint-1/fix/median.js
```diff
@@ -6,8 +6,17 @@
 // or 'list' has mixed values (the function is expected to sort only numbers).
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
+  if (!Array.isArray(list)) return null;
+  const num = list.filter((l) => typeof l === "number").sort((a, b) => a - b);
```
cjyuan, line 10: - Do you plan to include `-Infinity`, `Infinity`, and `NaN` in the median calculation (and also in the function in `implement/sum.js`)?

- Naming convention for variables that stores an array is: 
  - Append a suffix `Array` or `List` to the name (e.g., `numberArray`)

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,22 @@
...
+  }
+  // Check if array is empty
+  if (elements.length === 0) {
+    return 0;
+  }
+  // Filter only numeric values
+  const numericElements = elements.filter((item) => typeof item === "number");
+  if (numericElements.length === 0) {
+    return -Infinity;
+  }
```
cjyuan, line 13: Should the function preserve this characteristic?
`sum(["apple", "banana", 1, 10]) == sum(["apple", "banana"])  + sum([1, 10])`

Sprint-1/fix/median.js
ashaahmed7: I currently filter values using typeof l === "number", which includes Infinity, -Infinity, and NaN.
However, these values can distort the median calculation (especially NaN, which propagates).

I plan to refine the filter to include only finite numbers using Number.isFinite() for more robust behaviour. 

const numbers = list
  .filter((l) => Number.isFinite(l))
  .sort((a, b) => a - b);

Sprint-1/implement/sum.js
ashaahmed7: Good question — with the current implementation, this property does not hold because the function returns -Infinity when no numeric values are present.

This causes inconsistencies when combining results (e.g., -Infinity + 11 = -Infinity).

If we want the function to preserve this additive property, it would be better to return 0 when no numeric values are found, since 0 is the neutral element for addition.

I can update the function accordingly depending on the intended behaviour.  

if (numericElements.length === 0) {
  return 0;
}

cjyuan: The spec only mentions "least surprising" value. It is up to you what's that value should be. 
I can only provide you with additional info to make informed decision.

cjyuan:
Your explanation sounds good to me.

illicitonion:

PR #1033, opened 2026-03-17
London | 26-ITP-Jan | Karla Grajales| Sprint 1 | Data Groups

Grajales-K:
## Learners, PR Template

Self checklist

## Changelist

In this project, I mastered TDD (Test Driven Development) while implementing various functions. I learned different ways to run tests, such as using npm test and npx jest --watch to run specific tests or all of them at once.

I also learned how to transition from a classic for loop to a for...of loop, which is simpler and more readable. I gained experience using array methods like .filter() and .reduce(), working with Math.max(), using the modulo operator (%), and manipulating arrays with the spread operator (...). This has given me a more confident and solid foundation in TDD.

## Questions

I noticed that without the compare function (a, b) => a - b, the .sort() method treats numbers as strings (so '110' comes before '20'). Is it considered best practice to always provide a compare function even if we think the data is already clean?

Thank you for your time reviewing my code.


Sprint-1/fix/median.js
```diff
@@ -5,10 +5,45 @@
...
+
+function calculateMedian(list) { 
+  // if is not an array 
+  if(!Array.isArray(list)){
+    return null;
+  }
+
+  //checking only numbers and not null and empty
+  const numsOnly = list.filter(
+    (n) => typeof n === "number" && n !== null && !isNaN(n)
```
cjyuan: One of the checks is redundant.

Sprint-1/fix/median.js
```diff
@@ -5,10 +5,45 @@
...
+  );
+
+  //if return numbers check if after cleaning the length is same as expected or empty
+  if (numsOnly.length === 0) {
+    return null;
+  }
+
+  // now we change to use the numsOnly otherwise will take string as numbers
+  const sortArr = [...numsOnly].sort((a, b) => a - b);
```
cjyuan: Why make a copy of `numsOnly` before sorting?

Sprint-1/implement/dedupe.js
```diff
@@ -1 +1,9 @@
-function dedupe() {}
+
+
+function dedupe(val) {
```
cjyuan: Can you think of a name that could more clearly indicate the parameter is an array of numbers?

Sprint-1/implement/dedupe.test.js
```diff
@@ -16,13 +16,31 @@ E.g. dedupe([1, 2, 1]) returns [1, 2]
...
+test("given an empty array, it returns an empty array", () => {
+  expect(dedupe([])).toEqual([]);
+});
 
 // Given an array with no duplicates
 // When passed to the dedupe function
 // Then it should return a copy of the original array
+test("given an array with no duplicates, it returns a copy of the original array", () => {
+  expect(dedupe(["a", "b", "c"])).toEqual(["a", "b", "c"]);
+});
```
cjyuan, line 33: This test should fail if the function returns the original array (instead of a **copy** of the original array).

Sprint-1/implement/max.test.js
```diff
@@ -5,39 +5,63 @@ In this kata, you will need to implement a function that find the largest numeri
...
+});
+
 
 // Given an array with only non-number values
 // When passed to the max function
 // Then it should return the least surprising value given how it behaves for all other inputs
+test("given an array with only non-number values, it returns -Infinity", () => {
+  expect(findMax(["abc", "def"])).toEqual(-Infinity);
+});
+
```
cjyuan, line 81: When a string representing a valid numeric literal (for example, `"300"`) is compared to a number, 

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,10 @@
+// 1. filter: leep only elements that are onlyNumbers.
+// 2. spread (...) take the numbers out of the Array
+// 3. Math.max find the largest number and return -infinity if empty.
+
 function findMax(elements) {
+    const onlyNumbers = elements.filter((element) => typeof element === "number")
+    return Math.max(...onlyNumbers);
 }
```
cjyuan, line 8: What do you expect from the following function calls (on extreme cases)?

```
findMax([NaN])
findMax([0, NaN, 1])
```

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,6 @@
 function sum(elements) {
+
+    return elements.reduce((acc, curr) => (typeof curr === "number" ? acc + curr : acc), 0)
 }
```

```
sum([NaN, 1]);
sum([Infinity, -Infinity]);
```

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,41 @@ const sum = require("./sum.js");
...
+test("Should return the correct sum even when the array contains negative numbers", () => {
+  expect(sum([-1, -2, -3])).toEqual(-6);
+});
 
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
+test("Should return the correct sum even when the array contains decimal numbers", () => {
+  expect(sum([1.4, 1.6, 5.8])).toEqual(8.8);
+});
```

So the following could happen
```javascript

```

Suggestion: Look up 

Sprint-1/refactor/includes.js
```diff
@@ -1,12 +1,11 @@
...
-  for (let index = 0; index < list.length; index++) {
-    const element = list[index];
-    if (element === target) {
+function includes(list, target){
+  for (const element of list){
+    if (element === target){
       return true;
     }
-  }
+  }
```
cjyuan: Spacing is a bit inconsistently.

?

cjyuan:
> I noticed that without the compare function (a, b) => a - b, the .sort() method treats numbers as strings (so '110' comes before '20'). Is it considered best practice to always provide a compare function even if we think the data is already clean?

Have you read the documentation of `Array.prototype.sort()` on MDN to find out how the function works?
I think AI can also explain why.

I am not sure what you meant by "data is already clean".

Sprint-1/fix/median.js
Grajales-K: thanks @cjyuan Removed n !== null because typeof n === 'number' already excludes nulls. Since it's an && check, if the first part is false, JavaScript short-circuits and doesn't evaluate the rest

Sprint-1/fix/median.js
Grajales-K: @cjyuan, I removed the spread operator [...] because .filter() already returns a new array. Sorting numsOnly directly is safe and more efficient since it doesn't mutate the original input.

Sprint-1/implement/dedupe.js
Grajales-K: Hi @cjyuan, I've renamed the parameter to items since the function receives an array of different types, including both strings and numbers.

Sprint-1/implement/dedupe.test.js
Grajales-K: Hi @cjyuan, I've updated the test to use a variable for the input. This allows me to use not.toBe() to prove that the function returns a new reference (a copy) and not the original array, even though the content is identical.

Sprint-1/implement/max.test.js
Grajales-K: Hi @cjyuan, I've added the tests with numeric strings (like "300"). My function ignores them because it only accepts actual numbers, which helps avoid any confusion with type coercion.

Sprint-1/implement/max.js
Grajales-K: Got it! I added the !Number.isNaN check to the filter and included new test cases for NaN. Now the function correctly ignores them and returns the expected values, just like the median implementation.

Sprint-1/implement/sum.js
Grajales-K: @cjyuan 
The original function returned NaN because it didn't filter out NaN or Infinity. I've updated it to ignore these cases using !Number.isNaN, ensuring that sum([NaN, 1]) returns 1, and an array of only invalid values returns -Infinity for consistency.

Sprint-1/implement/sum.test.js
Grajales-K: @cjyuan  I see it.



```

test("Should return the correct sum even when the array contains decimal numbers", () => {

expect(sum([0.1, 0.2, 0.005])).toEqual(1.805);

});

    Expected: 1.805

    Received: 0.30500000000000005
```
```

test("Should return the correct sum even when the array contains decimal numbers", () => {

  expect(sum([0.1, 0.2, 0.005])).toBeCloseTo(0.305);

}); 


  ✓ pass
  
```
  
  I've updated the decimal test to use .toBeCloseTo(). This correctly handles the floating-point precision error I encountered (0.30500000000000005) without needing to convert the function output to a string.

Grajales-K:
@cjyuan Thank you for your review and am sorry for the delay! I was very sick this weekend with migraine.

I've updated the tests to handle floating-point precision using toBeCloseTo(). I also added edge cases for NaN and Infinity using a data-driven approach to ensure the function is robust

Thank you.

Sprint-1/implement/dedupe.test.js
```diff
@@ -16,13 +16,35 @@ E.g. dedupe([1, 2, 1]) returns [1, 2]
...
 // Given an array with no duplicates
 // When passed to the dedupe function
 // Then it should return a copy of the original array
+test("given an array with no duplicates, it returns a copy of the original array", () => {
+  const input = ["a", "b", "c"];
+  const result = dedupe(input);
+
+  expect(result).toEqual(input);
+  expect(result).not.toBe(input);
+});
```
cjyuan, line 33: There is a chance that, even though `result` has incorrect elements (for example, `[]`), 

Sprint-1/implement/max.test.js
cjyuan: I don't see this test updated in this file.

In practice, we write tests not only to verify our current implementation, but also to ensure that future changes do not alter the function's expected behavior.

Currently the following function could also pass this test:
```
function findMax(elements) {
    const onlyNumbers = elements.map(Number).filter((el)=>!Number.isNaN(el));
    return Math.max(...onlyNumbers);
}
```

Sprint-1/implement/dedupe.test.js
Grajales-K: @cjyuan, I have refactored tests to avoid false positives. Instead of comparing the result against the input variable (which could be mutated), I'm now using explicit literal values and verifying that the original array remains unchanged

Sprint-1/implement/max.test.js
Grajales-K: Hi @cjyuan, my implementation was already using typeof to strictly filter for numbers, the test suite wasn't explicitly challenging that logic. I've now added a test case with a numeric string ('300') to ensure that any future refactors using loose type coercion (like Number()) will fail. Thanks for this

cjyuan:
Changes look good. Well done.

illicitonion:

PR #1034, opened 2026-03-17
BIRMINGHAM | ITP 26 JAN | Merve Reis | Sprint 2 | Module Data Groups

mervereis:
## Learners, PR Template

Self checklist

Functions created and test cases written for

### DIFF
--- a/Sprint-2/debug/recipe.js
+++ b/Sprint-2/debug/recipe.js
@@ -11,5 +11,7 @@ const recipe = {
 };
 
 console.log(`${recipe.title} serves ${recipe.serves}
-  ingredients:
-${recipe}`);
+  ingredients:`);
+for (const ingredient of recipe.ingredients) {
+  console.log(ingredient);
+}
--- a/Sprint-2/debug/recipe.js
+++ b/Sprint-2/debug/recipe.js
@@ -11,7 +11,5 @@ const recipe = {
 };
 
 console.log(`${recipe.title} serves ${recipe.serves}
-  ingredients:`);
-for (const ingredient of recipe.ingredients) {
-  console.log(ingredient);
-}
+ingredients:
+${recipe.ingredients.join("\n")}`);
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,21 +1,9 @@
 function contains(obj, prop) {
-  if (obj === null) {
+  if (obj === null || typeof obj !== "object" || Array.isArray(obj)) {
     return false;
   }
 
-  if (typeof obj !== "object") {
-    return false;
-  }
-
-  if (Array.isArray(obj)) {
-    return false;
-  }
-
-  if (prop in obj) {
-    return true;
-  } else {
-    return false;
-  }
+  return Object.hasOwn(obj, prop);
 }
 
 module.exports = contains;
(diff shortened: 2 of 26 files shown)


Sprint-2/debug/recipe.js
```diff
@@ -11,5 +11,7 @@ const recipe = {
 };
 
 console.log(`${recipe.title} serves ${recipe.serves}
-  ingredients:
-${recipe}`);
+  ingredients:`);
+for (const ingredient of recipe.ingredients) {
+  console.log(ingredient);
+}
```
cjyuan: Note: Since ingredient values are separated by '\n' in the output, we could also use `Array.prototype.join()` to construct the equivalent string and then output the resulting string.

Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,21 @@
...
+
+  if (typeof obj !== "object") {
+    return false;
+  }
+
+  if (Array.isArray(obj)) {
+    return false;
+  }
+
+  if (prop in obj) {
```
```
```

Sprint-2/implement/contains.test.js
```diff
@@ -1,35 +1,28 @@
...
+    expect(contains({ a: 1, b: 2 }, "a")).toBe(true);
+  });
+
+  test("returns false when object does not contain the property", () => {
+    expect(contains({ a: 1, b: 2 }, "c")).toBe(false);
+  });
+
+  test("returns false when given an array instead of an object", () => {
+    expect(contains([], "a")).toBe(false);
+  });
```
cjyuan, line 18: This test does not yet confirm that the function correctly returns false when the first argument is an array. 

Sprint-2/implement/querystring.js
```diff
@@ -1,16 +1,34 @@
...
+      key = pair;
+      value = "";
+    } else {
+      key = pair.slice(0, index);
+      value = pair.slice(index + 1);
+    }
+
+    if (key) {
+      queryParams[key] = value;
+    }
```
cjyuan, line 28: Note: (No change required)

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,15 @@
-function tally() {}
+function tally(items) {
+  if (!Array.isArray(items)) {
+    throw new Error("Input must be an array");
+  }
+
+  const counts = {};
```
```
```

mervereis: @cjyuan I’ve switched to Object.create(null) to avoid issues with inherited properties

Sprint-2/implement/querystring.js
mervereis: @cjyuan thanks, I'll take a look at it

Sprint-2/implement/contains.test.js
mervereis: @cjyuan I’ve updated the test to use a non-empty array with a valid index

Sprint-2/implement/contains.js
mervereis: @cjyuan I’ve switched to Object.hasOwn property

Sprint-2/debug/recipe.js
mervereis: @cjyuan I’ve refactored it to use join("\n") for a cleaner and more concise implementation

cjyuan:
Changes look good. Well done.

illicitonion:

PR #1035, opened 2026-03-17
Glasgow | JAN-ITP-2026 | Alasdair MacDonald | Sprint 3 | Quote Generator app

MacDonald91:
## Learners, PR Template

Self checklist

## Changelist

Finished building a complete working quote generator app.

## Questions

no questions.

### DIFF
--- a/Sprint-3/quote-generator/quotes.js
+++ b/Sprint-3/quote-generator/quotes.js
@@ -491,3 +491,18 @@ const quotes = [
 ];
 
 // call pickFromArray with the quotes array to check you get a random quote
+
+function displayRandomQuote() {
+  const randomQuote = pickFromArray(quotes);
+
+  document.getElementById("quote").textContent = randomQuote.quote;
+  document.getElementById("author").textContent = randomQuote.author;
+}
+
+// run when page loads2
+displayRandomQuote();
+
+// button click
+document
+  .getElementById("new-quote")
+  .addEventListener("click", displayRandomQuote);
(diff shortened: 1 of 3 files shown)


Sprint-3/quote-generator/quotes.js
```diff
@@ -491,3 +491,18 @@ const quotes = [
...
+  document.getElementById("author").textContent = randomQuote.author;
+}
+
+// run when page loads2
+displayRandomQuote();
+
+// button click
+document
+  .getElementById("new-quote")
+  .addEventListener("click", displayRandomQuote);
```
cjyuan: Can you put all the "run on load" code inside a function? 
Doing so can make it clearer that "this is what runs when the page loads."
For examples,

```javascript
function setup() {
}

```
or
```javascript
});
```

cjyuan (commented):
Code is working good.

Note: We should respect instructions like `DO NOT EDIT BELOW HERE`; it is usually there for a reason. If you are curious about why, you can ask AI `Why should programmers respect "DO NOT EDIT BELOW HERE" instruction in a file?`

Sprint-3/quote-generator/quotes.js
MacDonald91: Thanks for the suggestion!

I’ve moved all the “run on load” logic into a `setup` function and attached it to the window load event to make the structure clearer and easier to understand.

Let me know if there’s anything else I can improve.

cjyuan:
All good. Well done.

PR #1036, opened 2026-03-18
London | ITP-Jan-2026 | Ping Wang | Sprint 3 | Quote Generator                                                      Sprint 3/quote generator

pathywang:
Self checklist

## Changelist

It did take me a while to get around about this course work because it is quite different from previous single JavaScript code work.

### DIFF
--- a/Sprint-3/quote-generator/index.html
+++ b/Sprint-3/quote-generator/index.html
@@ -3,7 +3,7 @@
   <head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <title>Title here</title>
+    <title>Quote generator app</title>
     <script defer src="quotes.js"></script>
   </head>
   <body>
--- a/Sprint-3/quote-generator/index.html
+++ b/Sprint-3/quote-generator/index.html
@@ -4,6 +4,7 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>Quote generator app</title>
+    <link rel="stylesheet" href="style.css">
     <script defer src="quotes.js"></script>
   </head>
   <body>
--- a/Sprint-3/quote-generator/index.html
+++ b/Sprint-3/quote-generator/index.html
@@ -9,9 +9,11 @@
   </head>
   <body>
     <h1>hello there</h1>
-    <p id="quote"></p>
-    <p id="author"></p>
-    <p class="container">Test</p>
-    <button type="button" id="new-quote">New quote</button>
+    
+    <div class="container">
+      <p id="quote"></p>
+      <p id="author"></p>
+      <button type="button" id="new-quote">New Quote</button>
+    </div>
   </body>
 </html>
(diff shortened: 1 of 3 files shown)


Sprint-3/quote-generator/index.html
```diff
@@ -3,13 +3,15 @@
...
-    <title>Title here</title>
+    <title>Quote generator app</title>
+    <link rel="stylesheet" href="style.css">
     <script defer src="quotes.js"></script>
   </head>
   <body>
     <h1>hello there</h1>
     <p id="quote"></p>
     <p id="author"></p>
+    <p class="container">Test</p>
```
cjyuan: What's the purpose of this element (line 14)?

Sprint-3/quote-generator/quotes.js
```diff
@@ -491,3 +491,20 @@ const quotes = [
 ];
 
 // call pickFromArray with the quotes array to check you get a random quote
+
+console.log(pickFromArray(quotes));
```
cjyuan: Any unnecessary code should be removed to keep our code clean.

Sprint-3/quote-generator/quotes.js
```diff
@@ -491,3 +491,20 @@ const quotes = [
...
+function showQuote() {
+  const randomQuote = pickFromArray(quotes);
+
+  quoteElement.textContent = randomQuote.quote;
+  authorElement.textContent =randomQuote.author;
+}
+
+button.addEventListener("click", showQuote);
+
+showQuote();
```
Would be even better to place the code inside a function to make it clearer that "this is what runs when the page loads."

For examples,

```javascript
function setup() {
}

```
or
```javascript
});
```

cjyuan (commented):
While Lighthouse score is not part of the requirement, it's a good to keep the score at 100.

Sprint-3/quote-generator/index.html
pathywang: sorry, my  intention is to create a container around the quote, author, and button, but <p class="container">Test</p> does not work so i amend the index.html

Sprint-3/quote-generator/quotes.js
pathywang: Thanks, i delete console.log(pickFromArray(quotes)) now even though i was going to test my code,but i should have deleted it before i commited my code

Sprint-3/quote-generator/quotes.js
pathywang: Thanks, it is.so i update my code now but it did take me a while to get round so i need more practice

pathywang: i checked my index.html with lighthouse, now the performance is 100. From now on, i should check my html with light house as a routine,thanks

cjyuan: According to Lighthouse,

> Background and foreground colors do not have a sufficient contrast ratio.

If you have made any changes, don't forget to push the commits to GitHub as well.

cjyuan:
Changes to the HTML and JS look good.

Sprint-3/quote-generator/quotes.js
pathywang: Thanks, i will have a look

illicitonion:

PR #1042, opened 2026-03-18
Sheffield | 26-ITP-jan | Richard Frimpong | Sprint 2 | Data Groups

Richiealx:
## Learners, PR Template

Self checklist

## Changelist

This pull request contains my completed work for **Sprint 2 - Data Groups**.

### Debug
- Fixed `address.js` (correct object property access)
- Fixed `author.js` (correct object iteration)
- Fixed `recipe.js` (proper formatting of output)

### Implement
- Implemented `contains` using own-property checking (avoids inherited properties)
- Implemented `createLookup` to convert key-value pairs into an object
- Implemented `tally` using `Object.create(null)` to handle edge cases like `"toString"`
- Improved `parseQueryString` to:
  - handle values containing "="
  - handle missing values and keys
  - ignore trailing "&"
  - decode URL-encoded keys and values

### Interpret
- Analysed and explained the issue in `invert.js`
- Fixed implementation using bracket notation
- Added tests to verify correctness

### Stretch
- Implemented `countWords` with:
  - punctuation handling
  - case normalisation
  - whitespace handling
  - safe object creation to handle words like `"constructor"`
- Refactored `calculateMode` into smaller helper functions:
  - frequency tracking
  - mode selection

### Testing
- All tests pass successfully
- 7 test suites and 27 tests passing

### DIFF
--- a/Sprint-2/debug/recipe.js
+++ b/Sprint-2/debug/recipe.js
@@ -1,8 +1,9 @@
 // Predict and explain first...
 
 // This program should log out the title, how many it serves and the ingredients.
-// Each ingredient should be logged on a new line
-// How can you fix it?
+// Each ingredient should be logged on a new line.
+// The original code tried to print the entire recipe object,
+// which resulted in "[object Object]" instead of the ingredients.
 
 const recipe = {
   title: "bruschetta",
@@ -10,6 +11,11 @@ const recipe = {
   ingredients: ["olive oil", "tomatoes", "salt", "pepper"],
 };
 
-console.log(`${recipe.title} serves ${recipe.serves}
-  ingredients:
-${recipe}`);
+// Print title and serving size
+console.log(`${recipe.title} serves ${recipe.serves}`);
+console.log("ingredients:");
+
+// Loop through the ingredients array and print each one
+for (const ingredient of recipe.ingredients) {
+  console.log(ingredient);
+}
--- a/Sprint-2/debug/recipe.js
+++ b/Sprint-2/debug/recipe.js
@@ -15,7 +15,5 @@ const recipe = {
 console.log(`${recipe.title} serves ${recipe.serves}`);
 console.log("ingredients:");
 
-// Loop through the ingredients array and print each one
-for (const ingredient of recipe.ingredients) {
-  console.log(ingredient);
-}
+// Print each ingredient on a new line using join()
+console.log(recipe.ingredients.join("\n"));
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,3 +1,27 @@
-function contains() {}
+/**
+ * contains()
+ *
+ * Checks whether an object contains a specific property.
+ *
+ * @param {object} obj - The object to check.
+ * @param {string} propertyName - The property name we want to check.
+ * @returns {boolean} True if the property exists, otherwise false.
+ */
+
+function contains(obj, propertyName) {
+  // Validate that obj is actually an object
+  // and not null or an array
+  if (obj === null || typeof obj !== "object" || Array.isArray(obj)) {
+    return false;
+  }
+
+  // Validate propertyName
+  if (typeof propertyName !== "string" || propertyName.length === 0) {
+    return false;
+  }
+
+  // Check if the object has the property as its own key
+  return Object.prototype.hasOwnProperty.call(obj, propertyName);
+}
 
 module.exports = contains;
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,27 +1,20 @@
 /**
  * contains()
  *
- * Checks whether an object contains a specific property.
+ * Checks whether an object contains a specific own property.
  *
  * @param {object} obj - The object to check.
- * @param {string} propertyName - The property name we want to check.
- * @returns {boolean} True if the property exists, otherwise false.
+ * @param {*} propertyName - The property name to check.
+ * @returns {boolean} True if the object has the property as its own key, otherwise false.
  */
-
 function contains(obj, propertyName) {
-  // Validate that obj is actually an object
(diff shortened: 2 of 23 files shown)


Sprint-2/debug/recipe.js
```diff
@@ -1,15 +1,21 @@
...
-  ingredients:
-${recipe}`);
+// Print title and serving size
+console.log(`${recipe.title} serves ${recipe.serves}`);
+console.log("ingredients:");
+
+// Loop through the ingredients array and print each one
+for (const ingredient of recipe.ingredients) {
+  console.log(ingredient);
+}
```
cjyuan: Since ingredient values are separated by '\n' in the output, we could also use `Array.prototype.join()` to construct the equivalent string and then output the resulting string.

Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,27 @@
...
+    return false;
+  }
+
+  // Validate propertyName
+  if (typeof propertyName !== "string" || propertyName.length === 0) {
+    return false;
+  }
+
+  // Check if the object has the property as its own key
+  return Object.prototype.hasOwnProperty.call(obj, propertyName);
```
cjyuan: Could also use `Object.hasOwn()`.

Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,27 @@
...
+  // Validate that obj is actually an object
+  // and not null or an array
+  if (obj === null || typeof obj !== "object" || Array.isArray(obj)) {
+    return false;
+  }
+
+  // Validate propertyName
+  if (typeof propertyName !== "string" || propertyName.length === 0) {
+    return false;
+  }
```
cjyuan: - Values of non-string type can be used as property names; they are just implicitly converted to equivalent strings. For example, 

```
let obj = {};
obj[3.0] = 12;           // 3.0 is treated as "3" when used as key
console.log(obj["3"]);   // output 12 
console.log(obj[4-1]);   // output 12 too
console.log(Object.hasOwn(obj, 3));  // output true
```

- Empty string can also serve as key.

Sprint-2/implement/querystring.js
```diff
@@ -1,12 +1,43 @@
...
+    if (separatorIndex === -1) {
+      queryParams[pair] = "";
+      continue;
+    }
+
+    // Extract the key and everything after the first "=" as the value
+    const key = pair.slice(0, separatorIndex);
+    const value = pair.slice(separatorIndex + 1);
+
     queryParams[key] = value;
```
cjyuan: No change required.

Note: 

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,33 @@
...
+    // If the item already exists in the object, increase the count
+    if (counts[item]) {
+      counts[item] += 1;
+    } else {
+      // Otherwise initialise it
+      counts[item] = 1;
+    }
+  }
+
+  return counts;
```
```
```

cjyuan (commented):
Code looks good. I just have some suggestions.

Sprint-2/implement/tally.js
Richiealx: Hi @cjyuan,

Thank you for your feedback.

I’ve now updated my implementation to address all the points raised:

- Updated `contains` to use proper own-property checking and improved the tests to cover inherited properties, arrays, non-string keys, and edge cases
- Fixed `tally` by using an object with no inherited properties and added an edge-case test for values like "toString"
- Improved `querystring` to handle URL decoding and additional edge cases
- Replaced the remaining `test.todo` in `lookup.test.js` with full test coverage, including validation and malformed input cases
- Reviewed `recipe.js` and applied the suggested improvement using `join` for cleaner output

All test suites are now passing with no remaining TODO tests.
I also corrected the branch issue so this PR now only contains Sprint-2 files.

Thanks again for the helpful suggestions.

cjyuan:
On this PR branch, I don't see any changes made to the files in the Sprint-2 folder (since my previous review).

That is, I don't see some of the changes which you said you did:

> - Updated contains to use proper own-property checking and improved the tests to cover inherited properties, arrays, non-string keys, and edge cases
> - Fixed tally by using an object with no inherited properties and added an edge-case test for values like "toString"
> - Improved querystring to handle URL decoding and additional edge cases
> - Reviewed recipe.js and applied the suggested improvement using join for cleaner output

Richiealx:
Thank you for pointing that out.

You were right. Those specific changes were not yet present on the Sprint-2 PR branch. I’ve now updated the Sprint-2 files to include the requested fixes:

- updated `contains` to use proper own-property checking and improved the tests to cover inherited properties, arrays, non-string keys, and empty-string keys
- fixed `tally` to use an object with no inherited properties and added the `"toString"` edge-case test
- improved `querystring` to handle URL decoding and added an extra decoding test
- updated `recipe.js` to use `join()` for cleaner ingredient output

I also fixed `countWords` so the full Sprint-2 test suite passes cleanly.

All Sprint-2 tests are now passing on this PR branch:
- 7 test suites passed
- 36 tests passed

Sprint-2/implement/contains.test.js
```diff
@@ -1,35 +1,45 @@
...
+  });
+
+  test("returns false when given null", () => {
+    expect(contains(null, "a")).toBe(false);
+  });
+
+  test("supports non-string property names", () => {
+    const obj = { 3: 12 };
+    expect(contains(obj, 3)).toBe(true);
+  });
```
cjyuan, line 39: This test can be included in the test category on line 16.

There is no non-string* properties. When we express a property as a non-string value, it is converted to a string. So `"{ 3: 12 }"` is identical to `{"3": 12}`. 


*: Except for the special type -- Symbol.

cjyuan (commented):
Changes look good. Well done.

illicitonion:

PR #1043, opened 2026-03-18
BIRMINGHAM | ITP-JAN 26 | Merve Reis | Sprint 3 | Quote Generator

mervereis:
## Learners, PR Template

Self checklist

## Changelist
Created Quote Generator application. 
Updated index.html file, added quote generator functionality and created styles for it.

### DIFF
--- a/Sprint-3/quote-generator/quotes.js
+++ b/Sprint-3/quote-generator/quotes.js
@@ -491,32 +491,36 @@ const quotes = [
 ];
 
 // call pickFromArray with the quotes array to check you get a random quote
-const quoteEl = document.getElementById("quote");
-const authorEl = document.getElementById("author");
-const btn = document.getElementById("new-quote");
-const toggle = document.getElementById("autoplay-toggle");
-const statusEl = document.getElementById("autoplay-status");
+function setup() {
+  const quoteEl = document.getElementById("quote");
+  const authorEl = document.getElementById("author");
+  const btn = document.getElementById("new-quote");
+  const toggle = document.getElementById("autoplay-toggle");
+  const statusEl = document.getElementById("autoplay-status");
 
-let interval = null;
+  let interval = null;
 
-function showQuote() {
-  const picked = pickFromArray(quotes);
-  console.log(picked);
-  quoteEl.textContent = picked.quote;
-  authorEl.textContent = "- " + picked.author;
+  function showQuote() {
+    const picked = pickFromArray(quotes);
+    console.log(picked);
+    quoteEl.textContent = picked.quote;
+    authorEl.textContent = picked.author;
+  }
+
+  btn.addEventListener("click", showQuote);
+
+  toggle.addEventListener("change", () => {
+    if (toggle.checked) {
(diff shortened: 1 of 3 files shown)


Sprint-3/quote-generator/quotes.js
```diff
@@ -491,3 +491,32 @@ const quotes = [
...
+    statusEl.textContent = "Auto Play: ON";
+    interval = setInterval(showQuote, 3000);
+  } else {
+    statusEl.textContent = "Auto Play: OFF";
+    clearInterval(interval);
+    interval = null;
+  }
+});
+
+showQuote();
```
Would be even better to place the code (and the constants/variables needed locally by these code) inside a function to make it clearer that "this is what runs when the page loads."

For examples,

```javascript
function setup() {
}

```
or
```javascript
});
```

cjyuan (commented):
Code is working good. 

- There is some unnecessary code you can remove to keep the code cleaner.

Sprint-3/quote-generator/quotes.js
```diff
@@ -491,3 +491,32 @@ const quotes = [
...
+const toggle = document.getElementById("autoplay-toggle");
+const statusEl = document.getElementById("autoplay-status");
+
+let interval = null;
+
+function showQuote() {
+  const picked = pickFromArray(quotes);
+  console.log(picked);
+  quoteEl.textContent = picked.quote;
+  authorEl.textContent = "- " + picked.author;
```
cjyuan: Since `"- "` is part of the presentation logic, it is better to keep it in the CSS or in HTML.

Sprint-3/quote-generator/quotes.js
mervereis: @cjyuan I moved the logic what we run while page load into the setup function then I called it with eventListener as you suggest

Sprint-3/quote-generator/quotes.js
mervereis: @cjyuan I moved the '-' part to CSS file

Sprint-3/quote-generator/quotes.js
```diff
@@ -491,32 +491,36 @@ const quotes = [
...
 
-function showQuote() {
-  const picked = pickFromArray(quotes);
-  console.log(picked);
-  quoteEl.textContent = picked.quote;
-  authorEl.textContent = "- " + picked.author;
-}
+  function showQuote() {
+    const picked = pickFromArray(quotes);
+    console.log(picked);
```
cjyuan, line 505: This seems like a piece of debugging code. Best practice is to delete all unnecessary code before submitting the code for review.

cjyuan (commented):
Moving all the code that implements the app into `setup()` is not what I had in mind; some of those code are just function definitions. What I had in mind was,

```
... // shared constants/variables declaration

... // function definitions

function setup() {
  ... // Code that setups the callback functions

  showQuote();
}

window.addEventListener("load", setup);
```

No change required. Organizing code is not an easy task and what I am suggesting here is not the only way. At least you have experienced one way to organize code in a local scope.

Code looks good. Well done.

illicitonion:

PR #1045, opened 2026-03-19
Manchester | 26-ITP-Jan | Ofonime Edak | Sprint 1 | Data group

Ofonimeedak:
Self checklist

## Changelist
I have made changes to the following files, sum.js,include.js and all test cases
I have followed the acceptance criteria for all katas
This PR contains functions for de-duplication and summation of arrays
It also contained a refactored function and different test cases

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -6,9 +6,23 @@
 // or 'list' has mixed values (the function is expected to sort only numbers).
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  if (!Array.isArray(list)) return null;
+
+  const numbers = [];
+  for (const x of list) {
+    if (typeof x === "number" && !isNaN(x)) {
+      numbers.push(Number(x));
+    }
+  }
+  if (numbers.length === 0) return null;
+  numbers.sort((a, b) => a - b);
+  const length = numbers.length;
+  const middleIndex = Math.floor(length / 2);
+  if (length % 2 === 0) {
+    return (numbers[middleIndex - 1] + numbers[middleIndex]) / 2;
+  } else {
+    return numbers[middleIndex];
+  }
 }
 
 module.exports = calculateMedian;
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -7,13 +7,8 @@
 
 function calculateMedian(list) {
   if (!Array.isArray(list)) return null;
+  const numbers = list.filter((e) => typeof e === "number" && !isNaN(e));
 
-  const numbers = [];
-  for (const x of list) {
-    if (typeof x === "number" && !isNaN(x)) {
-      numbers.push(Number(x));
-    }
-  }
   if (numbers.length === 0) return null;
   numbers.sort((a, b) => a - b);
   const length = numbers.length;
--- a/Sprint-1/implement/dedupe.js
+++ b/Sprint-1/implement/dedupe.js
@@ -1,15 +1,11 @@
 function dedupe(arr) {
-
-if(arr.length===0) return arr;
-const dedupeArray=[]
-for(let i=0;i<arr.length;i++){
-
-    if(!dedupeArray.includes(arr[i])){
-        dedupeArray.push(arr[i])
+  if (arr.length === 0) return [];
+  const dedupeArray = [];
+  for (let i = 0; i < arr.length; i++) {
+    if (!dedupeArray.includes(arr[i])) {
+      dedupeArray.push(arr[i]);
     }
+  }
+  return dedupeArray;
 }
-return dedupeArray;
-
-
-}
-module.exports=dedupe;
+module.exports = dedupe;
--- a/Sprint-1/implement/dedupe.js
+++ b/Sprint-1/implement/dedupe.js
@@ -6,6 +6,8 @@ function dedupe(arr) {
       dedupeArray.push(arr[i]);
     }
   }
+  
   return dedupeArray;
 }
+console.log(dedupe([1, 1, 2]));
 module.exports = dedupe;
--- a/Sprint-1/implement/dedupe.js
+++ b/Sprint-1/implement/dedupe.js
@@ -1,1 +1,15 @@
-function dedupe() {}
+function dedupe(arr) {
+
+if(arr.length===0) return arr;
+const dedupeArray=[]
+for(let i=0;i<arr.length;i++){
+
+    if(!dedupeArray.includes(arr[i])){
+        dedupeArray.push(arr[i])
+    }
+}
+return dedupeArray;
+
+
+}
+module.exports=dedupe;
--- a/Sprint-1/implement/dedupe.test.js
+++ b/Sprint-1/implement/dedupe.test.js
@@ -27,8 +27,8 @@ describe("dedupe()", () => {
   [{ input: [1, 2, 3, 4], expected: [1, 2, 3, 4] }].forEach(
     ({ input, expected }) =>
       it(`should return same input values [${input}] without duplicate`, () => {
-        expect(input).toStrictEqual(expected);
-        expect(input).not.toBe(expected);
+        expect(dedupe(input)).toStrictEqual(expected);
+        expect(dedupe(input)).not.toBe(expected);
       })
   );
 
@@ -63,4 +63,10 @@ it("does not mutate the original array", () => {
   dedupe(input);
 
   expect(input).toStrictEqual([2, 3, 5]);
+});
+
+test("result has only distinct elements", () => {
+  const result = dedupe([1, 1, 2, 2, 3]);
+
+  expect(new Set(result).size).toBe(result.length);
 });
--- a/Sprint-1/implement/dedupe.test.js
+++ b/Sprint-1/implement/dedupe.test.js
@@ -19,7 +19,7 @@ E.g. dedupe([1, 2, 1]) target output: [1, 2]
 describe("dedupe()", () => {
   [{ input: [], expected: [] }].forEach(({ input, expected }) =>
     it(`given an empty array, it returns an empty array [${input}]`, () => {
-      expect(dedupe(input)).toEqual(expected);
+      expect(dedupe(input)).not.toBe(expected);
     })
   );
   // Given an array with no duplicates
@@ -34,7 +34,8 @@ describe("dedupe()", () => {
   ].forEach(({ input, expected }) =>
     it(`should return same input values [${input}] without duplicate`, () => {
       expect(dedupe(input)).toEqual(expected);
-    }));
+    })
+  );
 
   // When passed to the dedupe function
   // Given an array with strings or numbers
@@ -45,12 +46,17 @@ describe("dedupe()", () => {
     { input: [1, 2, 5, 5, "a", 5, 10, 10, "a"], expected: [1, 2, 5, "a", 10] },
     {
       input: ["apple", "banana", "orange", "apple", "banana", 1, 3, 4, 1],
-      expected: ["apple", "banana", "orange", 1, 3,4],
+      expected: ["apple", "banana", "orange", 1, 3, 4],
     },
   ].forEach(({ input, expected }) =>
     it(`should return deduplicated array for [${input}]`, () => {
       expect(dedupe(input)).toEqual(expected);
     })
   );
- 
+
+  [{ input: [1, 1, 2], expected: [1, 1, 2] }].forEach((input, expected) =>
+    it("returns a copy  not the original array", () => {
+      expect(dedupe(input)).not.toBe(expected);
+    })
+  );
 });
--- a/Sprint-1/implement/dedupe.test.js
+++ b/Sprint-1/implement/dedupe.test.js
@@ -19,22 +19,17 @@ E.g. dedupe([1, 2, 1]) target output: [1, 2]
 describe("dedupe()", () => {
   [{ input: [], expected: [] }].forEach(({ input, expected }) =>
     it(`given an empty array, it returns an empty array [${input}]`, () => {
-      expect(dedupe(input)).not.toBe(expected);
+      expect(dedupe(input)).toStrictEqual(expected);
     })
   );
   // Given an array with no duplicates
   // Then it should return a copy of the original array
-  [
-    { input: [1, 2, 3, 4], expected: [1, 2, 3, 4] },
-    {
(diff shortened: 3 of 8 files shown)


Sprint-1/fix/median.js
```diff
@@ -6,9 +6,23 @@
...
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  if (!Array.isArray(list)) return null;
+
+  const numbers = [];
+  for (const x of list) {
+    if (typeof x === "number" && !isNaN(x)) {
+      numbers.push(Number(x));
+    }
+  }
```
cjyuan: The code work.

This is also a good opportunity to practice using the array's `.filter()` method to simplify the code on lines 11-16.

Sprint-1/implement/dedupe.js
```diff
@@ -1 +1,15 @@
...
+for(let i=0;i<arr.length;i++){
+
+    if(!dedupeArray.includes(arr[i])){
+        dedupeArray.push(arr[i])
+    }
+}
+return dedupeArray;
+
+
+}
```
cjyuan: Code is not consistently formatted.

as recommended in 
?

Sprint-1/implement/dedupe.test.js
```diff
@@ -1,28 +1,56 @@
...
+    { input: [1, 2, 3, 4], expected: [1, 2, 3, 4] },
+    {
+      input: ["apples", "banana", "orange"],
+      expected: ["apples", "banana", "orange"],
+    },
+    { input: [-1, 7, 1], expected: [-1, 7, 1] },
+  ].forEach(({ input, expected }) =>
+    it(`should return same input values [${input}] without duplicate`, () => {
+      expect(dedupe(input)).toEqual(expected);
+    }));
```

Sprint-1/implement/max.test.js
```diff
@@ -16,28 +16,73 @@ const findMax = require("./max.js");
 // When passed to the max function
 // Then it should return -Infinity
 // Delete this test.todo and replace it with a test.
-test.todo("given an empty array, returns -Infinity");
+describe("findMax()", () => {
+  [{ input: [], expected: Infinity }].forEach(({ input, expected }) =>
```
cjyuan: According to the spec on line 17, the function is expected to return `-Infinity` when given an empty array.

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,21 @@
 function findMax(elements) {
+  if (!Array.isArray(elements)) return "invalid elements";
+  if (elements.length === 0) return Infinity;
+  const number = [];
+  for (let i = 0; i < elements.length; i++) {
+    if (typeof elements[i] === "number" && !Number.isNaN(elements[i])) {
+      number.push(elements[i]);
+    }
+  }
+  if (number.length === 0) return "invalid elements";
```
cjyuan: Could consider treating both empty array and arrays that contains only non-numeric values as
"arrays that do not contain any number".

Sprint-1/implement/max.test.js
```diff
@@ -16,28 +16,73 @@ const findMax = require("./max.js");
...
+  // Given an array with only non-number values
+  // When passed to the max function
+  // Then it should return the least surprising value given how it behaves for all other inputs
+  [{ input: ["peter", "", "@", "Hi"], expected: "invalid elements" }].forEach(
+    ({ input, expected }) =>
+      it(`should return "invalid elements" for non-numeric values`, () => {
+        expect(findMax(input)).toEqual(expected);
+      })
+  );
+});
```
cjyuan, line 91: When a string representing a valid numeric literal (for example, `"300"`) is compared to a number, 

Sprint-1/implement/sum.test.js
```diff
@@ -10,27 +10,67 @@ const sum = require("./sum.js");
...
+  // Given an array with decimal/float numbers
+  // When passed to the sum function
+  // Then it should return the correct total sum
+
+  [{ input: [0.5, 0.2, 0.11, 0.89, 0.3], expected: 2 }].forEach(
+    ({ input, expected }) =>
+      it(`should return ${expected} for [${input}]`, () => {
+        expect(sum(input)).toEqual(expected);
+      })
+  );
```
cjyuan, line 52: Decimal numbers in most programming languages (including JS) are internally represented in "floating point number" format. Floating point arithmetic is not exact. For example, the result of `46.5678 - 46 === 0.5678` is false because `46.5678 - 46` only yield a value that is very close to `0.5678`. Even changing the order in which the program add/subtract numbers can yield different values.

So the following could happen
```javascript

```

Suggestion: Look up 

Sprint-1/implement/sum.test.js
```diff
@@ -10,27 +10,67 @@ const sum = require("./sum.js");
...
+
+  // Given an array containing non-number values
+  // When passed to the sum function
+  // Then it should ignore the non-numerical values and return the sum of the numerical elements
+
+  [
+    { input: ["evan", 3, "mike", 20, 6, "", "/", , , 20], expected: 49 },
```
cjyuan: Better to explicitly specify `undefined` instead of leaving the element blank.

cjyuan (commented):
Why not practice "committing files one by one, on purpose, and for a reason"?
In VSCode, you can select which file to ***stage***, and commit only the staged file.
See: https://www.youtube.com/watch?v=z5jZ9lrSpqk&t=705  (At around 12:50 minute marker, the video shows how to stage a single file).

Sprint-1/fix/median.js
Ofonimeedak: Thank you, array.filter() used, looks cleaner in just a line

Sprint-1/implement/max.js
Ofonimeedak: array.every() method used to loop through to check if all elements are strings

Ofonimeedak: This helps me handle both empty arrays, because [].every(e => typeof e === "string") is true.

Sprint-1/implement/max.test.js
Ofonimeedak: "300 " and "1.2" added to check for non-numeric values

Sprint-1/implement/sum.test.js
Ofonimeedak: Thank you for this. I have read about the matchers for numbers and floats now

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,21 @@
 function findMax(elements) {
+  if (!Array.isArray(elements)) return "invalid elements";
+  if (elements.every((e) => typeof e === "string")) return -Infinity;
+  const number = [];
+  for (let i = 0; i < elements.length; i++) {
+    if (typeof elements[i] === "number" && !Number.isNaN(elements[i])) {
+      number.push(elements[i]);
+    }
+  }
+  if (number.length === 0) return -Infinity;
```
cjyuan: The code works better without line 3. Can you figure out why?

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,19 @@
...
+    }
+  }
+  if(number.length===0) return "invalid elements";
+  let sumOfNum=0;
+  for(let i=0; i<number.length; i++){
+
+    sumOfNum+=number[i]
+  }
+  return sumOfNum
 }
```
cjyuan, line 16: - Can this function pass all the tests in `sum.test.js`?

- Code format (spacing) is not consistent. Have you successfully enabled "Format on save" and selected "Prettier" as your default JS code formatter?

Sprint-1/implement/dedupe.test.js
```diff
@@ -1,28 +1,62 @@
...
+    {
+      input: ["apples", "banana", "orange"],
+      expected: ["apples", "banana", "orange"],
+    },
+    { input: [-1, 7, 1], expected: [-1, 7, 1] },
+  ].forEach(({ input, expected }) =>
+    it(`should return same input values [${input}] without duplicate`, () => {
+      expect(dedupe(input)).toEqual(expected);
+    })
+  );
```
cjyuan, line 32: Both checks only check one criteria.
(A) When input is an empty array, the test only checks if the return value is a different array but not its content. 

(B) When input contains only distinct elements, the test only checks if the returned array has the correct content but does not check if the array is a different array.

Logically, even if the test in (A) passes, it cannot guarantee the function will return a different array when the input array is not empty.

Sprint-1/implement/sum.js
Ofonimeedak: Yes, it passes all the test cases

cjyuan: Unless you have a different `sum.js` or `sum.test.js` on your computer, I don't see how the current function can pass this test:
```
  [{ input: [], expected: 0 }].forEach(({ input, expected }) =>
    it(`should return ${expected} for [${input}]`, () => {
      expect(sum(input)).toEqual(expected);
    })
  );
```

Sprint-1/implement/dedupe.test.js
```diff
@@ -1,28 +1,62 @@
...
 // Given an empty array
 // When passed to the dedupe function
 // Then it should return an empty array
-test.todo("given an empty array, it returns an empty array");
+describe("dedupe()", () => {
+  [{ input: [], expected: [] }].forEach(({ input, expected }) =>
+    it(`given an empty array, it returns an empty array [${input}]`, () => {
+      expect(dedupe(input)).not.toBe(expected);
+    })
+  );
```
cjyuan, line 24: This check is not checking if the return value is an empty array.

Sprint-1/implement/dedupe.test.js
```diff
@@ -1,28 +1,62 @@
...
+    { input: [1, 2, 3, 4], expected: [1, 2, 3, 4] },
+    {
+      input: ["apples", "banana", "orange"],
+      expected: ["apples", "banana", "orange"],
+    },
+    { input: [-1, 7, 1], expected: [-1, 7, 1] },
+  ].forEach(({ input, expected }) =>
+    it(`should return same input values [${input}] without duplicate`, () => {
+      expect(dedupe(input)).toStrictEqual(expected);
+    })
```
cjyuan: This check cannot check if the function is returning a different array.

Sprint-1/implement/dedupe.test.js
```diff
@@ -1,28 +1,62 @@
...
+    it(`should return deduplicated array for [${input}]`, () => {
+      expect(dedupe(input)).toStrictEqual(expected);
+    })
+  );
+
+  [{ input: [1, 1, 2], expected: [1, 2] }].forEach(({ input, expected }) =>
+    it("returns a copy  not the original array", () => {
+      expect(dedupe(input)).toStrictEqual(expected);
+    })
+  );
```
cjyuan: This check is not checking if the function is returning a copy of the original array or not.

Sprint-1/implement/sum.test.js
cjyuan: I don't see any change made to the relevant Jest test.

Sprint-1/implement/dedupe.test.js
```diff
@@ -1,28 +1,66 @@
...
+  // Given an array with no duplicates
+  // Then it should return a copy of the original array
+  [{ input: [1, 2, 3, 4], expected: [1, 2, 3, 4] }].forEach(
+    ({ input, expected }) =>
+      it(`should return same input values [${input}] without duplicate`, () => {
+        expect(input).toStrictEqual(expected);
+        expect(input).not.toBe(expected);
+      })
+  );
```
cjyuan, line 32: `dedupe()` is not called in this test.

Sprint-1/implement/dedupe.test.js
```diff
@@ -1,28 +1,66 @@
...
+});
+
+test("returns a copy, not the original array", () => {
+  const input = [1, 1, 2];
+  const result = dedupe(input);
+
+  expect(result).toStrictEqual([1, 2]);
+  expect(result).not.toBe(input);
+});
```
cjyuan, line 58: This work. Would be better to test also an array with distinct elements.

Sprint-1/implement/dedupe.test.js
```diff
@@ -1,28 +1,66 @@
...
+
+it("does not mutate the original array", () => {
+  const input = [2, 3, 5];
+  dedupe(input);
+
+  expect(input).toStrictEqual([2, 3, 5]);
+});
```
cjyuan, line 65: Good thinking.

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,20 @@
 function findMax(elements) {
+  if (!Array.isArray(elements)) return "invalid elements";
+  const number = [];
```
cjyuan: Naming convention for arrays is 
- Append `Array` or `List` as a suffix to the name

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,18 @@
 function sum(elements) {
+  if (!Array.isArray(elements)) return "invalid elements";
+  if (elements.length === 0) return 0;
+  const number = [];
+  for (const x of elements) {
+    if (typeof x === "number" && !Number.isNaN(x)) {
+      number.push(x);
+    }
+  }
+  if (number.length === 0) return "invalid elements";
```
cjyuan, line 10: What do you think about designing the function to support this behavior?
   `sum(["A"]) + sum([2]) == sum(["A", 2])`

Ofonimeedak: I think  it will be nice to give it a try, and learn from it

Ofonimeedak: Homomorphism, I will take time to study it in detail, but I would like to complete this and proceed to other backlogs. I am weeks behind. Thank you

Sprint-1/implement/dedupe.test.js
```diff
@@ -27,8 +27,8 @@ describe("dedupe()", () => {
   [{ input: [1, 2, 3, 4], expected: [1, 2, 3, 4] }].forEach(
     ({ input, expected }) =>
       it(`should return same input values [${input}] without duplicate`, () => {
-        expect(input).toStrictEqual(expected);
-        expect(input).not.toBe(expected);
+        expect(dedupe(input)).toStrictEqual(expected);
+        expect(dedupe(input)).not.toBe(expected);
```
cjyuan: What is the purpose of the test on line 31? It will always pass.

cjyuan:
Looks good.

Ofonimeedak:
Thank you, Sir. It's been a long journey....... I am learning well

illicitonion:

PR #1046, opened 2026-03-19
London | ITP-JAN26 | Ihor Taradaiko | Sprint 2 | Coursework/sprint-2

ihortar:
Self checklist

## Changelist

- debug
- implement
- interpret
- stretch

### DIFF
--- a/Sprint-2/debug/recipe.js
+++ b/Sprint-2/debug/recipe.js
@@ -1,8 +1,13 @@
 // Predict and explain first...
+// This will not log the ingredients correctly. It will print the whole
+// recipe object as [object Object] instead of each ingredient on its own line,
+// because you're interpolating the entire object (`${recipe}`) instead of
+// its properties and the ingredients array items.
 
 // This program should log out the title, how many it serves and the ingredients.
 // Each ingredient should be logged on a new line
 // How can you fix it?
+
 
 const recipe = {
   title: "bruschetta",
@@ -11,5 +16,8 @@ const recipe = {
 };
 
 console.log(`${recipe.title} serves ${recipe.serves}
-  ingredients:
-${recipe}`);
+ingredients:`);
+
+for (const ingredient of recipe.ingredients) {
+  console.log(ingredient);
+}
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -50,6 +50,8 @@ test("contains returns false for non-existent property", () => {
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
-test("contains with invalid parameters returns false", () => {
-  expect(contains([], "a")).toBe(false);
+test("contains returns false when first argument is an array even if key exists", () => {
+  const arr = ["a", "b", "c"]; 
+
+  expect(contains(arr, "0")).toBe(false);
 });
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -1,35 +1,55 @@
 const contains = require("./contains.js");
 
+
 /*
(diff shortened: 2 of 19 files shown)


Sprint-2/debug/recipe.js
```diff
@@ -1,15 +1,23 @@
...
 };
 
 console.log(`${recipe.title} serves ${recipe.serves}
-  ingredients:
-${recipe}`);
+ingredients:`);
+
+for (const ingredient of recipe.ingredients) {
+  console.log(ingredient);
+}
```
cjyuan, line 23: Your code works.

Here is an alternative worth exploring: Since ingredient values are separated by '\n' in the output, we could also use `Array.prototype.join()` to construct the equivalent string and then output the resulting string.

Sprint-2/implement/contains.test.js
```diff
@@ -1,35 +1,55 @@
...
+  expect(contains({ a: 1, b: 2 }, "c")).toBe(false);
+});
+
 
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+test("contains with invalid parameters returns false", () => {
+  expect(contains([], "a")).toBe(false);
+});
```
A function that fails to check whether its first parameter is an array could still pass this test:
```javascript
function contains(obj, key) {
}
```

This happens because arrays are objects, with their indices acting as keys. For example, `Object.hasOwn([], "a")` returns `false` simply because `"a"` is not a key of the array.

To properly verify this behavior, the test should use a non-empty array along with a valid key to ensure the function returns `false` specifically because the input is an array, not because the key is missing.

Sprint-2/implement/querystring.js
```diff
@@ -3,14 +3,32 @@ function parseQueryString(queryString) {
...
+    if (indexOfEquals === -1) {
+      // no "=", treat whole pair as key with empty value
+      key = pair;
+      value = "";
+    } else {
+      key = pair.slice(0, indexOfEquals);
+      value = pair.slice(indexOfEquals + 1);
+    }
+
     queryParams[key] = value;
```
cjyuan, line 27: Note: (No change required)

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,15 @@
-function tally() {}
+function tally(items) {
+  if (!Array.isArray(items)) {
+    throw new Error("tally expects an array");
+  }
 
-module.exports = tally;
+  const counts = {};
```
```
```

Sprint-2/interpret/invert.js
```diff
@@ -1,29 +1,49 @@
...
+module.exports = invert;
+
+
 // a) What is the current return value when invert is called with { a : 1 }
+// Answer: { "1": "a" }
+
 
 // b) What is the current return value when invert is called with { a: 1, b: 2 }
+// Answer: { "1": "a", "2": "b" }
+
```
cjyuan, line 31: I think these two questions are asking for the values returned by the original code (and not the expected returned value).

Sprint-2/implement/contains.test.js
ihortar: Adjusted the test

Sprint-2/implement/tally.js
ihortar: Added   const counts = Object.create(null);

Sprint-2/interpret/invert.js
ihortar: Good catch, fixed the answer.

cjyuan:
Changes look good.

Sprint-2/debug/recipe.js
ihortar: It seems more complex at first glance.
I will check out later how it works, as I don't fully understand reading it.
Thanks

illicitonion:

PR #1050, opened 2026-03-19
Manchester | 26-ITP-Jan | Liban Jama | Sprint 2 | Data Groups 

libanj0161:
## Learners, PR Template

Self checklist

## Changelist

Completed Sprint 2 of Module Data Groups.

## Questions

N/A.

### DIFF
--- a/Sprint-2/debug/recipe.js
+++ b/Sprint-2/debug/recipe.js
@@ -12,8 +12,9 @@ const recipe = {
   ingredients: ["olive oil", "tomatoes", "salt", "pepper"],
 };
 
-console.log(`${recipe.title} serves ${recipe.serves}
+console.log(`${recipe.title} 
+  serves ${recipe.serves}
   ingredients:
-${recipe.ingredients.join(",")}`);
+${recipe.ingredients.join("\n")}`);
 
 // it printed [object Object], this is because ${recipe} tries to print the whole object, which shows as [object Object] instead of useful data. This was fixed by using ${recipe.ingredients} (and .join(", ")) to display the ingredients properly.
--- a/Sprint-2/debug/recipe.js
+++ b/Sprint-2/debug/recipe.js
@@ -4,6 +4,8 @@
 // Each ingredient should be logged on a new line
 // How can you fix it?
 
+// it should print the recipe title, how many people people it serves and lists the ingredients.
+
 const recipe = {
   title: "bruschetta",
   serves: 2,
@@ -12,4 +14,6 @@ const recipe = {
 
 console.log(`${recipe.title} serves ${recipe.serves}
   ingredients:
-${recipe}`);
+${recipe.ingredients.join(",")}`);
+
+// it printed [object Object], this is because ${recipe} tries to print the whole object, which shows as [object Object] instead of useful data. This was fixed by using ${recipe.ingredients} (and .join(", ")) to display the ingredients properly.
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,3 +1,5 @@
-function contains() {}
+function contains(object, propertyName) {
+  return propertyName in object;
+}
 
 module.exports = contains;
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,9 +1,9 @@
 function contains(object, propertyName) {
-  if(Object.prototype.toString.call(object) !== "[object Object]") {
-    // if user passes something that is not an object, then throw an error. 
-    throw new Error("not an object")
+  if (Object.prototype.toString.call(object) !== "[object Object]") {
+    // if user passes something that is not an object, then return false.
+    return false;
   }
-    return Object.hasOwn(object, propertyName);
+  return Object.hasOwn(object, propertyName);
 }
 
 module.exports = contains;
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,6 +1,6 @@
 function contains(object, propertyName) {
-  if (Object.prototype.toString.call(object) !== "[object Object]") {
-    // if user passes something that is not an object, then return false.
+  if (typeof object !== "object" || object === null || Array.isArray(object)) {
+    // if user passes something that is not an object, or object is null/array, then return false.
     return false;
   }
   return Object.hasOwn(object, propertyName);
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,7 +1,4 @@
 function contains(object, propertyName) {
-  if (object instanceof Map || object instanceof Set) {
-    return object.has(propertyName);
-  }
   if (typeof object !== "object" || object === null || Array.isArray(object)) {
     // if user passes something that is not an object, or object is null/array, then return false.
     return false;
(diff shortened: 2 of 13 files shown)


Sprint-2/debug/recipe.js
```diff
@@ -12,4 +14,6 @@ const recipe = {
 
 console.log(`${recipe.title} serves ${recipe.serves}
   ingredients:
-${recipe}`);
+${recipe.ingredients.join(",")}`);
```
cjyuan: Can you output each ingredient to a new line (to meet the requirement specified on line 4)?

Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,5 @@
-function contains() {}
+function contains(object, propertyName) {
+  return propertyName in object;
+}
```
cjyuan, line 7: - Consider the following two approaches for determining if an object contains a property:
```
```

---
- In addition, the function should return `false` if the first parameter is an array or is not an object .

Sprint-2/implement/contains.test.js
```diff
@@ -20,16 +20,30 @@ as the object doesn't contains a key of 'c'
...
+test("should return false when the object does not contain the property", () => {
+  expect(contains({ foot: "ball" }, "basket")).toEqual(false);
+});
 
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+test("should return false when input is not an object", () => {
+  expect(contains([], "a")).toEqual(false);
+});
```
cjyuan, line 45: This test does not yet confirm that the function correctly returns false when the first argument is an array. 


Sprint-2/implement/contains.test.js
```diff
@@ -20,16 +20,30 @@ as the object doesn't contains a key of 'c'
...
+
+test("should throw an error when input is not an object", () => {
+  expect(() => contains([], "c")).toThrow();
+});
```
cjyuan: Can your function pass this test?

Sprint-2/implement/lookup.js
```diff
@@ -1,5 +1,13 @@
...
+  // key = country code e.g. 'US'
+  // value = currency code e.g. 'USD'
+  // input = array of [country, currency]
+  // process = go through each pair, add to object
+  // output = object { country: currency }
+  return {
+    US: "USD",
+    CA: "CAD",
+  };
 }
```
cjyuan, line 17: This function is expected to take an argument in the form  
 `[[country1, currency1], [country2, currency2], ..., [countryN, currencyN]]` 
and return 
`{ country1: currency1, country2: currency2, ..., countryN: currencyN }`

Sprint-2/implement/querystring.js
```diff
@@ -1,12 +1,15 @@
...
-  const keyValuePairs = queryString.split("&");
+  if (queryString.length === 0) return queryParams;
+
+  const pairs = queryString.split("&");
+
+  for (const pair of pairs) {
+    const index = pair.indexOf("=");
+
+    const key = pair.slice(0, index).trim();
+    const value = pair.slice(index + 1).trim();
```
cjyuan: For each of the following function calls, does your function return the value you expect?
```
parseQueryString("key=")
parseQueryString("=value")
parseQueryString("key")
```

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,12 @@
-function tally() {}
+function tally(items) {
+  if (!Array.isArray(items)) {
+    throw new Error("Input must be an array");
+  }
+
+  return items.reduce((acc, item) => {
+    acc[item] = (acc[item] || 0) + 1;
+    return acc;
+  }, {});
```
```
```

cjyuan:
Can you also respond to each comment to let me know what you did?

Sprint-2/debug/recipe.js
libanj0161: Used .join to turn array into strings with each item seperated by a new line.

Sprint-2/implement/contains.js
libanj0161: First checked that the argument is an object, if it's not an object then throw an error. Used hasOwn to check the object's own properties. Then fixed the updated test to make sure if an array is passed and a valid propertyName is passed the function still returns false.

Sprint-2/implement/contains.test.js
libanj0161: Updated test to check if an array is passed as the first argument and a valid propertyName is passed as the second argument, the function still returns false.

Sprint-2/implement/contains.test.js
libanj0161: Updated to return false as per instructions.

Sprint-2/implement/lookup.js
libanj0161: Added tests.
Declared variable with empty object, and looped through the array and added each item into the object. Returned new object.

Sprint-2/implement/querystring.js
libanj0161: Added tests for malformed query strings argument.
If query parameter has no key, set value to empty string.
If value is missing, set value to empty string.
If there is no =, set value to empty string. 
If key is missing, ignore.

Sprint-2/implement/tally.js
libanj0161: The function call returned the prototype method.
Created an empty object with no inherited properties.
Added test for tally(["toString", "toString"]);

Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,9 @@
-function contains() {}
+function contains(object, propertyName) {
+  if (Object.prototype.toString.call(object) !== "[object Object]") {
```
cjyuan: This check will reject other kinds of objects such as Map and Set objects.

Can you change it to reject only arrays or any value that is not an object?

cjyuan (commented):
Thanks for the clear explanations.

Changes look good. 

I will mark this PR as complete first.

Sprint-2/implement/contains.js
libanj0161: Added a rejection for null and array, and added test. 
Made contains work with map/set.

Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,12 @@
-function contains() {}
+function contains(object, propertyName) {
+  if (object instanceof Map || object instanceof Set) {
+    return object.has(propertyName);
+  }
```
cjyuan: This is unnecessary because of the check on line 5.

Sprint-2/implement/contains.test.js
```diff
@@ -20,16 +20,46 @@ as the object doesn't contains a key of 'c'
...
+// Given Map as a parameter
+// When passed to contains
+// Then it should return true
+test("should return true when input is map", () => {
+  const obj = new Map();
+  obj.set("name", "John");
+  expect(contains(obj, "name")).toEqual(true);
+});
+
+test("should throw an error when input is an array even if propertyName is a valid key", () => {
```
cjyuan: The function does not throw any error.

Sprint-2/implement/contains.js
libanj0161: removed check for map/set

Sprint-2/implement/contains.test.js
libanj0161: updated test description

illicitonion:

PR #1051, opened 2026-03-20
Manchester | 26-ITP-Jan | Mehroz Munir | Sprint 2 | Coursework

MehrozMunir:
<!--

You must title your PR like this:

Region | Cohort | FirstName LastName | Sprint | Assignment Title

## Learners, PR Template

Self checklist

## Changelist

Completed the exercises with their test cases.

### DIFF
--- a/Sprint-2/debug/recipe.js
+++ b/Sprint-2/debug/recipe.js
@@ -14,6 +14,15 @@ const recipe = {
   ingredients: ["olive oil", "tomatoes", "salt", "pepper"],
 };
 
-console.log(`${recipe.title} serves ${recipe.serves} 
-ingredients:
-${recipe.ingredients[0]}\n${recipe.ingredients[1]}\n${recipe.ingredients[2]}\n${recipe.ingredients[3]}`);
+console.log(
+  `${recipe.title} serves ${recipe.serves} 
+ingredients:\n` + getRecipeIngredients(recipe.ingredients)
+);
+
+function getRecipeIngredients(recipeIngredients) {
+  let ingredients = "";
+  for (const element of recipeIngredients) {
+    ingredients += element + "\n";
+  }
+  return ingredients;
+}
--- a/Sprint-2/debug/recipe.js
+++ b/Sprint-2/debug/recipe.js
@@ -4,12 +4,16 @@
 // Each ingredient should be logged on a new line
 // How can you fix it?
 
+//Ans: The code will log title and serves but all the ingredients will be printed in the same line
+// In order to log each ingredient in a new line we need to access each ingredient
+// and add a new line character after that, right now we are just logging recipe object there
+
 const recipe = {
   title: "bruschetta",
   serves: 2,
   ingredients: ["olive oil", "tomatoes", "salt", "pepper"],
 };
 
-console.log(`${recipe.title} serves ${recipe.serves}
-  ingredients:
-${recipe}`);
+console.log(`${recipe.title} serves ${recipe.serves} 
+ingredients:
+${recipe.ingredients[0]}\n${recipe.ingredients[1]}\n${recipe.ingredients[2]}\n${recipe.ingredients[3]}`);
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -77,14 +77,14 @@ describe("containsProperty", () => {
   // When passed to contains
   // Then it should return false or throw an error
   [
-    { input: [1], property: "a", expected: true },
-    { input: null, property: "c", expected: false },
-    { input: undefined, property: "what", expected: false },
-    { input: [1, 2, 3], property: "what", expected: false },
-  ].forEach(({ input, property, expected }) =>
+    { input: [1], property: "a" },
+    { input: null, property: "c" },
+    { input: undefined, property: "what" },
+    { input: [1, 2, 3], property: "what" },
+  ].forEach(({ input, property }) =>
     it(`Should return false as ${input} is not an object`, () =>
       expect(() => contains(input, property)).toThrow(
-        "It is not a valid object"
+        "Input is not a valid object"
       ))
   );
 });
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -7,29 +7,84 @@ particular property
 E.g. contains({a: 1, b: 2}, 'a') // returns true
 as the object contains a key of 'a'
 
+
 E.g. contains({a: 1, b: 2}, 'c') // returns false
 as the object doesn't contains a key of 'c'
 */
 
-// Acceptance criteria:
+describe("containsProperty", () => {
+  // Given a contains function
+  // When passed an object and a property name
+  // Then it should return true if the object contains the property, false otherwise
+  [
+    { input: { a: 1, b: 2 }, property: "a", expected: true },
+    { input: { a: 1, b: 2 }, property: "c", expected: false },
+    { input: { a: 1, b: 2, sdk: 28299 }, property: "what", expected: false },
+  ].forEach(({ input, property, expected }) =>
+    it(`Should return true if the object ${input} contains the property ${property} otherwise false`, () =>
+      expect(contains(input, property)).toEqual(expected))
+  );
 
-// Given a contains function
-// When passed an object and a property name
-// Then it should return true if the object contains the property, false otherwise
+  // Given an empty object
+  // When passed to contains
+  // Then it should return false
+  [
+    { input: {}, property: "a", expected: false },
+    { input: {}, property: "", expected: false },
+    { input: {}, property: null, expected: false },
+  ].forEach(({ input, property, expected }) =>
+    it(`Should return false if the object ${input} is empty`, () =>
+      expect(contains(input, property)).toEqual(expected))
+  );
 
-// Given an empty object
-// When passed to contains
-// Then it should return false
-test.todo("contains on empty object returns false");
+  // Given an object with properties
+  // When passed to contains with an existing property name
+  // Then it should return true
+  [
+    { input: { a: 1, dog: 2 }, property: "dog", expected: true },
+    {
+      input: { a: 1, b: 2, property3: "here you go" },
+      property: "property3",
+      expected: true,
+    },
+    { input: { a: 1, b: 2, sdk: 28299 }, property: "sdk", expected: true },
+  ].forEach(({ input, property, expected }) =>
+    it(`Should return true if the object ${input} contains the property ${property}`, () =>
+      expect(contains(input, property)).toEqual(expected))
+  );
 
-// Given an object with properties
-// When passed to contains with an existing property name
-// Then it should return true
+  // Given an object with properties
+  // When passed to contains with a non-existent property name
+  // Then it should return false
+  [
+    { input: { a: 1, dog: 2 }, property: "cat", expected: false },
+    {
+      input: { a: 1, b: 2, property3: "here you go" },
+      property: "",
+      expected: false,
+    },
+    {
+      input: { a: 1, b: 2, thatProperty: 28299 },
+      property: "",
+      expected: false,
+    },
+  ].forEach(({ input, property, expected }) =>
+    it(`Should return false if the object ${input} doesn't contain the property ${property}`, () =>
+      expect(contains(input, property)).toEqual(expected))
+  );
 
-// Given an object with properties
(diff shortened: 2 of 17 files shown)


Sprint-2/debug/recipe.js
```diff
@@ -4,12 +4,16 @@
...
   serves: 2,
   ingredients: ["olive oil", "tomatoes", "salt", "pepper"],
 };
 
-console.log(`${recipe.title} serves ${recipe.serves}
-  ingredients:
-${recipe}`);
+console.log(`${recipe.title} serves ${recipe.serves} 
+ingredients:
+${recipe.ingredients[0]}\n${recipe.ingredients[1]}\n${recipe.ingredients[2]}\n${recipe.ingredients[3]}`);
```
cjyuan: Can you use an approach that can work for any number of ingredients?

Suggestion: A for-of loop or `Array.prototype.join()`.

Sprint-2/implement/contains.test.js
```diff
@@ -7,29 +7,84 @@ particular property
...
+  );
 
-// Given invalid parameters like an array
-// When passed to contains
-// Then it should return false or throw an error
+  // Given invalid parameters like an array
+  // When passed to contains
+  // Then it should return false or throw an error
+  [
+    { input: [1], property: "a" },
```
cjyuan, line 80: This test does not yet confirm that the function correctly returns false when the first argument is an array.  
A function that fails to check whether its first parameter is an array could still pass this test:
```javascript
function contains(obj, key) {
}
```

This happens because arrays are objects, with their indices acting as keys. For example, `Object.hasOwn([1], "a")` 
returns `false` simply because `"a"` is not a key of the array.

To properly verify this behavior, the test should use a non-empty array along with a valid key to 
ensure the function returns `false` specifically because the input is an array, not because the key is missing.

Sprint-2/implement/lookup.test.js
```diff
@@ -33,3 +31,41 @@ It should return:
...
+
+  // CASE 1: Should return an object with key as country codes and value
+  // as currency codes when an array is passed containing arrays of countries
+  // codes and their currencies codes
+  test(`should return an object of the array [${countryCurrencyArray}]`, () =>
+    expect(createLookup(countryCurrencyArray)).toEqual(countryCurrencyObject));
+
+  // CASE 2: Should return an empty object if empty array is passed
+  test(`should return an empty object if the array is empty`, () =>
+    expect(createLookup([[]])).toEqual({}));
```
cjyuan: The first argument is not quite an empty array; it is "an array containing an empty array".

For this input, your function actually returns `{ undefined: undefined }` instead of `{}`, but `toEqual()` consider them equal because it ignores property named "undefined". A stricter comparison `.toStrictEqual()` would have flagged them as an error.

Sprint-2/implement/querystring.js
```diff
@@ -1,12 +1,13 @@
...
   if (queryString.length === 0) {
     return queryParams;
   }
   const keyValuePairs = queryString.split("&");
 
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
+    const [key, value] = pair.split(/=(.*)/);
     queryParams[key] = value;
   }
```
cjyuan, line 12: (1) For the following function call, does your function return the value you expect?
```
parseQueryString("a=b&&c=d")
```
---

(2) Can you explain how `pair.split(/=(.*)/)` works?

Sprint-2/stretch/count-words.js
```diff
@@ -26,3 +26,28 @@
...
+  const wordsCountObject = {};
+
+  // Replace anything that is NOT a word character or a space with an empty string
+  const cleanText = string.replace(/[^\w\s]|_/g, "").toLowerCase();
+
+  cleanText.split(" ").forEach((word) => {
+    if (Object.hasOwn(wordsCountObject, word)) wordsCountObject[word] += 1;
+    else wordsCountObject[word] = 1;
+  });
+
```
cjyuan: Can you check if your function returns what you expect in the following function calls?

```
countWords("          Hello   World      ");
```

Sprint-2/stretch/mode.js
```diff
@@ -9,28 +9,20 @@
...
+    if (Object.hasOwn(frequencyObject, number)) frequencyObject[number] += 1;
+    else frequencyObject[number] = 1;
   }
 
-  return maxFreq === 0 ? NaN : mode;
+  //sorting the number with the highest frequency at the lowest index
+  const sortedArray = Object.entries(frequencyObject).sort(
+    (a, b) => b[1] - a[1]
+  );
+  return Number(sortedArray[0][0]);
```
cjyuan: The purpose of this exercise is to practice refactoring code in a "large" function into smaller functions.
The objective is to separate the original code into two functions:
Function 1 -- when given an array, returns the frequency of the element in a Map object.
Function 2 -- when given a Map object representing the frequency of of elements, return the mode.

Then, rewrite the original function as a composition of these two functions.

Sprint-2/stretch/till.js
```diff
@@ -8,7 +8,7 @@ function totalTill(till) {
   let total = 0;
 
   for (const [coin, quantity] of Object.entries(till)) {
-    total += coin * quantity;
+    total += Number(coin.slice(0, coin.length - 1)) * quantity;
```
cjyuan: We could also use `-1` as the "index of the last character" in `.slice()` -- shorter syntax.

Sprint-2/debug/recipe.js
MehrozMunir: Yes thank you for the suggestion. I have implemented it now.

Sprint-2/implement/contains.test.js
MehrozMunir: I have looked at my function, and it contains an extra condition in if statement as shown below in the contains.js file

`function contains(object, property) {
  if (object === null || typeof object !== "object" || Array.isArray(object))
    throw new Error("Input is not a valid object");
  else return Object.hasOwn(object, property);
}`

The extra condition of Array.isArray(object) checks if the item passes is an array it gonna throw an error, and the else condition is not even going to run for the arrays

But, I understand your point, and for testing purposes, it is a good approach to add the test case that truly ensures that if the passed object is an array with the right property of that array, even then the test should pass, therefore, I have added the following test case value 

`{ input: [1], property: "0" },`

If you think that I have not understood your point correctly, please let me know, and I am ready to make any changes. Thanks

Sprint-2/implement/lookup.test.js
MehrozMunir: Thanks. 
Yes that is absolutely right. Do you think that I should always use toStrictEqual in all the test cases?

Sprint-2/implement/querystring.js
MehrozMunir: 1. yes it does I have added this in test case 4 as follows:

`//Case 4 : extra & in the start, mid or at the end
test("parses querystring containing extra & in the start, mid or at the end", () => {
  expect(parseQueryString("&equation=x=y+1&&job=&")).toEqual({
    equation: "x=y+1",
    job: "",
  });
  expect(parseQueryString("a=b&&c=d")).toEqual({
    a: "b",
    c: "d",
  });
});`


2. As we already know that we can split the string into an array by providing a delimiter, the reason for writing it in a regular expression is because we want the split method to look for only first = sign and divdes the string before and after it ignoring any other equal signs coming after it till the end and that is done by the use of  .* in our regular expresson.

Sprint-2/stretch/count-words.js
MehrozMunir: Actually no.
I expected it to return me {hello:1, word:1} but it returned me the following 
{ '': 18, hello: 1, world: 1 }

And I need to correct that in my code because right now i am not trimming the extra spaces before and after and in the middle of words.
I have added that now.

Thanks

Sprint-2/stretch/mode.js
MehrozMunir: Thank you for clarifying. I should have read that properly. Thanks

Sprint-2/stretch/till.js
MehrozMunir: Oh Yes. Thanks

MehrozMunir:
@cjyuan  
Thank you for the feedback. I have implemented the changes and pushed the latest code. Kindly check, please.

Sprint-2/debug/recipe.js
```diff
@@ -4,12 +4,25 @@
...
+ingredients:\n` + getRecipeIngredients(recipe.ingredients)
+);
+
+function getRecipeIngredients(recipeIngredients) {
+  let ingredients = "";
+  for (const element of recipeIngredients) {
+    ingredients += element + "\n";
+  }
+  return ingredients;
+}
```
cjyuan, line 28: Note: This function can be replaced by a single function call to `Array.prototype.join()`. 

No change required but that array's method is worth exploring.

Sprint-2/implement/contains.test.js
```diff
@@ -7,29 +7,85 @@ particular property
...
+    { input: [1], property: "0" },
```
cjyuan, line 81: My concern was just about the test. 

We write tests not only to verify our current implementation, but also to ensure that future changes do not alter the function's expected behavior.

So after adding a test such as
> `{ input: [1], property: "0" },`

you can help ensure future modification of the function can still correctly reject array.

Sprint-2/stretch/mode.js
```diff
@@ -9,28 +9,34 @@
...
+function getFrequencyMapOfElements(list) {
+  if (!Array.isArray(list)) throw new Error("Not an array");
+  //saving the frequency of each element in key value pair using map object
+  const frequencyMap = new Map();
+  for (const number of list) {
+    if (typeof number !== "number") continue;
+    let currentNumberFrequency = frequencyMap.get(number);
+    if (currentNumberFrequency !== undefined)
+      frequencyMap.set(number, (currentNumberFrequency += 1));
+    else frequencyMap.set(number, 1);
```
cjyuan: - Could just use `+` instead of `+=` on line 24.

- Why not use the original syntax to implement the code on lines 22-25?
> frequencyMap.set(number, (frequencyMap.get(number) || 0) + 1);

Sprint-2/stretch/mode.js
```diff
@@ -9,28 +9,34 @@
...
+
+  for (let [key, value] of map) {
+    if (value > maxValue) {
+      maxValue = value;
+      maxKey = key;
     }
   }
-
-  return maxFreq === 0 ? NaN : mode;
+  return maxKey;
```
cjyuan: The original code returns `NaN` instead of `null` when there is no mode.

Sprint-2/stretch/count-words.js
```diff
@@ -26,3 +26,33 @@
...
+  const wordsCountObject = {};
+
+  // Replace anything that is NOT a word character or a space with an empty string
+  const cleanText = string
+    .replace(/[^\w\s]|_/g, "")
+    .toLowerCase()
+    .trim()
+    .replace(/\s+/g, " ");
+
+  cleanText.split(" ").forEach((word) => {
```
cjyuan: Note: We can also use regex with `.split()`, and combine the operation on line 43 and 45 as:

```
  cleanText.split(/\s+/).forEach((word) => {
```

Sprint-2/implement/lookup.test.js
cjyuan: Yes. I think `.toStrictEqual()` is better when comparing objects.

Sprint-2/debug/recipe.js
MehrozMunir: Thank you. I have read about join() method. Yes, the function getRecipeIngredients could be replaced by this method.

Sprint-2/stretch/mode.js
MehrozMunir: Yes right. I should use that. It is much simpler and concise.

Sprint-2/stretch/mode.js
MehrozMunir: Oh Yes. I just deleted the original code when I was making changes and lost track of that. Should have used this in the first place.

MehrozMunir:
@cjyuan  Thank you for the feedback. I have made changes in mode.js, mode.test.js, and count-words.js files and pushed the latest code after your feedback.

cjyuan:
Changes look good.

illicitonion:

PR #1053, opened 2026-03-20
Sheffield | 26-Jan-ITP | Daniel Aderibigbe | Sprint 2 | Coursework/sprint 2

Dan2Clouted:
## Learners, PR Template

Self checklist

## Changelist

Completed all Exercises.

### DIFF
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,4 +1,8 @@
 function contains(object, property) {
+  if (object === null || typeof object !== "object" || Array.isArray(object)) {
+    return false;
+  }
+
   return Object.prototype.hasOwnProperty.call(object, property);
 }
 
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,3 +1,5 @@
-function contains() {}
+function contains(object, property) {
+  return Object.prototype.hasOwnProperty.call(object, property);
+}
 
 module.exports = contains;
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -51,3 +51,7 @@ test("contains an object with non-existent property returns false", () => {
 test("contains with invalid parameters returns false", () => {
   expect(contains([], "a")).toBe(false);
 });
+
+test("contains with invalid parameters returns false", () => {
+  expect(contains([1, 2, 3], "1")).toBe(false);
+});
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -1,35 +1,53 @@
 const contains = require("./contains.js");
 
 /*
-Implement a function called contains that checks an object contains a
-particular property
+  Implement a function called contains that checks an object contains a
+  particular property
 
-E.g. contains({a: 1, b: 2}, 'a') // returns true
-as the object contains a key of 'a'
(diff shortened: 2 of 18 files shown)


Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,5 @@
-function contains() {}
+function contains(object, property) {
+  return Object.prototype.hasOwnProperty.call(object, property);
```
cjyuan, line 6: - The function is expected to throw an error or return `false` when object is not an object or is an array.

- Could also explore `Object.hasOwn()`.

Sprint-2/implement/contains.test.js
```diff
@@ -1,35 +1,53 @@
...
+  const obj = { a: 1, b: 2 };
+  expect(contains(obj, "c")).toBe(false);
+});
 
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+test("contains with invalid parameters returns false", () => {
+  expect(contains([], "a")).toBe(false);
+});
```
cjyuan, line 53: `expect(contains([], 'a')).toBe(false)` may fail to verify **whether
`contains()` can detect its first argument is an array**.

`contains([], "a")` could also return false because "a" is not a property (or key) of an array.
However, "0", "1", "2" are keys of `[1, 2, 3]`, so it is better to specify the test as
`expect(contains([1, 2, 3], "1")).toBe(false);` (to ensure you are checking what you describe).

You could also add a few more samples to ensure the function can properly handle arguments of different types.

Sprint-2/implement/querystring.js
```diff
@@ -1,13 +1,22 @@
...
-    queryParams[key] = value;
+    const indexOfEquals = pair.indexOf("=");
+
+    if (indexOfEquals === -1) {
+      queryParams[pair] = "";
+    } else {
+      const key = pair.slice(0, indexOfEquals);
+      const value = pair.slice(indexOfEquals + 1);
+      queryParams[key] = value;
+    }
```
cjyuan, line 23: For each of the following function calls, does your function return the value you expect?
```
parseQueryString("a=b&=&c=d")
parseQueryString("a=")
parseQueryString("=b")
parseQueryString("a=b&&c=d")
```

Note: (You don't have to implement this)

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,19 @@
...
+
+  const result = {};
+
+  for (const item of items) {
+    if (result[item]) {
+      result[item]++;
+    } else {
+      result[item] = 1;
+    }
+  }
```
```
```

Sprint-2/stretch/count-words.js
```diff
@@ -26,3 +26,24 @@
...
+  for (const word of words) {
+    if (result[word]) {
+      result[word]++;
+    } else {
+      result[word] = 1;
+    }
+  }
+
+  return result;
+}
```
cjyuan, line 52: Can you check if your function returns what you expect in the following function calls?

```
```

Dan2Clouted:
Thanks I’ve updated the implementation and tests to better cover the edge cases you pointed out. I handled invalid object inputs for `contains` added extra edge-case checks for `parseQueryString`, used objects without inherited properties for `tally`, and improved `countWords` to cope better with punctuation, repeated spaces, and special property names.

cjyuan:
LGTM.

illicitonion:

PR #1054, opened 2026-03-20
Sheffield | 26-Jan-ITP | Martha Ogunbiyi| Sprint 1 | Coursework/sprint 1

marthak1:
## Learners, PR Template

Self checklist

## Changelist

implemented sprint 1 course work

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -18,7 +18,7 @@ function calculateMedian(list) {
   if (numericValues.length === 0) return null;
 
   // Step 4: sort safely
-  const sortedList = numericValues.slice().sort((a, b) => a - b);
+  const sortedList = numericValues.sort((a, b) => a - b);
 
   const middleIndex = Math.floor(sortedList.length / 2);
 
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -6,9 +6,28 @@
 // or 'list' has mixed values (the function is expected to sort only numbers).
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  // Step 1: must be an array
+  if (!Array.isArray(list)) return null;
+
+  // Step 2: filter numeric values
+  const numericValues = list.filter(
+    (item) => typeof item === "number" && !Number.isNaN(item)
+  );
+
+  // Step 3: must have at least one number
+  if (numericValues.length === 0) return null;
+
+  // Step 4: sort safely
+  const sortedList = numericValues.slice().sort((a, b) => a - b);
+
+  const middleIndex = Math.floor(sortedList.length / 2);
+
+  // Step 5: median
+  if (sortedList.length % 2 !== 0) {
+    return sortedList[middleIndex];
+  }
+
+  return (sortedList[middleIndex - 1] + sortedList[middleIndex]) / 2;
 }
 
 module.exports = calculateMedian;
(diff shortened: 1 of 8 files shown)


Sprint-1/fix/median.js
```diff
@@ -6,9 +6,28 @@
...
+  // Step 2: filter numeric values
+  const numericValues = list.filter(
+    (item) => typeof item === "number" && !Number.isNaN(item)
+  );
+
+  // Step 3: must have at least one number
+  if (numericValues.length === 0) return null;
+
+  // Step 4: sort safely
+  const sortedList = numericValues.slice().sort((a, b) => a - b);
```
cjyuan: Why make a copy of `numericValues` before sorting?

Sprint-1/implement/dedupe.test.js
```diff
@@ -6,23 +6,41 @@ Dedupe Array
...
+});
 
 // Given an array with no duplicates
 // When passed to the dedupe function
 // Then it should return a copy of the original array
+test("given an array with no duplicates, it returns return a copy of the original array", () => {
+  const input = [1, 2, 3, 4, 5];
+  const result = dedupe(input);
+  expect(result).toEqual([1, 2, 3, 4, 5]);
+});
```

Sprint-1/implement/max.test.js
```diff
@@ -16,28 +16,62 @@ const findMax = require("./max.js");
...
+});
 
 // Given an array with only non-number values
 // When passed to the max function
 // Then it should return the least surprising value given how it behaves for all other inputs
+test("given an array with only non-number value, it should return the least surprising value given how it behaves for all other inputs", () => {
+  const input = [true, {}, "arr", null, [], " ", NaN];
+  const result = findMax(input);
+  expect(result).toBe(-Infinity);
+});
```
cjyuan, line 77: When a string representing a valid numeric literal (for example, `"300"`) is compared to a number, 

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,53 @@ const sum = require("./sum.js");
...
+});
 
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
+test("given an array with decimal/float numbers, it should return the correct total sum", () => {
+  const input = [5, 6.6, 10, 5.12];
+  const result = sum(input);
+  expect(result).toBe(26.720000000000002);
+});
```
cjyuan, line 47: Decimal numbers in most programming languages (including JS) are internally represented in "floating point number" format. Floating point arithmetic is not exact. For example, the result of `46.5678 - 46 === 0.5678` is false because `46.5678 - 46` only yield a value that is very close to `0.5678`. Even changing the order in which the program add/subtract numbers can yield different values.

So the following could happen
```javascript

```

Suggestion: Look up 

Sprint-1/refactor/includes.js
```diff
@@ -11,3 +10,10 @@ function includes(list, target) {
 }
 
 module.exports = includes;
+
+// const array = ["a", "b", "c"];
+
+// for (const element of array) {
+//   console.log(element);
+// } for (variable of iterable) statement
+
```
cjyuan, line 14: It's best practice to remove unnecessary code to keep the code clean.

Sprint-1/fix/median.js
marthak1: Thanks for the feedback and apologies in delayed response.  Thanks for pointing this out. Creating a second copy wasn't necessary when filter() already creates a new array.

Sprint-1/implement/dedupe.test.js
marthak1: Adding reference checks => expect(result).not.toBe(input); This checks if NOT the same array in memory

Sprint-1/implement/max.test.js
marthak1: Thanks for the feedback. I have refactored the implementation to not only check the 'type' ,  but to also  check whether it can become a valid number

cjyuan:
Changes look good. Well done.

illicitonion:

PR #1057, opened 2026-03-20
London|26-ITP-January|Alexandru Pocovnicu|Sprint 3| alarm clock

alexandru-pocovnicu:
## Learners, PR Template

Self checklist

## Changelist

implemented an alarm clock that counts down from the input value

### DIFF
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -3,7 +3,11 @@ function setAlarm() {
   const alarmSetEl = document.getElementById("alarmSet");
   const timeRemainingEl = document.getElementById("timeRemaining");
   let totalSeconds = +alarmSetEl.value;
-  if (totalSeconds === 0) {
+  if (
+    totalSeconds <= 0 ||
+    isNaN(totalSeconds) ||
+    !Number.isInteger(totalSeconds)
+  ) {
     alarmSetEl.value = null;
     return;
   }
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -1,4 +1,11 @@
-function setAlarm() {}
+function setAlarm() {
+ const alarmSetEl=document.getElementById("alarmSet")
+ const timeRemainingEl = document.getElementById("timeRemaining");
+ timeRemainingEl.innerHTML=`Time Remaining: ${alarmSetEl.value}`
+ console.log(alarmSetEl.value);
+ 
+
+}
 
 // DO NOT EDIT BELOW HERE
 
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -4,9 +4,8 @@ let remainingSeconds = 0;
 function setAlarm() {
   isPaused = false;
   const pauseButton = document.getElementById("pause-button");
-  if (pauseButton) {
+    pauseButton.textContent = "Pause";
     pauseButton.addEventListener("click", pauseCountDown);
-  }
 
   const alarmSetEl = document.getElementById("alarmSet");
   const timeRemainingEl = document.getElementById("timeRemaining");
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -26,12 +26,8 @@ function setAlarm() {
   cleanInitialState();
 
   function updateCountDown() {
-    let seconds = totalSeconds % 60;
-    let minutes = (totalSeconds - seconds) / 60;
-
-    let paddedSeconds = seconds.toString().padStart(2, "0");
-    let paddedMinutes = minutes.toString().padStart(2, "0");
-    timeRemainingEl.innerHTML = `Time Remaining: ${paddedMinutes}:${paddedSeconds}`;
+    remainingSeconds = totalSeconds;
+    updateDisplayedTime(remainingSeconds);
 
     if (totalSeconds <= 0) {
       document.body.classList.add("finish-countdown");
@@ -62,12 +58,7 @@ function pauseCountDown(e) {
         return;
       }
       remainingSeconds -= 1;
-      const seconds = remainingSeconds % 60;
-      const minutes = (remainingSeconds - seconds) / 60;
-      const paddedSeconds = seconds.toString().padStart(2, "0");
-      const paddedMinutes = minutes.toString().padStart(2, "0");
-      document.getElementById("timeRemaining").innerHTML =
-        `Time Remaining: ${paddedMinutes}:${paddedSeconds}`;
+      updateDisplayedTime(remainingSeconds);
 
       if (remainingSeconds <= 0) {
         document.body.classList.add("finish-countdown");
@@ -84,6 +75,16 @@ function cleanInitialState() {
   document.getElementById("timeRemaining").innerHTML = "Time Remaining: 00:00";
   document.getElementById("alarmSet").value = null;
   pauseAlarm();
+}
+
+function updateDisplayedTime(totalSeconds) {
+  let seconds = totalSeconds % 60;
+  let minutes = (totalSeconds - seconds) / 60;
+
+  let paddedSeconds = seconds.toString().padStart(2, "0");
+  let paddedMinutes = minutes.toString().padStart(2, "0");
+  document.getElementById("timeRemaining").innerHTML =
+    `Time Remaining: ${paddedMinutes}:${paddedSeconds}`;
 }
 
 // DO NOT EDIT BELOW HERE
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -64,7 +64,7 @@ function pauseCountDown(e) {
 function cleanInitialState() {
   clearInterval(intervalId);
   document.body.classList.remove("finish-countdown");
-  document.getElementById("timeRemaining").innerHTML = "Time Remaining: 00:00";
+  updateDisplayedTime(0)//document.getElementById("timeRemaining").innerHTML = "Time Remaining: 00:00";
   document.getElementById("alarmSet").value = null;
   pauseAlarm();
 }
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -1,9 +1,21 @@
 let intervalId;
 let isPaused;
 let remainingSeconds = 0;
+
+function updateCountDown() {
+  updateDisplayedTime(remainingSeconds);
+
+  if (remainingSeconds <= 0) {
+    finishCountDown();
+    return;
+  }
+
+  remainingSeconds -= 1;
+}
+
 function setAlarm() {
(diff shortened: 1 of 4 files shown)


Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,33 @@
-function setAlarm() {}
+function setAlarm() {
+  const alarmSetEl = document.getElementById("alarmSet");
+  const timeRemainingEl = document.getElementById("timeRemaining");
+  let totalSeconds = +alarmSetEl.value;
```
cjyuan: Some unusual input values can make your app behave abnormally. Can you add code to sanitise them?

cjyuan (commented):
Can you fix the issue?

---

alexandru-pocovnicu:
when i press stopalarm i want the countdown to go to 0,call pauseAlarm(), and the class to be removed from the body,but on line 48 i have this message // DO NOT EDIT BELOW HERE, I can't really think of a way of doing it without adding code bellow that

Sprint-3/alarmclock/alarmclock.js
alexandru-pocovnicu: done

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,49 @@
...
+  if (
+    totalSeconds <= 0 ||
+    isNaN(totalSeconds) ||
+    !Number.isInteger(totalSeconds)
+  ) {
+    alarmSetEl.value = null;
```
cjyuan: `.value` property is of type "string". Setting its value to `""` is better.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,49 @@
...
+  cleanInitialState();
+
+  function updateCountDown() {
+    let seconds = totalSeconds % 60;
+    let minutes = (totalSeconds - seconds) / 60;
+
+    let paddedSeconds = seconds.toString().padStart(2, "0");
+    let paddedMinutes = minutes.toString().padStart(2, "0");
+    timeRemainingEl.innerHTML = `Time Remaining: ${paddedMinutes}:${paddedSeconds}`;
+    alarmSetEl.value = null;
```
cjyuan: Why clear the input field when updating the timer display? When a countdown is active, it will be cleared every second.

cjyuan (commented):
Note: To reset countdown display when the user clicks "Stop", and without modifying existing code, you can add another event listener to the stop button:
```
    // Your code here ...
  });
```

Sprint-3/alarmclock/alarmclock.js
alexandru-pocovnicu: didn't realise it was doing it every second , now there is a dedicated function to clear everything when the alarm is set

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,91 @@
...
+let isPaused;
+let remainingSeconds = 0;
+function setAlarm() {
+  isPaused = false;
+  const pauseButton = document.getElementById("pause-button");
+  if (pauseButton) {
+    pauseButton.removeAttribute("id");
+    pauseButton.classList.add("unpause-alarm");
+    pauseButton.addEventListener("click", pauseCountDown);
+  }
```
cjyuan: This is an unusual way to perform the tasks once (or to make the pause button visible when the "Start alarm" button is clicked the first time). Why not perform the task once on page load?

Currently `unpause-alarm` class does not have any style defined in `style.css`. (Work in progress?)

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,91 @@
...
+    isNaN(totalSeconds) ||
+    !Number.isInteger(totalSeconds)
+  ) {
+    alarmSetEl.value = "";
+    return;
+  }
+  cleanInitialState();
+
+  function updateCountDown() {
+    remainingSeconds = totalSeconds;
```
cjyuan: You have already assigned `totalSeconds` to `remainingSeconds` on line 16, why not just use only `remainingSeconds` in the remaining part of the function?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,91 @@
...
+    let seconds = totalSeconds % 60;
+    let minutes = (totalSeconds - seconds) / 60;
+
+    let paddedSeconds = seconds.toString().padStart(2, "0");
+    let paddedMinutes = minutes.toString().padStart(2, "0");
+    timeRemainingEl.innerHTML = `Time Remaining: ${paddedMinutes}:${paddedSeconds}`;
```
cjyuan: Better to implement the logic to "update the displayed time" in a function, and then use a function call to replace the code on lines 30-35, 66-71, 85. 

Pros:
- Centralised logic (improved maintainability)
- Less code

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,91 @@
...
+      const paddedMinutes = minutes.toString().padStart(2, "0");
+      document.getElementById("timeRemaining").innerHTML =
+        `Time Remaining: ${paddedMinutes}:${paddedSeconds}`;
+
+      if (remainingSeconds <= 0) {
+        document.body.classList.add("finish-countdown");
+        playAlarm();
+        clearInterval(intervalId);
+      }
+    }, 1000);
```
cjyuan: It seems the code here is very similar to those perform in `setAlarm()`. Can you try reusing the code you have in setAlarm()?

Sprint-3/alarmclock/alarmclock.js
alexandru-pocovnicu: i was thinking to use it , since i was removing the id I needed something to work with, i removed both from css

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,89 @@
...
+        finishCountDown();
+      }
+    }, 1000);
+  }
+}
+
+function cleanInitialState() {
+  clearInterval(intervalId);
+  document.body.classList.remove("finish-countdown");
+  document.getElementById("timeRemaining").innerHTML = "Time Remaining: 00:00";
```
cjyuan: Could also call `updateDiplayedTime(0)` to keep the logic of displaying time in one place.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,89 @@
-function setAlarm() {}
+let intervalId;
+let isPaused;
+let remainingSeconds = 0;
+function setAlarm() {
+  isPaused = false;
+  const pauseButton = document.getElementById("pause-button");
+    pauseButton.textContent = "Pause";
+    pauseButton.addEventListener("click", pauseCountDown);
```
cjyuan: You can reset the `isPaused` flag and the "Pause" button state in `cleanInitialState()`, and add event listener only once on page load.

Currently, the same event listener will be added to the button multiple times (once per function call).

alexandru-pocovnicu: but if i move the listener inside cleanInitialState(), which is called everytime setAlarm() is called, isn't that the same thing, how about instead, i remove the listener inside setAlarm just before adding it?
thank you

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,91 @@
...
+    }, 1000);
+  }
+}
+
+function cleanInitialState() {
+  isPaused = false;
+  clearInterval(intervalId);
+  document.body.classList.remove("finish-countdown");
+  updateDisplayedTime(0)
+  document.getElementById("alarmSet").value = null;
```
cjyuan: `.value` is a string, so better to write `document.getElementById("alarmSet").value = "";`.

Sprint-3/alarmclock/alarmclock.js
cjyuan: I mean, perform this in `cleanInitialState()`
```
    const pauseButton = document.getElementById("pause-button");
    pauseButton.textContent = "Pause";
```

and perform this on page load
```
window.addEventListener("load", function() {
    const pauseButton = document.getElementById("pause-button");
    pauseButton.addEventListener("click", pauseCountDown);
});
```

Sprint-3/alarmclock/alarmclock.js
alexandru-pocovnicu: removed the one on line 16

Sprint-3/alarmclock/alarmclock.js
alexandru-pocovnicu: I'm sorry but I am unable to do it

Sprint-3/alarmclock/alarmclock.js
alexandru-pocovnicu: ok, got it, I was under the impression that because there was already window.onload , I couldn't add anything to the window anymore, thank you

Sprint-3/alarmclock/alarmclock.js
alexandru-pocovnicu: done

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,90 @@
...
+    if (totalSeconds <= 0) {
+      finishCountDown();
+      return;
+    }
+
+    totalSeconds -= 1;
+  }
+
+  updateCountDown();
+  intervalId = setInterval(updateCountDown, 1000);
```
cjyuan, line 33: 1. Move `remainingSeconds = totalSeconds;` out of `updateCountDown()` and replace `totalSeconds` in the function by `remainingSeconds` so that the function depends only on `remainingSeconds`. 

2. Move `updateCountDown()` to outer scope so that it can be used by `pauseCountDown()`.

That is,
```javascript
function updateCountDown() {
   ...
}

function setAlarm() {
  // Code to deal with input
  ...

  remainingSeconds = totalSeconds;  // Initialise the global countdown number

  // Display the timer and start the countdown
  updateCountDown();
  intervalId = setInterval(updateCountDown, 1000);
}
```

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,90 @@
...
+        clearInterval(intervalId);
+        return;
+      }
+      remainingSeconds -= 1;
+      updateDisplayedTime(remainingSeconds);
+
+      if (remainingSeconds <= 0) {
+        finishCountDown();
+      }
+    }, 1000);
```
cjyuan: This is supposed to be "continuing the countdown". So it can be replaced by

```javascript
  updateCountDown();
  intervalId = setInterval(updateCountDown, 1000);
```

---
Note: Currently, `finishCountDown()` is not called when `remainingSeconds` reaches zero.

cjyuan (commented):
Other changes look good.

Sprint-3/alarmclock/alarmclock.js
alexandru-pocovnicu: done

cjyuan:
Looks good.

illicitonion:

PR #1061, opened 2026-03-21
Sheffield | 26-ITP-Jan | Karim M'hamdi | Sprint 2 | Module-Data-Groups

KKtech06:
## Learners, PR Template

Self checklist

## Questions

N/A

### DIFF
--- a/Sprint-2/debug/recipe.js
+++ b/Sprint-2/debug/recipe.js
@@ -13,6 +13,5 @@ const recipe = {
 console.log(recipe.title);
 console.log(`Serves: ${recipe.serves}`);
 console.log("Ingredients:");
-for (const ingredient of recipe.ingredients) {
-  console.log(ingredient);
-}
+console.log(recipe.ingredients.join("\n"));
+
--- a/Sprint-2/debug/recipe.js
+++ b/Sprint-2/debug/recipe.js
@@ -10,6 +10,9 @@ const recipe = {
   ingredients: ["olive oil", "tomatoes", "salt", "pepper"],
 };
 
-console.log(`${recipe.title} serves ${recipe.serves}
-  ingredients:
-${recipe}`);
+console.log(recipe.title);
+console.log(`Serves: ${recipe.serves}`);
+console.log("Ingredients:");
+for (const ingredient of recipe.ingredients) {
+  console.log(ingredient);
+}
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -2,9 +2,6 @@ function contains(obj, prop) {
   if (typeof obj !== "object" || obj === null) {
     return false;
   }
-  if (typeof prop !== "string") {
-    return false;
-  }
 
   if (Array.isArray(obj)) {
     return false;
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,3 +1,16 @@
-function contains() {}
+function contains(obj, prop) {
+  if (typeof obj !== "object" || obj === null) {
+    return false;
+  }
+  if (typeof prop !== "string") {
+    return false;
+  }
+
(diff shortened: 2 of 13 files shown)


Sprint-2/debug/recipe.js
```diff
@@ -10,6 +10,9 @@ const recipe = {
...
 
-console.log(`${recipe.title} serves ${recipe.serves}
-  ingredients:
-${recipe}`);
+console.log(recipe.title);
+console.log(`Serves: ${recipe.serves}`);
+console.log("Ingredients:");
+for (const ingredient of recipe.ingredients) {
+  console.log(ingredient);
+}
```

Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,16 @@
-function contains() {}
+function contains(obj, prop) {
+  if (typeof obj !== "object" || obj === null) {
+    return false;
+  }
+  if (typeof prop !== "string") {
+    return false;
+  }
```
cjyuan: Other types of value can also be used as key (property name) -- They are just converted to equivalent string implicitly.

So `obj[1]` or `obj[1.0]` or `obj[3-2]` are treated the same as `obj["1"]`.

Sprint-2/implement/contains.test.js
```diff
@@ -20,16 +20,27 @@ as the object doesn't contains a key of 'c'
...
+test("contains returns false when property does not exist", function () {
+  expect(contains({ a: 1, b: 2 }, "c")).toBe(false);
+});
 
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+test("contains returns false when passed an array", function () {
+  expect(contains([1, 2, 3], "a")).toBe(false);
+});
```

Sprint-2/implement/querystring.js
```diff
@@ -6,7 +6,7 @@ function parseQueryString(queryString) {
   const keyValuePairs = queryString.split("&");
 
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
+    const [key, value] = pair.split(/=(.*)/s);
     queryParams[key] = value;
```
cjyuan, line 17: For the following function call, does your function return the value you expect?

```
```

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,13 @@
-function tally() {}
+function tally(items) {
+    if (!Array.isArray(items)) {
+        throw new Error('Items must be an array');
+    }
+    const counts = {};
+    for (const item of items) {
+        counts[item] = (counts[item] || 0) + 1;
+    }
+    return counts;
```
cjyuan: Does the following function call return the value you expect?
```
```

KKtech06: Hi CJ I saw the issue with "toString". I fixed it by using Object.create(null).

Thanks you

Sprint-2/implement/querystring.js
KKtech06: Hello CJ I fixed it by skipping empty pairs and splitting at the first "=".

Thank you

Sprint-2/implement/contains.test.js
KKtech06: Hello CJ I have fixed the issue by "a" to 0 so it actually checks the array .

thank you

Sprint-2/implement/contains.js
KKtech06: Hello CJ I didn’t realise keys get converted to strings automatically. I removed that check now and hope everything is fine now.

Thank you

Sprint-2/debug/recipe.js
KKtech06: Hello CJ  I have  updated my code to use join() hopefully my code is improved.

Thank you

cjyuan:
Changes look good. Well done.

illicitonion:

PR #1062, opened 2026-03-21
London | 26-ITP-Jan | Oussama Mouggal | Sprint 3 | Alarm clock

Oussama-Mouggal:
Self checklist

## Changelist

I changed the title and added the alarm clock logic to the file as requested.

### DIFF
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -1,4 +1,44 @@
-function setAlarm() {}
+let countdownTimerId = null;
+
+function setAlarm() {
+  const heading = document.getElementById("timeRemaining");
+  const input = document.getElementById("alarmSet");
+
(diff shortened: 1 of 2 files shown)


cjyuan (commented):
(1) Currently when starting a new countdown, the application does not always return to a clean initial state, which can lead to inconsistent behaviour between runs.

---
(2) When input is 1, the alarm is played one second later -- an expected behavior.
However, when input is 0, the alarm is also played one second later (instead of being played immediately).
Can you make the app's behaviour more consistent?

Oussama-Mouggal:
Thank you @cjyuan for your valuable feedback. I added a dedicated reset path for clean restarts and made 0-second input trigger the alarm immediately for consistent behavior.

cjyuan:
Changes look good.

illicitonion:

PR #1064, opened 2026-03-21
Sheffield | 26-ITP-jan | Richard Frimpong | Sprint 3 | Alarm Clock

Richiealx:
## Learners, PR Template

Self checklist

## Changelist

This PR implements the Sprint 3 Alarm Clock app.

The application starts with the heading showing `Time Remaining: 00:00` and no alarm sound playing.

When the user enters a number of seconds and clicks the "Set Alarm" button, the heading updates immediately in `mm:ss` format. The countdown then decreases by one second every 1000 ms.

When the timer reaches `00:00`, the background colour changes and the alarm sound starts playing continuously. Clicking the "Stop Alarm" button stops the alarm sound.

I also added logic to clear any existing timer before starting a new one, reset the background state when a new alarm is set, and format the display consistently.

The app was tested in the browser to confirm:
- the initial heading shows `00:00`
- the timer formats values correctly
- the timer counts down every second
- the background colour changes when the timer reaches zero
- the alarm sound plays at zero
- the stop button stops the alarm sound

I also ran the provided tests for the alarm clock task, and all tests passed successfully.

## Questions

No questions at this time.


Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,10 +1,114 @@
...
+function resetBackground() {
+  document.body.classList.remove("alarm-finished");
+}
+
+/**
+ * Changes the background colour when the timer reaches zero.
+ */
+function showAlarmFinishedState() {
+  document.body.classList.add("alarm-finished");
+}
```
cjyuan, line 38: How about combining them into one function and use a parameter to control whether to add/remove the CSS class?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,10 +1,114 @@
...
+  }
+
+  // Stop any previous countdown before starting a new one
+  stopCountdown();
+
+  // Stop any alarm sound that may already be playing
+  pauseAlarm();
+
+  // Reset the page state for a new alarm
+  resetBackground();
```
cjyuan: Could consider 
1. Including these three operations into a `resetAll()` function, and
2. Call the function to reset the application states as soon as the "Start" button is clicked (even when the input is incorrect).

cjyuan (commented):
When input is 1, the alarm is played one second later -- an expected behavior.
However, when input is 0, the alarm is also played one second later (instead of being played immediately). Can you make the app's behaviour more consistent?

Sprint-3/alarmclock/alarmclock.js
Richiealx: Thank you for the feedback.

I’ve now updated the alarm clock based on your comments:

1. I changed the behaviour so that when the input is `0`, the alarm is triggered immediately instead of one second later.
2. I combined the background update logic into a single function that adds or removes the CSS class based on a parameter.
3. I created a `resetAll()` function to group the reset operations together, and I now call it as soon as the "Set Alarm" button is clicked.

I also cleaned up the alarm clock code so the duplicated starter functions were removed and the final file structure is now clear and consistent.

I tested the app again and confirmed the updated behaviour works as expected.

cjyuan:
Please revert the changes made in the Sprint-2 folder.

Richiealx:
Thank you for your feedback.

I identified that I had accidentally committed changes to the wrong branch, which caused unrelated files to appear in this PR.

I have now cleaned the branch so that this PR only includes changes within `Sprint-3/alarmclock/`.

I also implemented the suggested improvements:

* The alarm now triggers immediately when the input is 0
* The background update logic has been combined into a single function with a parameter
* A `resetAll()` function now handles resetting the application state when the Set Alarm button is clicked

cjyuan:
Changes look good. Well done.

illicitonion:

PR #1065, opened 2026-03-21
Birmingham | 26-ITP-Jan | Tasleem Adedokun | Sprint 2 | Data group-coursework

tasleemadedokun:
## Learners, PR Template

Self checklist

## Changelist

Brief;
I have done all the coursework to the best of my abilities.

### DIFF
--- a/Sprint-2/debug/recipe.js
+++ b/Sprint-2/debug/recipe.js
@@ -13,8 +13,5 @@ const recipe = {
 };
 
 console.log(`${recipe.title} serves ${recipe.serves}
-ingredients:`);
-
-for (const ingredient of recipe.ingredients) {
-  console.log(ingredient);
-}
+ingredients:
+${recipe.ingredients.join("\n")}`);
--- a/Sprint-2/debug/recipe.js
+++ b/Sprint-2/debug/recipe.js
@@ -1,5 +1,7 @@
 // Predict and explain first...
 
+// This is trying to print the whole object (recipe) instead of the ingredients array.
+
 // This program should log out the title, how many it serves and the ingredients.
 // Each ingredient should be logged on a new line
 // How can you fix it?
@@ -11,5 +13,8 @@ const recipe = {
 };
 
 console.log(`${recipe.title} serves ${recipe.serves}
-  ingredients:
-${recipe}`);
+ingredients:`);
+
+for (const ingredient of recipe.ingredients) {
+  console.log(ingredient);
+}
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -4,7 +4,8 @@ function contains(obj, prop) {
     return false;
   }
   // Check if the property exists directly on the object
-  return Object.prototype.hasOwnProperty.call(obj, prop);
+ return Object.hasOwn(obj, prop);
 }
+
 
 module.exports = contains;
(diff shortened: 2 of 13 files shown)


Sprint-2/debug/recipe.js
```diff
@@ -11,5 +13,8 @@ const recipe = {
 };
 
 console.log(`${recipe.title} serves ${recipe.serves}
-  ingredients:
-${recipe}`);
+ingredients:`);
+
+for (const ingredient of recipe.ingredients) {
+  console.log(ingredient);
+}
```

Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,10 @@
-function contains() {}
+function contains(obj, prop) {
+  // Check that obj is an object (not null) and not an array
+  if (typeof obj !== "object" || obj === null || Array.isArray(obj)) {
+    return false;
+  }
+  // Check if the property exists directly on the object
+  return Object.prototype.hasOwnProperty.call(obj, prop);
```
cjyuan: Suggestion: Look up
1. `Object.hasOwn()` 
2. `Object.prototype.hasOwnProperty.call(obj, prop)` vs. `obj.hasOwnProperty.call(prop)`

Sprint-2/implement/querystring.js
```diff
@@ -1,12 +1,13 @@
...
     return queryParams;
   }
   const keyValuePairs = queryString.split("&");
 
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
+    const [key, ...rest] = pair.split("=");
+    const value = rest.join("="); // this join back the remaining parts
     queryParams[key] = value;
   }
```
cjyuan, line 12: Note: (No change required)

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,15 @@
...
+function tally(items) {
+  if (!Array.isArray(items)) {
+    throw new Error("Input must be an array");
+  }
+
+  const result = {};
+
+  for (const item of items) {
+    result[item] = (result[item] || 0) + 1;
+  }
```
```
```

Sprint-2/debug/recipe.js
tasleemadedokun: > Your code works.
> 
> Here is an alternative worth exploring: Since ingredient values are separated by '\n' in the output, we could also use `Array.prototype.join()` to construct the equivalent string and then output the resulting string.

Thank you very I have implemented changes.

Sprint-2/implement/contains.js
tasleemadedokun: > Suggestion: Look up
> 
> 1. `Object.hasOwn()`
> 2. `Object.prototype.hasOwnProperty.call(obj, prop)` vs. `obj.hasOwnProperty.call(prop)`

I have worked on it and understand there is a modern way.

Sprint-2/implement/querystring.js
tasleemadedokun: noted with thanks

Sprint-2/implement/tally.js
tasleemadedokun: > Does the following function call returns the value you expect?
> 
> ```
> tally(["toString", "toString"]);
> ```
> 
> Suggestion: Look up an approach to create an empty object with no inherited properties.

I have effect changes.

tasleemadedokun: Thank you very much.

cjyuan:
Changes look good. Well done.

tasleemadedokun:
> Changes look good. Well done.

Thank you very much always.

illicitonion:

PR #1072, opened 2026-03-22
Birmingham | ITP-Jan-26 | Ayodeji Ayorinde | Sprint 2| Coursework/Sprint2

Ayogit1:
## Learners, PR Template

Self checklist

## Changelist

Sprint 2 completed

### DIFF
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,11 +1,11 @@
-
-function contains(obj, propName) {
-
-    if (typeof obj !== "object" || obj === null || Array.isArray(obj)) {
+function contains(item, propertyName) {
+    // Guard clause for non-objects, null, and arrays
+    if (item === null || typeof item !== "object" || Array.isArray(item)) {
         return false;
     }
-    return propName in obj;
+
+    // Object.hasOwn is safer because it ignores inherited prototype properties
+    return Object.hasOwn(item, propertyName);
 }
 
 module.exports = contains;
-
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,3 +1,19 @@
+/*
 function contains() {}
 
 module.exports = contains;
+*/
+
+function contains(obj, propName) {
+    if (
+        typeof obj !== 'object' ||
+        obj === null ||
+        (typeof propName !== 'string' && typeof propName !== 'number' && typeof propName !== 'symbol')
+    ) {
+        return false;
+    }
+    return propName in obj;
+}
+
+module.exports = contains;
+
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,15 +1,7 @@
-/*
-function contains() {}
-
-module.exports = contains;
-*/
 
 function contains(obj, propName) {
-    if (
-        typeof obj !== 'object' ||
-        obj === null ||
-        (typeof propName !== 'string' && typeof propName !== 'number' && typeof propName !== 'symbol')
-    ) {
+
+    if (typeof obj !== "object" || obj === null || Array.isArray(obj)) {
         return false;
     }
     return propName in obj;
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -1,5 +1,3 @@
-const contains = require("./contains.js");
-
 /*
 Implement a function called contains that checks an object contains a
 particular property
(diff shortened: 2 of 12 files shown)


Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,19 @@
...
 function contains() {}
 
 module.exports = contains;
+*/
+
+function contains(obj, propName) {
+    if (
+        typeof obj !== 'object' ||
+        obj === null ||
+        (typeof propName !== 'string' && typeof propName !== 'number' && typeof propName !== 'symbol')
```
cjyuan: - Why reject these types of value as property name? All types of value can be used as property name. Besides symbols, all values are simply converted to a string implicitly before they are used a key.

- The descriptions in `contains.test.js` also mention that the function should return `false` when the first parameter is an array.

Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,19 @@
...
+    ) {
+        return false;
+    }
+    return propName in obj;
```
```
```

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,18 @@
...
+
+    if (!Array.isArray(arr)) {
+        throw new Error("Input must be an array");
+    }
+
+    // The reduction method effectively builds the frequency map
+    return arr.reduce((acc, item) => {
+        acc[item] = (acc[item] || 0) + 1;
+        return acc;
+    }, {});
```
```
```

cjyuan (commented):
- This exercise also expects the `*.test.js` files in the "implement" folder to be updated. There are tests to be implemented.

Sprint-2/implement/contains.js
Ayogit1: This has now been modified. Thanks.

Sprint-2/implement/contains.js
Ayogit1: Thanks, this has now been fixed. All tests now created.

Sprint-2/implement/tally.js
Ayogit1: Thanks for the observation. This has now been addressed.

Sprint-2/implement/contains.js
cjyuan: Can you elaborate what you meant by "fixed"? I don't see any change.

Sprint-2/implement/tally.js
cjyuan: Can you elaborate how you addressed this issue?

Sprint-2/implement/contains.test.js
```diff
@@ -33,3 +31,27 @@ test.todo("contains on empty object returns false");
...
+    const obj = { a: 1, b: 2 };
+    expect(contains(obj, "c")).toBe(false);
+});
+
+test("returns false when passed invalid parameters like an array", () => {
+    expect(contains([], "a")).toBe(false);
+    expect(contains(null, "a")).toBe(false);
+    expect(contains(123, "a")).toBe(false);
+    expect(contains("string", "a")).toBe(false);
+});
```
cjyuan: This tests `expect(contains([], "a")).toBe(false)` cannot not yet confirm that the function correctly returns false when the first argument is an array. This is because both function calls could also return `false` simply because "a" is not a key of the array.

**Arrays are objects, with their indices acting as keys**. A proper test should use a **non-empty array along with a valid key** to ensure the function returns `false` specifically because the input is an array, not because the key is missing.

Note: These output `true`
```
  console.log( "0" in ["A"] );
  console.log( Object.hasOwn(["A"], "0") );
```
and these output `false`
```
  console.log( "a" in ["A"] );
  console.log( Object.hasOwn(["A"], "a") );
```

Sprint-2/implement/contains.js
Ayogit1: Object.hasOwn() is used because it is the superior choice.
Object.hasOwn(): Returns true only if the property is a "direct" property of that specific object.

in operator: Returns true if the property exists on the object or anywhere in its "Prototype Chain" (inherited properties like toString, valueOf, etc.).

Sprint-2/implement/contains.test.js
Ayogit1: Contains.js now modified to
-Checks if a value is a valid object and contains a specific property.
-Arrays, null, and non-object types return false.


Contains.test.js modified to the following:-
Safety: The function won't crash if you pass it null or undefined.
Accuracy: It distinguishes between a "plain object" and an "array object."
The "Array Trap": By testing contains(["A"], "0"), it verifies that the function specifically blocks arrays as requested by the Acceptance Criteria, rather than returning false just because a key was missing.

Sprint-2/implement/tally.js
Ayogit1: When you use Object.create(null), the resulting object does not have a toString property.
The code checks acc["toString"].
It finds undefined (because there is no inherited method).
It falls back to 0.
It calculates 0 + 1 = 1.
Result: { toString: 1 }.

cjyuan:
Changes look good.

illicitonion:

PR #1074, opened 2026-03-22
Sheffield | 26-ITP-Jan | Mahammad Osman | Sprint 1 | Sprint-1

Darkidd77:
## Learners, PR Template

Self checklist

## Changelist

`dedupe.js`: Implemented array deduplication logic.

`max.js`: Refactored to use a functional pipeline (`.filter()` and `Math.max` with the spread operator `...`) to safely ignore NaN and non-numeric strings.

`sum.js`: Implemented sum calculation using `.reduce()`.

`includes.js`: Fixed a premature return bug inside the loop so the function correctly evaluates the entire array before returning false.

`solution.js` (Advent of Code): Built a data parsing pipeline using `fs.readFileSync`, `.split('\n')`, and `.map(Number)` to process a text file and calculate a final total using `.reduce()`.

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -6,9 +6,28 @@
 // or 'list' has mixed values (the function is expected to sort only numbers).
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  if (!Array.isArray(list) || list.length === 0) {
+    return null;
+  }
+
+  const numbers = list.filter((item) => typeof item === "number");
+
+  if (numbers.length === 0) {
+    return null;
+  }
+
+  numbers.sort((a, b) => a - b);
+
+  const middleIndex = Math.floor(numbers.length / 2);
+
+  if (numbers.length % 2 !== 0) {
+    const median = numbers.splice(middleIndex, 1)[0];
+    return median;
+  } else {
+    const middleTwo = numbers.splice(middleIndex - 1, 2);
+    const median = (middleTwo[0] + middleTwo[1]) / 2;
+    return median;
+  }
 }
 
 module.exports = calculateMedian;
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -10,7 +10,7 @@ function calculateMedian(list) {
     return null;
   }
 
-  const numbers = list.filter((item) => typeof item === "number");
+  const numbers = list.filter((item) => Number.isFinite(item));
 
   if (numbers.length === 0) {
     return null;
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -21,12 +21,12 @@ function calculateMedian(list) {
   const middleIndex = Math.floor(numbers.length / 2);
 
   if (numbers.length % 2 !== 0) {
-    const median = numbers.splice(middleIndex, 1)[0];
-    return median;
+    return numbers[middleIndex];
   } else {
-    const middleTwo = numbers.splice(middleIndex - 1, 2);
-    const median = (middleTwo[0] + middleTwo[1]) / 2;
-    return median;
+    const leftMiddle = numbers[middleIndex - 1];
+    const rightMiddle = numbers[middleIndex];
+
+    return (leftMiddle + rightMiddle) / 2;
   }
 }
 
--- a/Sprint-1/implement/dedupe.test.js
+++ b/Sprint-1/implement/dedupe.test.js
@@ -27,6 +27,8 @@ describe("dedupe", () => {
   test("should return a copy of the original array when there are no duplicates", () => {
     const input = [1, 2, 3];
     expect(dedupe(input)).toEqual(input);
+
+    expect(dedupe(input)).not.toBe(input);
   });
 
   // Given an array of strings or numbers
--- a/Sprint-1/implement/dedupe.test.js
+++ b/Sprint-1/implement/dedupe.test.js
@@ -11,18 +11,31 @@ E.g. dedupe([5, 1, 1, 2, 3, 2, 5, 8]) returns [5, 1, 2, 3, 8]
 E.g. dedupe([1, 2, 1]) returns [1, 2]
 */
 
-// Acceptance Criteria:
+describe("dedupe", () => {
+  // Acceptance Criteria:
 
-// Given an empty array
-// When passed to the dedupe function
-// Then it should return an empty array
-test.todo("given an empty array, it returns an empty array");
+  // Given an empty array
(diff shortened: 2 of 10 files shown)


Sprint-1/fix/median.js
```diff
@@ -6,9 +6,28 @@
...
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  if (!Array.isArray(list) || list.length === 0) {
+    return null;
+  }
+
+  const numbers = list.filter((item) => typeof item === "number");
```
cjyuan: Do you plan to consider also `-Infinity`, `Infinity`, and `NaN` in the median calculation?

Sprint-1/fix/median.js
```diff
@@ -6,9 +6,28 @@
...
+  const middleIndex = Math.floor(numbers.length / 2);
+
+  if (numbers.length % 2 !== 0) {
+    const median = numbers.splice(middleIndex, 1)[0];
+    return median;
+  } else {
+    const middleTwo = numbers.splice(middleIndex - 1, 2);
+    const median = (middleTwo[0] + middleTwo[1]) / 2;
+    return median;
+  }
```
cjyuan, line 30: There is a simpler syntax to access the element positioned at `middleIndex` in an array.

Sprint-1/implement/dedupe.test.js
```diff
@@ -11,18 +11,31 @@ E.g. dedupe([5, 1, 1, 2, 3, 2, 5, 8]) returns [5, 1, 2, 3, 8]
...
-// Given an array with no duplicates
-// When passed to the dedupe function
-// Then it should return a copy of the original array
+  // Given an array with no duplicates
+  // When passed to the dedupe function
+  // Then it should return a copy of the original array
+  test("should return a copy of the original array when there are no duplicates", () => {
+    const input = [1, 2, 3];
+    expect(dedupe(input)).toEqual(input);
+  });
```
cjyuan, line 34: This test should fail if the function returns the original array (instead of a **copy** of the original array).

Sprint-1/implement/max.test.js
```diff
@@ -12,32 +12,56 @@ We have set things up already so that this file can see your function from the o
...
+  test("should return the max and ignore non-numeric values when passed an array with non-number values", () => {
+    expect(findMax(["hey", 10, "hi", 60, 10])).toEqual(60);
+  });
+
+  // Given an array with only non-number values
+  // When passed to the max function
+  // Then it should return the least surprising value given how it behaves for all other inputs
+  test("should return -Infinity when passed an array with ONLY non-number values", () => {
+    expect(findMax(["hey", "hi", "hello"])).toEqual(-Infinity);
+  });
```

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,12 @@
 function sum(elements) {
+  const numbersOnly = elements.filter(
+    (element) => typeof element === "number" && !isNaN(element)
+  );
+
+  if (numbersOnly.length === 0) {
+    return 0;
+  }
+  return Number(numbersOnly.reduce((acc, curr) => acc + curr, 0).toFixed(1));
```
cjyuan: Why round the sum to one decimal place?

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,41 @@ const sum = require("./sum.js");
...
+test("should return the correct total sum when passed an array containing negative numbers", () => {
+  expect(sum([-10, 20, -5])).toEqual(5);
+});
 
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
+test("should return the correct total sum when passed an array with decimal/float numbers", () => {
+  expect(sum([1.5, 2.3, 0.8, 3.1])).toEqual(7.7);
+});
```

So the following could happen
```javascript

```

Suggestion: Look up 

Sprint-1/fix/median.js
Darkidd77: No, definitely I don't want to consider those values. I swapped `typeof` with `Number.isFinite()`
Thank you.

Sprint-1/fix/median.js
Darkidd77: Done, indeed `splice()` method was just complicating the syntax, I swapped it with bracket notation `numbers[middleIndex]`

Sprint-1/implement/dedupe.test.js
Darkidd77: I added .not.toBe(). I learned that .toEqual() only checks the values, but .toBe() checks the memory reference to ensure we return a brand-new array

Sprint-1/implement/max.test.js
Darkidd77: Yes I added the string `"300"` tested it, it does successfully ignore it.

Sprint-1/implement/sum.test.js
Darkidd77: I swapped `toEqual()` for `.toBeCloseTo()` in the decimal test suite. This ensures the test evaluates the mathematical intent correctly without failing due to JS

Sprint-1/implement/sum.js
Darkidd77: I used `.toFixed(1)` to avoid JavaScript floating-point rounding errors . I removed `.toFixed(1)` to prevent losing data precision and preserve the exact sum.

Sprint-1/implement/dedupe.test.js
```diff
@@ -11,18 +11,33 @@ E.g. dedupe([5, 1, 1, 2, 3, 2, 5, 8]) returns [5, 1, 2, 3, 8]
...
+  test("should return a copy of the original array when there are no duplicates", () => {
+    const input = [1, 2, 3];
+    expect(dedupe(input)).toEqual(input);
 
-// Given an array of strings or numbers
-// When passed to the dedupe function
-// Then it should return a new array with duplicates removed while preserving the 
-// first occurrence of each element from the original array.
+    expect(dedupe(input)).not.toBe(input);
+  });
```
cjyuan, line 34: There is a chance that, even though the returned value has incorrect elements (for example, `[1, 1, 1]`), 

cjyuan (commented):
Other changes look good.

Sprint-1/implement/dedupe.test.js
Darkidd77: Done! Yes, it's because this test could still pass even if the original array were destroyed or mutated in the function. It's a very interesting trap. 
To fix it, I saved the output to a result variable and properly compared it against a hardcoded array. Thinking about TDD this deeply has been a great learning experience. Thank you

cjyuan:
All good! Well done.

illicitonion:

PR #1076, opened 2026-03-23
Sheffield | 26-ITP-Jan | Mona -Eltantawy | Sprint 1 | Data- Groups/sprint 1

Mona-Eltantawy:
Self checklist

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -6,36 +6,29 @@
 // or 'list' has mixed values (the function is expected to sort only numbers).
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
-}
-
-//function calculateMedian(list) {
   // Check if input is an array
   if (!Array.isArray(list)) return null;
 
   // Filter only valid numbers
   const numbers = list.filter(
-    (item) => typeof item === "number" && !isNaN(item)
+    (item) => typeof item === "number" && Number.isFinite(item)
   );
 
   // If no valid numbers, return null
   if (numbers.length === 0) return null;
 
-  // Sort numbers (without mutating original array)
+  // Sort without mutating original array
   const sorted = [...numbers].sort((a, b) => a - b);
 
   const middle = Math.floor(sorted.length / 2);
 
-  // Even length → average of two middle values
+  // Even length
   if (sorted.length % 2 === 0) {
     return (sorted[middle - 1] + sorted[middle]) / 2;
   }
 
-  // Odd length → middle value
+  // Odd length
   return sorted[middle];
 }
 
-module.exports = calculateMedian;
 module.exports = calculateMedian;
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -1,33 +1,26 @@
-// Fix this implementation
-// Start by running the tests for this function
-// If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
-
-// Hint: Please consider scenarios when 'list' doesn't have numbers (the function is expected to return null)
-// or 'list' has mixed values (the function is expected to sort only numbers).
-
 function calculateMedian(list) {
   // Check if input is an array
   if (!Array.isArray(list)) return null;
 
   // Filter only valid numbers
   const numbers = list.filter(
-    (item) => typeof item === "number" && Number.isFinite(item)
+    (item) => typeof item === "number" && !isNaN(item)
   );
 
-  // If no valid numbers, return null
+  // Return null if no valid numbers
   if (numbers.length === 0) return null;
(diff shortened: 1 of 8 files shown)


Sprint-1/fix/median.js
```diff
@@ -6,9 +6,29 @@
...
+  // Filter only valid numbers
+  const numbers = list.filter(
+    (item) => typeof item === "number" && Number.isFinite(item)
+  );
+
+  // If no valid numbers, return null
+  if (numbers.length === 0) return null;
+
+  // Sort without mutating original array
+  const sorted = [...numbers].sort((a, b) => a - b);
```
cjyuan: Why make a copy of `numbers` before sorting?

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,7 @@
-function findMax(elements) {
+function findMax(arr) {
+  return arr
+    .filter((item) => typeof item === "number")
+    .reduce((acc, num) => (num > acc ? num : acc), -Infinity);
```
cjyuan: Does your function return the value you expected from the following function calls?

```
findMax([NaN])
findMax([0, NaN, 1])
```

Sprint-1/implement/max.test.js
```diff
@@ -16,28 +16,47 @@ const findMax = require("./max.js");
...
-
+test("array with non-number values, ignores non-numeric and returns max number", () => {
+  expect(findMax(["hey", 10, "hi", 60, 10])).toBe(60);
+});
 // Given an array with only non-number values
 // When passed to the max function
 // Then it should return the least surprising value given how it behaves for all other inputs
+test("array with only non-number values, returns -Infinity", () => {
+  expect(findMax(["a", "b", null, {}])).toBe(-Infinity);
+});
```
cjyuan, line 62: When a string representing a valid numeric literal (for example, `"300"`) is compared to a number, 

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,40 @@ const sum = require("./sum.js");
...
+  expect(sum([1, 2, 3, -2])).toEqual(4);
+});
 
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
-
+test("given an array with decimal/float number, returns the correct total sum", () => {
+  expect(sum([1, 2, 3, 1.5, 2.1])).toEqual(9.6);
+});
```

So the following could happen
```javascript

```

Suggestion: Look up 

cjyuan (commented):
There is another file in `refactor` that needs to be updated.

Sprint-1/fix/median.js
Mona-Eltantawy: So that .sort( ) modifies the copy not the original input.

Sprint-1/implement/max.js
Mona-Eltantawy: I fixed it. Thanks CJ

Sprint-1/implement/max.test.js
Mona-Eltantawy: I added the string "300" to the non number test case values.

Sprint-1/implement/sum.test.js
Mona-Eltantawy: I changed the expacted output in the test case to : toBeCloseTo( ) instead of .toEqual to handle the issue.

Sprint-1/fix/median.js
cjyuan: `numbers` is not the original array though.

Mona-Eltantawy: I tried to re-fix it. can you have another look please.

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,8 @@
-function findMax(elements) {
+function findMax(arr) {
+  return arr
+    .filter((item) => typeof item === "number" && Number.isFinite(item));
+    if(numbers.length=== 0) return null; 
+   return numbers.reduce((acc, num) => (num > acc ? num : acc), -Infinity);
 }
```
cjyuan, line 5: This function is not working as expected. Can you fix it?

Mona-Eltantawy: I fixed it . Thanks CJ .

Sprint-1/implement/max.js
```diff
@@ -1,8 +1,7 @@
 function findMax(arr) {
   return arr
-    .filter((item) => typeof item === "number" && Number.isFinite(item));
-    if(numbers.length=== 0) return null; 
-   return numbers.reduce((acc, num) => (num > acc ? num : acc), -Infinity);
+    .filter((item) => typeof item === "number" && Number.isFinite(item))
```
cjyuan, line 3: Note: `Number.isFinite(item)` also checks `typeof item === "number"`

cjyuan (commented):
Changes look good.

illicitonion:

PR #1077, opened 2026-03-23
London | 26-ITP-January | Mouawia Elkhalifa | Sprint 2 | Data Groups

MouawiaElkhalifa:
## Learners, PR Template

Self checklist

## Changelist

Implemented all Sprint 2 exercises including:
- Debug tasks (address, author, recipe)
- Implement tasks (contains, lookup, querystring, tally)
- Interpret task (invert)

All functions were tested and verified to meet the expected behavior. Edge cases such as invalid input and duplicate values were handled where required.


Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,9 @@
-function contains() {}
+function contains(obj, prop) {
+  if (typeof obj !== "object" || obj === null || Array.isArray(obj)) {
+    return false;
+  }
+
+  return prop in obj;
```
```
```

Sprint-2/implement/contains.test.js
```diff
@@ -20,16 +20,27 @@ as the object doesn't contains a key of 'c'
...
-
+test("return false when the property does not exist", () =>{
+    expect(contains({a:1, b:3}, "c")).toEqual(false)
+});
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+test("return false when paramters invaliud ", () =>{
+    expect(contains([], "a")).toEqual(false)
+});
```

Sprint-2/implement/querystring.js
```diff
@@ -1,13 +1,18 @@
...
     return queryParams;
   }
+
   const keyValuePairs = queryString.split("&");
 
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
-    queryParams[key] = value;
+    const [key, ...valueParts] = pair.split("=");
+    queryParams[key] = valueParts.join("=");
```
cjyuan: Note: (No change required)

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,26 @@
...
+  const result = {};
+
+  //  count items
+  for (const item of arr) {
+    if (result[item] === undefined) {
+      result[item] = 1;
+    } else {
+      result[item] = result[item] + 1;
+    }
+  }
```
```
```

Sprint-2/interpret/invert.js
```diff
@@ -5,25 +5,63 @@
...
 
   for (const [key, value] of Object.entries(obj)) {
-    invertedObj.key = value;
+    if (invertedObj[value] === undefined) {
+      invertedObj[value] = key;
+    } else if (Array.isArray(invertedObj[value])) {
+      invertedObj[value].push(key);
+    } else {
+      invertedObj[value] = [invertedObj[value], key];
+    }
```
cjyuan: Good job in recognising the possibility of collision and taking an extra step to address it.

Sprint-2/implement/contains.js
MouawiaElkhalifa: "Updated to use Object.hasOwn() and fixed the logic to correctly identify values inside arrays."

Sprint-2/implement/querystring.js
MouawiaElkhalifa: Added the rest operator and .join("=") to prevent values with extra = signs from being cut off. Also used decodeURIComponent() to handle special characters like spaces."

Sprint-2/implement/tally.js
MouawiaElkhalifa: Switched to Object.create(null) to prevent prototype pollution. Added a test for reserved words like toString."

Sprint-2/interpret/invert.js
MouawiaElkhalifa: Added collision handling so shared values are grouped into arrays. I also included Jest tests to verify this works for multiple duplicate values."

Sprint-2/interpret/invert.js
```diff
@@ -1,29 +1,27 @@
...
-
-// b) What is the current return value when invert is called with { a: 1, b: 2 }
-
-// c) What is the target return value when invert is called with {a : 1, b: 2}
-
-// c) What does Object.entries return? Why is it needed in this program?
-
-// d) Explain why the current return value is different from the target output
-
-// e) Fix the implementation of invert (and write tests to prove it's fixed!)
```
cjyuan, line 29: You deleted your answers to these questions.

Anyway, do you recognise what this statement (in the original code) does?
>  invertedObj.key = value;

Sprint-2/implement/contains.test.js
MouawiaElkhalifa: Strengthened the array test to return false even if an index exists (e.g., contains(["a"], 0)).

Sprint-2/interpret/invert.js
MouawiaElkhalifa: Restored missing answers (a-e) and corrected the logic. Replaced literal .key with dynamic [value] notation to prevent overwriting and ensure keys are swapped correctly.

Sprint-2/interpret/invert.js
```diff
@@ -1,27 +1,42 @@
-/**
- * Invert Function
- * Swaps keys and values of an object. If multiple keys share the same value,
- * it groups those keys into an array to prevent overwriting data.
- */
+// a) What is the current return value when invert is called with { a : 1 }
+// {"1": "a"}
+
+// b) What is the current return value when invert is called with { a: 1, b: 2 }
+// {"1": "a", "2": "b"}
```
cjyuan: I think the "current return value" refers to the return value of the original code. 
The value you showed is the "target return value". 
I am sure you know the answers.

cjyuan (commented):
Changes look good. Well done.

Sprint-2/interpret/invert.js
MouawiaElkhalifa: thanks,Fixed (a) and (b) — they now show the correct current output.

illicitonion:

PR #1079, opened 2026-03-23
London | 26-ITP-January | Karla Grajales | Sprint 2 | Data Groups

Grajales-K:
## Learners, PR Template

Self checklist

## Changelist

This project focuses on manipulating arrays of objects, accessing their properties, and returning specific values. I practised using object.entries() and object.values() to transform and retrieve data from particular structures.

Furthermore, when applying TDD (Test-Driven Development) to return object operations, I learned the importance of data cleansing. This involves removing unnecessary data types and validating the output by checking for empty objects arrays and null values to ensure the integrity  of the final result.

### DIFF
--- a/Sprint-2/debug/recipe.js
+++ b/Sprint-2/debug/recipe.js
@@ -15,9 +15,6 @@ const recipe = {
   ingredients: ["olive oil", "tomatoes", "salt", "pepper"],
 };
 
+
 console.log(`Recipe: ${recipe.title} \nserves: ${recipe.serves}`);
-for(const ingredient of recipe.ingredients) {
-  console.log(`${ingredient}`);
-}
-
-
+console.log(recipe.ingredients.join("\n"));
--- a/Sprint-2/debug/recipe.js
+++ b/Sprint-2/debug/recipe.js
@@ -4,12 +4,20 @@
 // Each ingredient should be logged on a new line
 // How can you fix it?
 
+// it won't work because the last line is calling the object property recipe and we
+// need the ingredientes, the console.log is trying to log the whole array of ingredients 
+// as a string, but we want to log each ingredient on a new line. We can fix this by 
+// using a for loop to iterate over the ingredients array and log each ingredient separately.
+
 const recipe = {
-  title: "bruschetta",
+  title: "Bruschetta",
   serves: 2,
   ingredients: ["olive oil", "tomatoes", "salt", "pepper"],
 };
 
-console.log(`${recipe.title} serves ${recipe.serves}
-  ingredients:
-${recipe}`);
+console.log(`Recipe: ${recipe.title} \nserves: ${recipe.serves}`);
+for(const ingredient of recipe.ingredients) {
+  console.log(`${ingredient}`);
+}
+
+
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -3,10 +3,8 @@ function contains(obj, property) {
         return false;
     }
 
-    return property in obj;
+    return Object.hasOwn(obj, property);
 }
 
 
 module.exports = contains;
-
-
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,3 +1,12 @@
-function contains() {}
+function contains(obj, property) {
+    if (typeof obj !== "object" || obj === null || Array.isArray(obj)){
+        return false;
+    }
+
+    return property in obj;
+}
+
 
 module.exports = contains;
+
+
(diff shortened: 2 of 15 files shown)


Sprint-2/debug/recipe.js
```diff
@@ -4,12 +4,20 @@
...
   ingredients: ["olive oil", "tomatoes", "salt", "pepper"],
 };
 
-console.log(`${recipe.title} serves ${recipe.serves}
-  ingredients:
-${recipe}`);
+console.log(`Recipe: ${recipe.title} \nserves: ${recipe.serves}`);
+for(const ingredient of recipe.ingredients) {
+  console.log(`${ingredient}`);
+}
```

Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,12 @@
-function contains() {}
+function contains(obj, property) {
+    if (typeof obj !== "object" || obj === null || Array.isArray(obj)){
+        return false;
+    }
+
+    return property in obj;
```
```
```

Sprint-2/implement/contains.test.js
```diff
@@ -16,20 +16,35 @@ as the object doesn't contains a key of 'c'
...
 // Then it should return false
+test("return false when pass an object that does not contain the property name", () => {
+    expect(contains({id: 12, status: "active"}, "age")).toBe(false);
+})
 
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+test("return false when pass an array instead of an object", () => {
+    expect(contains([1, 2, 3], "name")).toBe(false);
```

cjyuan (commented):
These four files should also be updated:
- Sprint-2/implement/querystring.js
- Sprint-2/implement/querystring.test.js
- Sprint-2/implement/tally.js
- Sprint-2/implement/tally.test.js

Sprint-2/implement/contains.js
Grajales-K: I see it @cjyuan, it's more robust `Object.hasOwn() `because `in` can give falsy test. thank you.

Sprint-2/implement/contains.test.js
Grajales-K: understand, So even when the function detects it was an array (stopping the program), in case the function changes later, the test still proves if it's doing what we definitely expect!

Sprint-2/implement/querystring.js
```diff
@@ -1,15 +1,25 @@
...
+    if(!pair) continue; 
+  
+    const [key, ...rest] = pair.split("=");
+
+    const value = rest.join("=");
+
+    if (key) {
 
+      const decodedkey = decodeURIComponent(key.replace(/\+/g, ""));
+      const decodedValue = decodeURIComponent(value.replace(/\+/g, " "));
```
cjyuan: Why decode key and value differently?

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,14 @@
...
+  const counts = {};
+
+  if (!Array.isArray(items)) {
+    throw new Error("Input must be an array");
+  }
+
+  for (const item of items) {
+    counts[item] = (counts[item] || 0) + 1;
+  }
+  return counts;
```
```
```

Sprint-2/interpret/invert.js
```diff
@@ -10,20 +10,50 @@ function invert(obj) {
...
 
-// c) What does Object.entries return? Why is it needed in this program?
+/*
+a) What is the current return value when invert is called with { a : 1 }
+    *the value { key: 1 }
+
+b) What is the current return value when invert is called with { a: 1, b: 2 }
+    *the first pair is { key: 1 }  
+    *and the second pair the value is  { key: 'b' } that means the first pair is 
+    being overwritten by the second pair, so the final return value is { key: 'b' }
```
cjyuan: The second value is not 'b'.

Grajales-K: ohh, my bad, yes it's key: 2

Sprint-2/implement/querystring.js
Grajales-K: Hi @cjyuan,  I apologise for the mistake; I was experimenting with different options and have since made an update. Thank you for your understanding.

Sprint-2/implement/tally.js
Grajales-K: No, it didn't. It returned { toString: 'function toString() { [native code] }11' }.

This happens because a regular object {} inherits the toString method from Object.prototype. When the loop runs, it finds that function instead of undefined and tries to add 1 to it, breaking the count.

Using Object.create(null) fixes this because it creates an object with no inherited properties, so it treats 'toString' as a simple string key.

Thank you! I didn't know about this behaviour.

Grajales-K:
@cjyuan, thanks for your time reviewing my work.

cjyuan:
Changes look good.

illicitonion:

PR #1080, opened 2026-03-23
Birmingham | ITP-Jan | Roman Sanaye | Sprint 3 | Alarm-clock

RomanSanaye:
## Learners, PR Template

Self checklist

## Changelist

- Implemented the **Alarm Clock** assignment.
- Countdown works in **seconds** with proper `MM:SS` format.
- Countdown stops at `00:00` and **plays the alarm audio**.
- Stop button  pauses audio.
- Added **Jest setup** for DOM matchers to pass all tests.
- All **5 automated tests pass**.

## Questions

No questions.

### DIFF
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -1,4 +1,44 @@
-function setAlarm() {}
+// DOM elements
+const setAlarmBtn = document.getElementById("set");
+const stopAlarmBtn = document.getElementById("stop");
+const timeRemaining = document.getElementById("timeRemaining");
+
+// Interval variable for countdown
+let alarmInterval;
+
+// Your function to develop;
+function setAlarm() {
+  const timeValue = document.getElementById("alarmSet").value;
+  if (!timeValue || timeValue <= 0) {
+    alert("Please enter a number greater than 0");
+    return;
+  }
+
+  let totalSeconds = parseInt(timeValue);
+
+  // Stop previous countdown if any
+  if (alarmInterval) clearInterval(alarmInterval);
+
+  // Show initial countdown immediately
+  let minutes = Math.floor(totalSeconds / 60);
+  let seconds = totalSeconds % 60;
+  timeRemaining.innerText = `Time Remaining: ${String(minutes).padStart(2, "0")}:${String(seconds).padStart(2, "0")}`;
+
+  // Start countdown
+  alarmInterval = setInterval(() => {
+    totalSeconds--;
+
+    if (totalSeconds <= 0) {
+      clearInterval(alarmInterval); // stop countdown
+      timeRemaining.innerText = "Time Remaining: 00:00";
+      playAlarm(); // play alarm sound
+    } else {
+      let minutes = Math.floor(totalSeconds / 60);
+      let seconds = totalSeconds % 60;
+      timeRemaining.innerText = `Time Remaining: ${String(minutes).padStart(2, "0")}:${String(seconds).padStart(2, "0")}`;
+    }
+  }, 1000); // repeat every 1 second
+}
 
 // DO NOT EDIT BELOW HERE
 
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -3,13 +3,18 @@ const setAlarmBtn = document.getElementById("set");
 const stopAlarmBtn = document.getElementById("stop");
 const timeRemaining = document.getElementById("timeRemaining");
 let alarmInterval;
-let totalSeconds;
+
+// Update Display function;
+function updateDisplay(secondsValue) {
+  let minutes = Math.floor(secondsValue / 60);
+  let seconds = secondsValue % 60;
+  timeRemaining.innerText = `Time Remaining: ${String(minutes).padStart(2, "0")}:${String(seconds).padStart(2, "0")}`;
+}
 
 // reset function - returns app to clean initial state
 function resetAlarm() {
   function resetToInitialState() {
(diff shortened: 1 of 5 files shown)


Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,44 @@
...
+
+  let totalSeconds = parseInt(timeValue);
+
+  // Stop previous countdown if any
+  if (alarmInterval) clearInterval(alarmInterval);
+
+  // Show initial countdown immediately
+  let minutes = Math.floor(totalSeconds / 60);
+  let seconds = totalSeconds % 60;
+  timeRemaining.innerText = `Time Remaining: ${String(minutes).padStart(2, "0")}:${String(seconds).padStart(2, "0")}`;
```
cjyuan: This code is repeated on lines 36-38. 
Consider refactoring the code into a function.

cjyuan (commented):

Sprint-3/alarmclock/alarmclock.js
RomanSanaye: Hi @cjyuan , Thank you for your feedback. I have added a resetAlarm() function that resets all state variables and UI, and now setAlarm() calls it first so the user doesn't need to press Stop before starting a new countdown.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -2,55 +2,94 @@
...
+    document.getElementById("alarmSet").value = "";
+    stopAlarmBtn.style.display = "none";
+    setAlarmBtn.style.display = "inline-block"; // ADD THIS - show set button
+  }
+
+  function stopSound() {
+    if (typeof pauseAlarm === "function") {
+      pauseAlarm();
+    }
+  }
```
cjyuan, line 27: Common practice is to define the inner functions at the beginning of the function (before the if statement on line 11).

RomanSanaye: I have placed the inner function at the very beginning, thanks for guiding.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,16 +1,95 @@
...
+function setAlarm() {
+  const input = document.getElementById("alarmSet");
+  const timeValue = parseInt(input.value);
+  const hasActiveTimer = alarmInterval !== null;
+
+  function updateDisplay() {
+    let minutes = Math.floor(totalSeconds / 60);
+    let seconds = totalSeconds % 60;
+    timeRemaining.innerText = `Time Remaining: ${String(minutes).padStart(2, "0")}:${String(seconds).padStart(2, "0")}`;
+  }
```
cjyuan: To improve reusability and avoid sharing variables in the outer/global scope, consider introducing a parameter to this function and define it in the outer scope. 

You can then call `updateDisplay(0)` on lines 12, and keep `totalSeconds` as a local variable in `setAlarm()`.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,16 +1,95 @@
...
+  // Ensure clean initial state when page loads
+  resetAlarm();
+
   document.getElementById("set").addEventListener("click", () => {
     setAlarm();
   });
 
   document.getElementById("stop").addEventListener("click", () => {
     pauseAlarm();
+    resetAlarm();
```
cjyuan: Should you need to avoid modifying the code beyond the marker  
`// DO NOT EDIT BELOW HERE`

you can use `addEventListener()` to add additional callbacks above the marker.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,16 +1,95 @@
...
+
+  function updateDisplay() {
+    let minutes = Math.floor(totalSeconds / 60);
+    let seconds = totalSeconds % 60;
+    timeRemaining.innerText = `Time Remaining: ${String(minutes).padStart(2, "0")}:${String(seconds).padStart(2, "0")}`;
+  }
+
+  function startTimer() {
+    alarmInterval = setInterval(() => {
+      if (totalSeconds > 0) {
```
cjyuan: Why check if `totalSeconds > 0`? Isn't it always true?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,16 +1,95 @@
...
+let totalSeconds;
+
+// reset function - returns app to clean initial state
+function resetAlarm() {
+  function resetToInitialState() {
+    totalSeconds = 0;
+    timeRemaining.innerText = "Time Remaining: 00:00";
+    document.getElementById("alarmSet").value = "";
+    stopAlarmBtn.style.display = "none";
+    setAlarmBtn.style.display = "inline-block"; // ADD THIS - show set button
```
cjyuan, line 20: Could also consider 
1. using `visibility: hidden`  (suggestion: look up the difference between `display: none` and `visibility: false`
2. disable/enable the buttons by setting their `disabled` property


(No change required)

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,17 +1,94 @@
...
+  // Ensure clean initial state when page loads
+  resetAlarm();
+
   document.getElementById("set").addEventListener("click", () => {
     setAlarm();
   });
-
-  document.getElementById("stop").addEventListener("click", () => {
-    pauseAlarm();
-  });
```
cjyuan, line 14: You don't have to delete this. You can just perform `resetAlarm()` in the event listener you introduced on line 76.

cjyuan (commented):
Changes look good.

illicitonion:

PR #1087, opened 2026-03-24
London | 26 ITP January | Mouawia Elkhalifa | Sprint 3 | Alarm Clock

MouawiaElkhalifa:
## Learners, PR Template

Self checklist

## Changelist

In this PR, I have completed the Alarm Clock assignment for Sprint 3.

### DIFF
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -1,29 +1,36 @@
-function setAlarm() {
-  let timeRemaining = document.getElementById("alarmSet").value;
-  const timeDisplay = document.getElementById("timeRemaining");
+let countdown = null; 
 
-  // Create a reusable function to update the text on the screen
-  const updateDisplay = (time) => {
-    let minutes = Math.floor(time / 60)
-      .toString()
-      .padStart(2, "0");
-    let seconds = (time % 60).toString().padStart(2, "0");
-    timeDisplay.innerText = "Time Remaining: " + minutes + ":" + seconds;
+function setAlarm() {
+  let secondsLeft = parseInt(document.getElementById("alarmSet").value, 10); 
+  const timeDisplay = document.getElementById("timeRemaining"); 
+
+  if (isNaN(secondsLeft) || secondsLeft <= 0) { 
+    timeDisplay.innerText = "Time Remaining: 00:00";
+    return;
+  }
+
+  if (countdown !== null) { 
+    clearInterval(countdown);
+  }
+
+  const updateDisplay = (time) => { 
+    let minutes = Math.floor(time / 60).toString().padStart(2, "0"); 
+    let seconds = (time % 60).toString().padStart(2, "0"); 
+    timeDisplay.innerText = "Time Remaining: " + minutes + ":" + seconds; 
   };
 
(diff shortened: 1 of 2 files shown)


Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,30 @@
-function setAlarm() {}
+function setAlarm() {
+  let timeRemaining = document.getElementById("alarmSet").value;
```
cjyuan: Some unusual input values can make your app behave abnormally. Can you add code to sanitise the input?

cjyuan (commented):
- Code works fine if a user only clicks the "Set Alarm" button once.
Can you fix the issue?

Sprint-3/alarmclock/alarmclock.js
MouawiaElkhalifa: thanks! I have updated the code to sanitize the input using parseInt() and isNaN() to handle unusual values.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,37 @@
-function setAlarm() {}
+let countdown = null; 
+
+function setAlarm() {
+  let secondsLeft = parseInt(document.getElementById("alarmSet").value, 10); 
+  const timeDisplay = document.getElementById("timeRemaining"); 
+
+  if (isNaN(secondsLeft) || secondsLeft <= 0) { 
+    timeDisplay.innerText = "Time Remaining: 00:00";
```
cjyuan: Could consider calling `updateDisplay(0)` on line 8, trading few milliseconds for simpler code, and keep the logic of displaying time in one place.

cjyuan (commented):
(1) Currently when starting a new countdown, the application does not always return to a clean initial state, which can lead to inconsistent behavior between runs. (Hint: a user may not click the "Stop" button first before starting a new count down.)

You can consider introducing a dedicated reset function to return the app to a clean initial state to help ensure consistency.

(2) Can you make the "Stop" button to also stop the count down? You can add another "click" event listener to the button to avoid modifying the original code.

Sprint-3/alarmclock/alarmclock.js
MouawiaElkhalifa: Thanks 
I added a resetAlarm() function so every new countdown starts from a clean state.
I added another click listener to the Stop button so it also clears the active countdown.
I also moved the display formatting into updateDisplay() to keep the time rendering in one place.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -12,6 +56,13 @@ function setup() {
   document.getElementById("stop").addEventListener("click", () => {
     pauseAlarm();
   });
+
+  document.getElementById("stop").addEventListener("click", () => {
+    if (countdown !== null) {
+      clearInterval(countdown);
+      countdown = null;
+    }
```
cjyuan, line 64: Could also consider using `resetAlarm` as the callback, trading slight inefficiency for simplicity (and to keep the logic of resetting the alarm in one place). The code won't break when `pauseAlarm()` is called twice.

cjyuan (commented):
Changes look good. Well done.

illicitonion:

PR #1090, opened 2026-03-24
London | 26-ITP-Jan | Zadri Abdule | Sprint 1 | Data - Groups

Zadri415:
## Learners, PR Template

Self checklist

## Changelist
- fix: corrected calculateMedian in median.js and added tests
- feat: implemented dedupe with tests (dedupe.js, dedupe.test.js)
- feat: implemented findMax and tests (max.js, max.test.js)
- test: added tests for sum (sum.test.js)
- refactor: switched includes loop to for...of (includes.js)
- feat: added AoC 2018 Day 1 solution (solution.js)

## Questions
N/A

### DIFF
--- a/Sprint-1/implement/dedupe.test.js
+++ b/Sprint-1/implement/dedupe.test.js
@@ -16,7 +16,28 @@ E.g. dedupe([1, 2, 1]) returns [1, 2]
 // Given an empty array
 // When passed to the dedupe function
 // Then it should return an empty array
-test.todo("given an empty array, it returns an empty array");
+test("given an empty array, it returns an empty array", () => {
+  expect(dedupe([])).toEqual([]);
+});
+
+test("array with no duplicates returns a copy", () => {
+  const arr = [1, 2, 3];
+  const out = dedupe(arr);
+  expect(out).toEqual(arr);
+  expect(out).not.toBe(arr); // should be a new array
+});
+
+test("removes duplicates while preserving first occurrence order", () => {
+  const input = [5, 1, 1, 2, 3, 2, 5, 8];
+  expect(dedupe(input)).toEqual([5, 1, 2, 3, 8]);
+});
+
+test("does not mutate the original array", () => {
+  const arr = ["a", "a", "b"];
+  const copy = arr.slice();
+  dedupe(arr);
+  expect(arr).toEqual(copy);
+});
 
 // Given an array with no duplicates
 // When passed to the dedupe function
@@ -24,5 +45,5 @@ test.todo("given an empty array, it returns an empty array");
 
 // Given an array of strings or numbers
 // When passed to the dedupe function
-// Then it should return a new array with duplicates removed while preserving the 
+// Then it should return a new array with duplicates removed while preserving the
 // first occurrence of each element from the original array.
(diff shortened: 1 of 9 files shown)


Sprint-1/implement/dedupe.test.js
```diff
@@ -16,13 +16,34 @@ E.g. dedupe([1, 2, 1]) returns [1, 2]
...
+test("given an empty array, it returns an empty array", () => {
+  expect(dedupe([])).toEqual([]);
+});
+
+test("array with no duplicates returns a copy", () => {
+  const arr = [1, 2, 3];
+  const out = dedupe(arr);
+  expect(out).toEqual(arr);
+  expect(out).not.toBe(arr); // should be a new array
+});
```
cjyuan, line 28: There is a chance that, even though `out` has incorrect elements (for example, `[]`), 
the two tests on lines 27-28 could still pass. Can you figure out why, and then fix the tests accordingly?

Sprint-1/implement/max.test.js
```diff
@@ -16,28 +16,47 @@ const findMax = require("./max.js");
...
+test("given an array with non-number values, returns the max and ignores non-numeric values", () => {
+  expect(findMax(["hey", 10, "hi", 60, 10])).toBe(60);
+});
 
 // Given an array with only non-number values
 // When passed to the max function
 // Then it should return the least surprising value given how it behaves for all other inputs
+test("given an array with only non-number values, returns -Infinity", () => {
+  expect(findMax(["hey", "hi", "there"])).toBe(-Infinity);
+});
```
cjyuan, line 62: When a string representing a valid numeric literal (for example, `"300"`) is compared to a number, 

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,13 @@
 function sum(elements) {
+  if (!Array.isArray(elements)) return 0;
+
+  let total = 0;
+  for (const el of elements) {
+    if (typeof el === "number") {
+      total += el;
+    }
+  }
```

```
sum([NaN, 1]);
sum([Infinity, -Infinity]);
```

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,46 @@ const sum = require("./sum.js");
...
-
+test("given an array with negative numbers, returns the correct sum", () => {
+  expect(sum([-10, 20, -5])).toBe(5);
+});
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
+test("given an array with decimal numbers, returns the correct sum", () => {
+  expect(sum([1.5, 2.5, 3.5])).toBe(7.5);
+});
```
cjyuan, line 44: Decimal numbers in most programming languages (including JS) are internally represented in "floating point number" format. Floating point arithmetic is not exact. For example, the result of `46.5678 - 46 === 0.5678` is false because `46.5678 - 46` only yield a value that is very close to `0.5678`. Even changing the order in which the program add/subtract numbers can yield different values.

So the following could happen
```javascript

```

Suggestion: Look up 

Sprint-1/implement/max.test.js
Zadri415: Thanks. I'll add a test that includes a numeric-looking string("300") to ensure those strings are ignored, and implement findMax to only consider real number values.

Sprint-1/implement/sum.js
Zadri415: Thanks - great question. I chose to ignore non finite values so the function only adds finite numbers. That means:
sum[NaN, 1] -> 1
sum [Infinity, -Infinity]) -> 0

illicitonion:

PR #1101, opened 2026-03-25
London | 26-ITP-Jan | Maryanne Mosonik | Sprint 2 | Module Data Groups

Maryanne-K:
Self checklist

## Changelist

Corrected and completed the implementations and jest test cases.


Sprint-2/debug/recipe.js
```diff
@@ -12,4 +12,4 @@ const recipe = {
 
 console.log(`${recipe.title} serves ${recipe.serves}
   ingredients:
-${recipe}`);
+${recipe.ingredients}`);
```
cjyuan: The requirement on line 4 has not yet been met:
> Each ingredient should be logged on a new line

Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,8 @@
-function contains() {}
+function contains(obj, key) {
+    if (typeof obj !== 'object' || obj === null || Array.isArray(obj)) {
+        return false; 
+    }
+    return key in obj;
```
```
```

Sprint-2/implement/querystring.js
```diff
@@ -6,10 +6,15 @@ function parseQueryString(queryString) {
...
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
-    queryParams[key] = value;
+    const index = pair.indexOf("=");
+    if (index === -1) {
+      queryParams[pair] = "";
+    } else {
+      const key = pair.substring(0, index);
+      const value = pair.substring(index + 1);
+      queryParams[key] = value;
```
cjyuan, line 15: Note: (No change required)

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,16 @@
...
+        throw new Error("Input must be an array");
+    }
+    const counts = {};
+    for (const item of arr) {
+        if (counts[item]) {
+            counts[item] += 1;
+        } else {
+            counts[item] = 1;
+        }
+    }
```
```
```

cjyuan (commented):
`test.todo(" ... ")` is just a placeholder to serve as a reminder for "what needs to be done".
In this exercise, that means "implementing the corresponding Jest tests described in the comments."

Can you replace them (in all `.test.js` files) with actual Jest tests?

Sprint-2/debug/recipe.js
Maryanne-K: Apologies taking a while to respond to the comments CJYuan, I've been really busy. I have now made the corrections.

Sprint-2/implement/contains.js
Maryanne-K: Corrected to Object.hasOwn to ensure only properties that were added were checked.

Sprint-2/implement/tally.js
Maryanne-K: I have used the following: 

 const counts = Object.create(null);

I have updated my file. Thank you

cjyuan:
Can you also address this comment (about implementing the Jest test in `.test.js` files):
https://github.com/CodeYourFuture/Module-Data-Groups/pull/1101#pullrequestreview-4013841810

The other changes are good.

Maryanne-K:
> Can you also address this comment (about implementing the Jest test in `.test.js` files): [#1101 (review)](https://github.com/CodeYourFuture/Module-Data-Groups/pull/1101#pullrequestreview-4013841810)
> 
> The other changes are good.

Hey CJ, replaced all the test todo in the files. Kindly check, thank you.

Sprint-2/implement/contains.test.js
```diff
@@ -16,22 +16,29 @@ as the object doesn't contains a key of 'c'
 // Given a contains function
 // When passed an object and a property name
 // Then it should return true if the object contains the property, false otherwise
-test.todo("contains returns true for existing property");
-test.todo("contains returns false for non-existent property");
+test("contains returns true for existing property");
+expect (contains({a: 1, b: 2}, 'a')).toBe(true);
```
cjyuan: Have you tried running Jest test to see if these tests work?

Sprint-2/implement/contains.test.js
```diff
@@ -16,22 +16,29 @@ as the object doesn't contains a key of 'c'
...
-test.todo("contains returns false for non-existent property");
+test("contains returns false for non-existent property");
+expect (contains({a: 1, b: 2}, 'c')).toBe(false);
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
-test.todo("contains with invalid parameters returns false or throws error");
\ No newline at end of file
+test("contains with invalid parameters returns false or throws error");
+expect (contains([1, 2, 3], 'a')).toBe(false);
```
cjyuan: Arrays are objects, with their indices acting as keys. `"a"` is not a key of an array. A proper test should use a valid key to ensure the function returns `false` specifically because the input is an array, not because the key is missing.

You could also test invalid parameters such as `null`, string, and `undefined`.

Sprint-2/implement/contains.test.js
Maryanne-K: Hi, I have modified the tests, it now passes.

Sprint-2/implement/contains.test.js
Maryanne-K: I have amended and now added these test cases.

Sprint-2/implement/contains.test.js
```diff
@@ -16,20 +16,45 @@ as the object doesn't contains a key of 'c'
...
 // Then it should return false
-
+test("contains returns false for non-existent property", () => {
+  expect(contains({a: 1, b: 2}, 'c')).toBe(false);
+});
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+test("contains with invalid input returns false", () => {
+  expect(contains([], 'a')).toBe(false);
```
cjyuan, line 47: Your function is correct, but we should prepare the tests not only to verify our current implementation, but also to ensure that future changes do not alter the function's expected behavior.

This test `expect(contains([], 'a')).toBe(false)` cannot confirm that the function correctly returns `false when the first argument is an array. This is because `contains([], "a")` could also return `false` simply because "a" is not a key of the array.

**Arrays are objects, with their indices acting as keys.** A proper test should **use a non-empty array along with a valid key** to ensure the function returns `false` specifically because the input is an array, not because the key is missing.

Note: Strings are not objects but their indices also act as their keys.

cjyuan (commented):
Other changes are good. I will mark this PR as complete first.

illicitonion:

PR #1105, opened 2026-03-26
London | 26-ITP-Jan | Carlos Abreu |Sprint 1 | Data Module Groups

carlosyabreu:
## Learners, PR Template

Self checklist

## Changelist
PR for Sprint 1 of Data Module Group Coursework

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -6,9 +6,32 @@
 // or 'list' has mixed values (the function is expected to sort only numbers).
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  // Check if input is an array
+  if (!Array.isArray(list)) {
+    return null;
+  }
+
+  // Filter out non-numbers (keep only numbers)
+  const numbers = list.filter(item => typeof item === 'number' && !isNaN(item));
+
+  // If no numbers in the array, return null
+  if (numbers.length === 0) {
+    return null;
+  }
+
+  // Sort the numbers in ascending order
+  const sorted = [...numbers].sort((a, b) => a - b);
+
+  const middleIndex = Math.floor(sorted.length / 2);
+
+  // If length is odd, return the middle element
+  if (sorted.length % 2 !== 0) {
+    return sorted[middleIndex];
+  }
+  // If length is even, return the average of the two middle elements
+  else {
+    return (sorted[middleIndex - 1] + sorted[middleIndex]) / 2;
+  }
 }
 
 module.exports = calculateMedian;
--- a/Sprint-1/implement/max.js
+++ b/Sprint-1/implement/max.js
@@ -1,4 +1,15 @@
+
 function findMax(elements) {
+    // Filter out non-numeric values
+    const numbers = elements.filter(element => typeof element === 'number');
+
+    // If no numbers in array, return -Infinity
+    if (numbers.length === 0) {
+        return -Infinity;
+    }
+
+    // Find and return the maximum value
+    return Math.max(...numbers);
 }
 
 module.exports = findMax;
--- a/Sprint-1/implement/max.test.js
+++ b/Sprint-1/implement/max.test.js
@@ -1,43 +1,62 @@
-/* Find the maximum element of an array of numbers
-
-In this kata, you will need to implement a function that find the largest numerical element of an array.
-
-E.g. max([30, 50, 10, 40]), target output: 50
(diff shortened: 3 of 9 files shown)


Sprint-1/fix/median.js
```diff
@@ -6,9 +6,32 @@
...
+  // Filter out non-numbers (keep only numbers)
+  const numbers = list.filter(item => typeof item === 'number' && !isNaN(item));
+
+  // If no numbers in the array, return null
+  if (numbers.length === 0) {
+    return null;
+  }
+
+  // Sort the numbers in ascending order
+  const sorted = [...numbers].sort((a, b) => a - b);
```
cjyuan, line 23: Is it necessary to make a copy of `numbers` before sorting?

Sprint-1/implement/max.js
```diff
@@ -1,4 +1,15 @@
+
 function findMax(elements) {
+    // Filter out non-numeric values
+    const numbers = elements.filter(element => typeof element === 'number');
+
```
cjyuan, line 5: What do you expect from the following function calls (on extreme cases)?

```
findMax([NaN])
findMax([0, NaN, 1])
```

Sprint-1/implement/max.test.js
```diff
@@ -1,43 +1,62 @@
...
+});
 
 // Given an array with non-number values
 // When passed to the max function
 // Then it should return the max and ignore non-numeric values
+test("given an array with non-number values, returns the max and ignores non-numeric values", () => {
+    expect(findMax(['hey', 10, 'hi', 60, 10])).toBe(60);
+    expect(findMax([20, 'abc', 30, 'def', 40])).toBe(40);
+    expect(findMax(['x', 'y', 'z', 100, 50])).toBe(100);
+});
```
cjyuan, line 53: When a string representing a valid numeric literal (for example, `"300"`) is compared to a number, 

Sprint-1/fix/median.js
carlosyabreu: Yes, it is necessary to make a copy of numbers before sorting in this implementation.
By using the spread operator [...numbers], we create a shallow copy of the filtered numbers array. 
This ensures:
We never mutate the filtered array.
The code is more predictable and follows functional programming good practices.
And it communicates that we're working with a copy for sorting purposes.

Sprint-1/implement/max.js
carlosyabreu: For findMax([NaN])
The filter typeof element === 'number' would include NaN because typeof NaN === 'number' returns true.
So the numbers would be [NaN], and then trying to find the maximum would return NaN.
The expected return is NaN.

For findMax([0, NaN, 1])
Similarly, NaN passes the typeof check, so numbers would be [0, NaN, 1].
When using Math.max(...numbers) or a custom max function, NaN in the array will cause the result to be NaN, because any arithmetic operation involving NaN propagates NaN.
Expected return is NaN.

Does the function return the value expected?
Probably not. The test cases in the original problem likely expect to filter out NaN values as well, since they're not valid numbers for calculating a median.
The comment in the original code says:
"list has mixed values (the function is expected to sort only numbers)"
This implies that NaN should likely be excluded from the calculation, not included.

Sprint-1/implement/max.test.js
carlosyabreu: Based on the test cases provided, if we include a string like "300" in the test:
expect(findMax([20, "300", 40])).toBe(???)

The current implementation likely uses something like:
const numbers = elements.filter(element => typeof element === 'number');

This would filter out "300" because typeof "300" === "string", not "number".
So numbers would be [20, 40], and findMax would return 40.

Sprint-1/fix/median.js
cjyuan: "Good practice" does not mean "it is necessary". Should performance or resource consumption become a concern, we may need to sacrifice good practices.

Your reasoning is sound. Your code, your call. 

Note: For non-mutable sorting, you can also use `Array.prototype.toSorted()`.

Sprint-1/implement/max.test.js
cjyuan: We write tests not only to verify our current implementation, but also to ensure that future changes do not alter the function's expected behavior. In some cases, developers even rely on tests to understand the expected behavior of a function.

Your function implementation is correct, but the test would not fail when someone re-implement the function as:

```javascript
function findMax(elements) {
    // Change strings to equivalent numeric values before checking NaN
    const onlyNumbers = elements.map(Number).filter((el)=>!Number.isNaN(el));
    return Math.max(...onlyNumbers);
}
```

cjyuan:
Your explanation is clear. Changes is optional.

illicitonion:

PR #1106, opened 2026-03-26
London | 26-ITP-Jan | Xuanming Hu | Sprint 2 | Data Groups

Samual-Hu:
## Learners, PR Template

Self checklist

## Changelist

Completed all mandatory exercises in the Sprint 2 directory.

## Questions

N/A

### DIFF
--- a/Sprint-2/implement/querystring.js
+++ b/Sprint-2/implement/querystring.js
@@ -6,7 +6,8 @@ function parseQueryString(queryString) {
   const keyValuePairs = queryString.split("&");
 
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
+    const [key, ...rest] = pair.split("=");
+    const value = rest.join("=");
     queryParams[key] = value;
   }
 
--- a/Sprint-2/implement/querystring.js
+++ b/Sprint-2/implement/querystring.js
@@ -6,6 +6,7 @@ function parseQueryString(queryString) {
   const keyValuePairs = queryString.split("&");
 
   for (const pair of keyValuePairs) {
+    if (!pair) continue;
     const [key, ...rest] = pair.split("=");
     const value = rest.join("=");
     queryParams[key] = value;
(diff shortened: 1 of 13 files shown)


Sprint-2/implement/querystring.js
```diff
@@ -6,7 +6,8 @@ function parseQueryString(queryString) {
   const keyValuePairs = queryString.split("&");
 
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
+    const [key, ...rest] = pair.split("=");
+    const value = rest.join("=");
     queryParams[key] = value;
```
cjyuan, line 12: For this function call, does your function return the value you expect?

```
```

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,15 @@
...
+function tally(itemsList) {
+  if (!Array.isArray(itemsList)) {
+    throw new Error("Input must be an array");
+  }
+
+  const counts = {};
+
+  for (const item of itemsList) {
+    counts[item] = (counts[item] || 0) + 1;
+  }
```
```
```

Sprint-2/implement/querystring.js
Samual-Hu: `&&` produces an empty string after splitting, causing an unwanted `{ "": "" }` entry. Fixed by adding `if (!pair) continue;` to skip empty pairs.

Sprint-2/implement/tally.js
Samual-Hu: Using `{}` inherits Object prototype properties like `toString`, so `tally(["toString", "toString"])` gives unexpected results. Fixed by using `Object.create(null)` to create a truly empty object with no inherited properties.

Samual-Hu:
@cjyuan Thanks for the feedback, I've fixed the edge cases for both querystring.js and tally.js

cjyuan:
Changes look good. Well done.

illicitonion:

PR #1107, opened 2026-03-26
GLASGOW | Jan-26-ITP | Prakash Dcosta | Sprint 3 | Quote Generator

dcostaprakash:
Changelist
I have titled my PR with Region | Cohort | FirstName LastName | Sprint | Assignment Title
My changes meet the requirements of the task
I have tested my changes
My changes follow the style guide
my coursework

### DIFF
--- a/Sprint-3/quote-generator/index.html
+++ b/Sprint-3/quote-generator/index.html
@@ -10,9 +10,14 @@
   <body>
     <div class="page-wrapper">
     <h1>Your Quote for the day</h1>
+    <!-- Quote container added -->
     <div class="quote-container">
-    <p id="quote"></p>
-    <p id="author"></p>
+
+      <!-- Quote text section -->
+      <div class="quote-text">
+        <p id="quote"></p>
+        <p id="author"></p>
+      </div>
     <button type="button" id="new-quote">New quote</button>
     </div>
     </div>
--- a/Sprint-3/quote-generator/index.html
+++ b/Sprint-3/quote-generator/index.html
@@ -18,7 +18,7 @@
         <p id="quote"></p>
         <p id="author"></p>
       </div>
-    <button type="button" id="new-quote">New quote</button>
+      <button type="button" id="new-quote">New quote</button>
     </div>
     </div>
   </body>
--- a/Sprint-3/quote-generator/index.html
+++ b/Sprint-3/quote-generator/index.html
@@ -3,13 +3,18 @@
   <head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <title>Title here</title>
+    <title>Quote generator app</title>
+    <link rel="stylesheet" href="style.css" />
     <script defer src="quotes.js"></script>
   </head>
   <body>
-    <h1>hello there</h1>
+    <div class="page-wrapper">
+    <h1>Your Quote for the day</h1>
+    <div class="quote-container">
     <p id="quote"></p>
     <p id="author"></p>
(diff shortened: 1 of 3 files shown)


Sprint-3/quote-generator/index.html
```diff
@@ -3,13 +3,18 @@
...
   </head>
   <body>
-    <h1>hello there</h1>
+    <div class="page-wrapper">
+    <h1>Your Quote for the day</h1>
+    <div class="quote-container">
     <p id="quote"></p>
     <p id="author"></p>
     <button type="button" id="new-quote">New quote</button>
+    </div>
```
Luro91, line 22: How could you make it more clear what HTML elements are nested inside the new divs?

Sprint-3/quote-generator/style.css
Luro91, line 1: An advanced topic regarding UX:
At the moment the boxes containing quotes have various sizes depending on the text. This also leads to the button changing the position. (it would be nice if the button would stay in position and the user would not have to move the mouse to click it again).
How could you accomplish this?

Image
Image

Sprint-3/quote-generator/index.html
dcostaprakash: Got it. I can add comments and i can make it visually better. I have modified it. Thank you

Sprint-3/quote-generator/style.css
dcostaprakash: Understood. I made the following changes to incorporate it. I increased the quote-container height and max-width, made the display to flex and further the quote-text i made it to overflow with minimum heing 0 so that it can scroll and not disturb the button. Thanks

Sprint-3/quote-generator/index.html
Luro91: I was referring to code indentation. With the correct indentation the comments are not needed. indentation is a standard that makes it easy to see visually which elements are on which hierarchy. Tools like the IDE also allow code folding based on indentation

Sprint-3/quote-generator/style.css
Luro91: The box size still changes in width and make the button move in horizontal direction.
This is an optional improvement that you could do. In this case the list of quotes is defined so you can design the box for the longest quote and it will fit all of them.

Screenshot from 2026-04-05 07-18-33
Screenshot from 2026-04-05 07-18-26

Luro91:
I only now noticed your commit messages. When I read the commit messages I don't know what the change was.
Commit messages should give everyone an understanding on what was changed in the commit without having to look at the actual file changes. It's also good practice to separate changes into separate commits. (For example one for adjusting the box size and one for changing the html structure) 
You can also give context in the message about the reason for the change. 

I recommend reading this article on commit messages: https://chris.beams.io/git-commit

Sprint-3/quote-generator/style.css
dcostaprakash: I made the modifications and tested it. Pl see if its alright now. Thanks

Sprint-3/quote-generator/index.html
dcostaprakash: I have made the change now. Pl check. I will also go go through IDE thoroughly. Thank you

Sprint-3/quote-generator/style.css
Luro91: Looks good now

Sprint-3/quote-generator/index.html
Luro91: The indentation looks good now. If you run prettier it should always do the correct formatting for you and you do not have to think about it

illicitonion:

PR #1109, opened 2026-03-26
Cape Town | 26-ITP-Jan | Pretty Taruvinga | Sprint 2 | Data Groups

Pretty548:
## Learners, PR Template

Self checklist

## Changelist

- Fixed parseQueryString implementation and handled edge cases
- Fixed contains function bug
- Implemented tally function with validation
- Fixed calculateMedian to handle invalid inputs


Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,7 @@
-function contains() {}
+function contains(obj, key) {
+  if (!obj) return false;
```
cjyuan: Supposedly, when `obj` is one of the following types of value, the function should return `false`.

> `[1,2,3]`, `123`, `true`, `"string"`

Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,7 @@
...
+
+  return Object.prototype.hasOwnProperty.call(obj, key);
```
cjyuan: Can also consider `Object.hasOwn()`.

Sprint-2/implement/contains.test.js
```diff
@@ -1,35 +1,36 @@
...
+  expect(contains({ a: 1, b: 2 }, "c")).toBe(false);
+});
 
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+test("returns false for an array", () => {
+  expect(contains([], "a")).toBe(false);
+});
+
```
cjyuan, line 30: This test does not yet confirm that the function correctly returns false when the first argument is an array. 

Sprint-2/implement/querystring.js
```diff
@@ -1,16 +1,24 @@
...
+    const index = pair.indexOf("=");
+
+    if (index === -1) {
+      result[pair] = "";
+    } else {
+      const key = pair.slice(0, index);
+      const value = pair.slice(index + 1);
+      result[key] = value;
+    }
+  });
```
cjyuan, line 19: For the following function call, does your function return the value you expect?

```
```

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,19 @@
...
+
+  const result = {};
+
+  items.forEach((item) => {
+    if (result[item]) {
+      result[item]++;
+    } else {
+      result[item] = 1;
+    }
+  });
```
```
```

Sprint-2/interpret/invert.js
```diff
@@ -27,3 +27,27 @@ function invert(obj) {
...
+  expect(invert({})).toEqual({});
+});
+
+test("handles non-string values as keys in the inverted object", () => {
+  expect(invert({ a: 1, b: true, c: null })).toEqual({
+    1: "a",
+    true: "b",
+    null: "c",
+  });
+});
```
cjyuan: Shouldn't this be implemented in `invert.test.js`?

Sprint-2/implement/contains.js
Pretty548: Good catch! I updated the function to ensure it only accepts plain objects by checking for non-object types, null, and arrays. It now correctly returns false for values like arrays, numbers, Booleans, and strings.

Sprint-2/implement/contains.js
Pretty548: Thanks! I updated the implementation to use Object.hasOwn() and ensured the function returns false for non-object inputs instead of throwing an error.

Sprint-2/implement/contains.test.js
Pretty548: Good point! I’ve updated the test to use a non-empty array with a valid index as the key, so it confirms the function returns false specifically because the input is an array.

Sprint-2/implement/querystring.js
Pretty548: Yes, the function correctly handles this case. The double "&&" creates an empty string when splitting, and the check `if (!pair) return;` ensures it is ignored. The function returns the expected result with both key-value pairs.

Sprint-2/implement/tally.js
Pretty548: Good catch! I updated the implementation to use Object.create(null) so the result object has no inherited properties. This ensures keys like "toString" are handled correctly.

Sprint-2/interpret/invert.js
Pretty548: Good catch, thanks! You're right — these tests should live in invert.test.js. I'll move them there to keep the test structure consistent and separated from implementation.

Sprint-2/implement/tally.js
cjyuan: The proposed solution is sound, but did you forget to push your changes to GitHub?

cjyuan (commented):
There are also some unanswered questions in `interpret/invert.js`.

Sprint-2/implement/tally.js
Pretty548: I had forgotten, sorry

cjyuan:
Looks good.

illicitonion:

PR #1110, opened 2026-03-26
Manchester| 26-ITP-Jan | fonime Edak | Sprint 2 | Data-groups

Ofonimeedak:
Self checklist

This PR contains fixed bug codes with  test cases for sprint 2
Please kindly find this as the  link to my closed pull request for all commits
https://github.com/CodeYourFuture/Module-Data-Groups/pull/1102

### DIFF
--- a/Sprint-2/debug/author.js
+++ b/Sprint-2/debug/author.js
@@ -12,6 +12,6 @@ const author = {
   alive: true,
 };
 
-for (const value in author) {
+for (const value of Object.values(author)) {
   console.log(value);
 }
--- a/Sprint-2/debug/author.js
+++ b/Sprint-2/debug/author.js
@@ -1,5 +1,6 @@
-// Predict and explain first...
-
+//Predict and explain first...
+// The loop method used is for arrays and not for object literals.
+// for object literals the "in" key word is used to access the keys.
 // This program attempts to log out all the property values in the object.
 // But it isn't working. Explain why first and then fix the problem
 
@@ -11,6 +12,6 @@ const author = {
   alive: true,
 };
 
-for (const value of author) {
+for (const value in author) {
   console.log(value);
 }
--- a/Sprint-2/debug/recipe.js
+++ b/Sprint-2/debug/recipe.js
@@ -1,15 +1,17 @@
 // Predict and explain first...
-
+// The program cant log the ingredient because the object literal was not reference inside the console.log()
 // This program should log out the title, how many it serves and the ingredients.
 // Each ingredient should be logged on a new line
 // How can you fix it?
-
+ 
 const recipe = {
   title: "bruschetta",
   serves: 2,
   ingredients: ["olive oil", "tomatoes", "salt", "pepper"],
 };
 
-console.log(`${recipe.title} serves ${recipe.serves}
-  ingredients:
-${recipe}`);
+console.log(`${recipe.title} serves ${recipe.serves}`);
+
+for (const x of recipe.ingredients) {
+  console.log(x);
+}
--- a/Sprint-2/debug/recipe.js
+++ b/Sprint-2/debug/recipe.js
@@ -12,6 +12,6 @@ const recipe = {
 
 console.log(`${recipe.title} serves ${recipe.serves}`);
 
-for (const x of recipe.ingredients) {
-  console.log(x);
+for (const part of recipe.ingredients) {
+  console.log(part);
 }
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,17 +1,12 @@
 function contains(elements, itemKey) {
   if (!elements) return false;
-  if (Object.getOwnPropertyNames(elements).length === 0) {
-    return false;
-  }
-
   if (typeof elements !== "object" || Array.isArray(elements)) {
     return false;
   }
-  for (const key in elements) {
-    if (key === itemKey) {
-      return true;
-    }
+  if (Object.hasOwn(elements, itemKey)) {
+    return true;
   }
+
   return false;
 }
 
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,3 +1,18 @@
-function contains() {}
+function contains(elements, itemKey) {
+  if (!elements) return false;
+  if (Object.getOwnPropertyNames(elements).length === 0) {
+    return false;
+  }
+
+  if (typeof elements !== "object" || Array.isArray(elements)) {
+    return false;
+  }
+  for (const key in elements) {
+    if (key === itemKey) {
+      return true;
+    }
+  }
+  return false;
+}
 
 module.exports = contains;
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -13,23 +13,53 @@ as the object doesn't contains a key of 'c'
 
 // Acceptance criteria:
 
-// Given a contains function
-// When passed an object and a property name
-// Then it should return true if the object contains the property, false otherwise
+describe("contains()", () => {
+  // Given an object with properties
+  // When passed to contains with an existing property name
+  // Then it should return true
 
-// Given an empty object
-// When passed to contains
-// Then it should return false
-test.todo("contains on empty object returns false");
+  [{ elements: { a: 1, b: 2 }, itemKey: "a", expected: true }].forEach(
+    ({ elements, itemKey, expected }) =>
+      it(` should return ${expected} for key ${itemKey} in ${elements}`, () => {
+        expect(contains(elements, itemKey)).toEqual(expected);
+      })
+  );
 
-// Given an object with properties
-// When passed to contains with an existing property name
-// Then it should return true
+  // Given an empty object
+  // When passed to contains
+  // Then it should return false
 
-// Given an object with properties
-// When passed to contains with a non-existent property name
-// Then it should return false
+  [{ elements: {}, itemKey: "c", expected: false }].forEach(
+    ({ elements, itemKey, expected }) =>
+      it(`contains on empty object returns false`, () => {
+        expect(contains(elements, itemKey)).toEqual(expected);
+      })
+  );
 
-// Given invalid parameters like an array
-// When passed to contains
-// Then it should return false or throw an error
+  // Given an object with properties
+  // When passed to contains with a non-existent property name
+  // Then it should return false
+
+  [{ elements: { a: 1, b: 2 }, itemKey: "c", expected: false }].forEach(
+    ({ elements, itemKey, expected }) =>
+      it(`When passed to contains with non-existing property name, should return false`, () => {
+        expect(contains(elements, itemKey)).toEqual(expected);
+      })
+  );
+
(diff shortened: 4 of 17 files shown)


Sprint-2/debug/author.js
```diff
@@ -11,6 +12,6 @@ const author = {
   alive: true,
 };
 
-for (const value of author) {
+for (const value in author) {
   console.log(value);
 }
```
cjyuan: Have you ran your code to ensure it outputs the value of the properties? The values are `'Zadie', 'Smith', 'writer', 40, true`.

Sprint-2/debug/recipe.js
```diff
@@ -1,15 +1,17 @@
...
 };
 
-console.log(`${recipe.title} serves ${recipe.serves}
-  ingredients:
-${recipe}`);
+console.log(`${recipe.title} serves ${recipe.serves}`);
+
+for (const x of recipe.ingredients) {
+  console.log(x);
+}
```
cjyuan: (1) Better practice is to give every variable/constant a descriptive name. 
A common practice is to use the singular form of the array name (when the array name is in plural form).

(2) Here is an alternative worth exploring: 

Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,18 @@
...
+
+  if (typeof elements !== "object" || Array.isArray(elements)) {
+    return false;
+  }
+  for (const key in elements) {
+    if (key === itemKey) {
+      return true;
+    }
+  }
+  return false;
```
cjyuan: Can also explore `Object.hasOwn()`.

Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,18 @@
-function contains() {}
+function contains(elements, itemKey) {
+  if (!elements) return false;
+  if (Object.getOwnPropertyNames(elements).length === 0) {
+    return false;
+  }
```
cjyuan: This "relatively costly" check is optional.

Sprint-2/implement/contains.test.js
```diff
@@ -13,23 +13,53 @@ as the object doesn't contains a key of 'c'
...
+
+  [
+    {
+      elements: [3, 4, 5, "kofoworola", "Evan,"],
+      itemKey: "Evan",
+      expected: false,
+    },
+  ].forEach(({ elements, itemKey, expected }) =>
+    it(`Given invalid parameters it should return false`, () => {
+      expect(contains(elements.itemKey)).toEqual(expected);
```
- Currently, the first argument is `undefined` instead of an array.

Arrays are objects, with their indices acting as keys. `"Evan"` is not a valid key of an array.
A proper test should use a valid key to ensure the function returns `false` specifically because the input is an array, not because the key is missing.

Sprint-2/implement/lookup.js
```diff
@@ -1,5 +1,14 @@
...
+function createLookup(elements) {
+  if (!Array.isArray(elements) || !elements.every(Array.isArray)) {
+    return "Invalid input";
+  }
+  if (elements.length === 0) return "Invalid input";
+  let lookup = {};
+
+  for (x of elements) {
+    lookup[x[0]] = x[1];
+  }
```
cjyuan: `x` is not declared, and `x` is not a very descriptive name.

Sprint-2/implement/lookup.js
```diff
@@ -1,5 +1,14 @@
-function createLookup() {
-  // implementation here
+function createLookup(elements) {
+  if (!Array.isArray(elements) || !elements.every(Array.isArray)) {
+    return "Invalid input";
+  }
+  if (elements.length === 0) return "Invalid input";
+  let lookup = {};
```
cjyuan: `lookup` is not going to be reassigned a different value, so it is better to declare it using `const`.

Sprint-2/implement/querystring.js
```diff
@@ -3,13 +3,26 @@ function parseQueryString(queryString) {
   if (queryString.length === 0) {
     return queryParams;
   }
+  if (
+    typeof queryString !== "string" ||
+    !queryString.includes("=") ||
+    queryString === undefined
```
cjyuan: The check on line 9 is unnecessary.

Sprint-2/implement/querystring.js
```diff
@@ -3,13 +3,26 @@ function parseQueryString(queryString) {
   if (queryString.length === 0) {
     return queryParams;
   }
+  if (
+    typeof queryString !== "string" ||
+    !queryString.includes("=") ||
```
cjyuan: `key1&key2&key3` is considered a valid query string.

Sprint-2/implement/querystring.js
```diff
@@ -3,13 +3,26 @@ function parseQueryString(queryString) {
...
-    const [key, value] = pair.split("=");
-    queryParams[key] = value;
+    const arr = [];
+    const indexEquality = pair.indexOf("=");
+    if (indexEquality !== -1) {
+      const sliced = pair.slice(0, indexEquality);
+      arr.push(sliced);
+      const sliced2 = pair.slice(indexEquality + 1);
+      arr.push(sliced2);
+      const [key, value] = arr;
```
cjyuan: Why not just declare `sliced` and `sliced2` as `key` and `value` ?

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,21 @@
...
+  const frequency = {};
+  for (let i = 0; i < elements.length; i++) {
+    const item = elements[i];
+    if (frequency[item] === undefined) {
+      frequency[item] = 1;
+    } else {
+      frequency[item]++;
+    }
+    }
+    return frequency;
```

as recommended in 
?

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,21 @@
-function tally() {}
+function tally(elements) {
+    if (!Array.isArray(elements)) {
+        throw new Error("Invalid input ")
+    }
+      if (elements.length === 0) {
+        return {};
+      }
+  let count = 0;
+  const frequency = {};
```
```
```

Sprint-2/stretch/count-words.js
```diff
@@ -26,3 +26,17 @@
...
+  const wordArr = words.toLowerCase().split(" ");
+  const filteredWord = wordArr.map((ele) => ele.replace(/[.,!?']/g, ""));
+  const wordCount = {};
+  for (let word of filteredWord) {
+    if (wordCount[word] === undefined) {
+      wordCount[word] = 1;
+    } else {
+      wordCount[word]++;
+    }
+  }
```
cjyuan: Can you check if your function returns what you expect in the following function calls?

```
```

Sprint-2/stretch/count-words.js
```diff
@@ -26,3 +26,17 @@
...
+  console.log(wordCount);
```
cjyuan: Debugging code?

Sprint-2/implement/tally.js
Ofonimeedak: Thank you for your suggestion Object.create(null) works fine, I have read through

Sprint-2/stretch/count-words.js
Ofonimeedak: Yes, I was using it to  debug code on the console

Sprint-2/stretch/count-words.js
Ofonimeedak: Split by any white space
filter off non-alphabetical characters, and create an object using Object. create(null) to ensure the constructor reserved keyword is not used

Sprint-2/debug/author.js
Ofonimeedak: fixed using Object.values()

Sprint-2/debug/recipe.js
Ofonimeedak: descriptive name part

Sprint-2/implement/contains.js
Ofonimeedak: Object.hasOwn() returns true, use instead of looping

Sprint-2/implement/contains.js
Ofonimeedak: checks removed, dead code

Sprint-2/implement/contains.test.js
Ofonimeedak: .dot instead of comma,

Sprint-2/implement/querystring.js
Ofonimeedak: Refactor code, using key and values as variable declarations

Sprint-2/implement/contains.test.js
```diff
@@ -13,23 +13,53 @@ as the object doesn't contains a key of 'c'
...
+      elements: [3, 4, 5, "kofoworola", "Evan,"],
+      itemKey: "Evan",
+      expected: false,
+    },
+  ].forEach(({ elements, itemKey, expected }) =>
+    it(`Given invalid parameters it should return false`, () => {
+      expect(contains(elements, itemKey)).toEqual(expected);
+    })
+  );
+});
```
cjyuan, line 65: This test still cannot not yet confirm that the function correctly returns false when the first argument is an array.

Arrays are objects, with their **indices acting as keys**. "Evan" is not a valid key of an array.
`Object.hasOwn(["Evan"], "0")` evaluates to `true`.
`Object.hasOwn(["Evan"], "Evan")` evaluates to `false`.

A proper test should use a valid key to ensure the function returns false specifically because the input is an array, not because the key is missing.

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,23 @@
...
+  for (let i = 0; i < elements.length; i++) {
+    const item = elements[i];
+    if (frequency[item] === undefined) {
+      frequency[item] = 1;
+    } else {
+      frequency[item]++;
+    }
+  }
+  console.log(frequency);
+  return frequency;
```
cjyuan, line 18: There are some dead code and unnecessary code in this function. Can you remove them?

cjyuan (commented):
4 of the inline comments in my previous review were hidden, and you probably missed them. Can you also address those comment?

Image

cjyuan:
What about the 4 unaddressed comments in the first review?

Ofonimeedak:
I searched  I thought it was the one on look up.js that I have addressed

Sprint-2/implement/querystring.js
```diff
@@ -3,13 +3,18 @@ function parseQueryString(queryString) {
   if (queryString.length === 0) {
     return queryParams;
   }
+  if (typeof queryString !== "string" || !queryString.includes("="))
```
cjyuan: > `!queryString.includes("=")`

This check would reject `key1&key2&key3` but accept `key1=&key2&key3`.

Ofonimeedak: Thank you for this edge case. It got me thinking 2 things: whether I should consider such input as invalid, or I should return an empty object literal.
I had to consider the latter to keep the loop running

cjyuan: You should change your code to meet your expectation (instead of the other way around). Since the spec wasn't clear how the function is expected to handle empty key or value, you code is good.

illicitonion:

PR #1111, opened 2026-03-27
London | 26-ITP-Jan | Zadri Abdule | Sprint 2 | Data - Groups

Zadri415:
## Learners, PR Template

Self checklist

Sprint-2: add/fix utility functions and tests (contains, lookup, querystring, tally, invert, till)

## Questions
N/A

### DIFF
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,3 +1,13 @@
-function contains() {}
+function contains(obj, prop) {
+  // Reject null/undefined, arrays, and non-objects
+  if (obj === null || obj === undefined) return false;
+  if (Array.isArray(obj)) return false;
+  if (typeof obj !== "object") return false;
+
+  // prop must be provided
+  if (prop === undefined || prop === null) return false;
+
+  return Object.prototype.hasOwnProperty.call(obj, prop);
+}
 
 module.exports = contains;
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -20,16 +20,30 @@ as the object doesn't contains a key of 'c'
 // Given an empty object
 // When passed to contains
 // Then it should return false
-test.todo("contains on empty object returns false");
-
+test("contains on empty object returns false", () => {
+  expect(contains({}, "anyKey")).toBe(false);
+});
 // Given an object with properties
 // When passed to contains with an existing property name
 // Then it should return true
-
+test("contains on object with existing property return true", () => {
+  const obj = { a: 1, b: 2 };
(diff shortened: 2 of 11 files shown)


Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,13 @@
...
+function contains(obj, prop) {
+  // Reject null/undefined, arrays, and non-objects
+  if (obj === null || obj === undefined) return false;
+  if (Array.isArray(obj)) return false;
+  if (typeof obj !== "object") return false;
+
+  // prop must be provided
+  if (prop === undefined || prop === null) return false;
+
+  return Object.prototype.hasOwnProperty.call(obj, prop);
```
cjyuan, line 10: Can also explore `Object.hasOwn()`.

Sprint-2/implement/contains.test.js
```diff
@@ -20,16 +20,30 @@ as the object doesn't contains a key of 'c'
...
+test("contains on object with non-existent property return false", () => {
+  const obj = { a: 1, b: 2 };
+  expect(contains(obj, "c")).toBe(false);
+});
 
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+test("contains on invalid parameters returns false", () => {
+  expect(contains([], "a")).toBe(false);
```

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,9 @@
-function tally() {}
+function tally(items) {
+  if (!Array.isArray(items)) throw new Error("Invalid input");
+  return items.reduce((acc, it) => {
+    acc[it] = (acc[it] || 0) + 1;
+    return acc;
+  }, {});
```
```
```

Sprint-2/stretch/till.js
```diff
@@ -1,31 +1,38 @@
...
+    'p') and accumulate pence, then format as pounds to two decimal places.
+*/
 
 function totalTill(till) {
   let total = 0;
 
   for (const [coin, quantity] of Object.entries(till)) {
-    total += coin * quantity;
+    // Convert a label like '50p' -> 50 (pence)
+    const pence = parseInt(String(coin).replace(/p$/i, ""), 10);
```
cjyuan: `const pence = parseInt(coin);`  would work equally well.

If you are curious why that's the case, you can look up the documentation of `parseInt()` on MDN to find out how the function works.

cjyuan (commented):
You missed updating the code and answering questions in `Sprint-2/interpret/invert.js`.

Sprint-2/stretch/till.js
Zadri415: Thanks for the suggestion. I've applied the change.

illicitonion:

PR #1118, opened 2026-03-27
London | 26-ITP-Jan | Zadri Abdule | Sprint 3 | Alarm Clock app

Zadri415:
## Learners, PR Template

Self checklist

## Changelist

Added visual helpers: triggerVisualAlarm() and clearVisualAlarm()
Call triggerVisualAlarm() when the countdown reaches zero (and when a 0-second alarm is set)
Added a small load-time listener that attaches click handlers to the Set and Stop buttons to call clearVisualAlarm() 

## Questions
N/A

### DIFF
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -1,5 +1,79 @@
-function setAlarm() {}
+let timerId = null;
+let remainingSeconds = 0;
 
+function pad(n) {
+  return String(n).padStart(2, "0");
+}
+
+function updateHeading() {
+  const heading = document.getElementById("timeRemaining");
+  if (!heading) return;
+  const m = Math.floor(remainingSeconds / 60);
+  const s = remainingSeconds % 60;
+  heading.textContent = `Time Remaining: ${pad(m)}:${pad(s)}`;
+}
+
+function clearExistingTimer() {
+  if (timerId !== null) {
+    clearInterval(timerId);
+    timerId = null;
+  }
+}
+
+function setAlarm() {
+  const input = document.getElementById("alarmSet");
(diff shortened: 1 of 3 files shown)


Sprint-3/alarmclock/alarmclock.js
cjyuan, line 14: For a "simple" function used locally in a function, we can define it as an arrow function in the function where it is used:

```javascript

function updateHeading() {
  const pad = (n) => String(n).padStart(2, "0");

  ...
}
```

This way, we can keep only relatively key functions in the outermost scope.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,5 +1,79 @@
-function setAlarm() {}
+let timerId = null;
+let remainingSeconds = 0;
```
cjyuan, line 2: Best practice is to avoid declaring variable in the global/outermost scope.
You can declare `remainingSeconds` as a local variable in `setAlarm()` and then pass it via a parameter to `updateHeading()`.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,5 +1,79 @@
...
+  }
+}
+
+function clearVisualAlarm() {
+  try {
+    document.body.classList.remove("alarm-active");
+  } catch (e) {
+    // ignore if DOM not available in test environment
+  }
+}
```
cjyuan, line 65: Nice touch in using a CSS class to manage animated background.

Could also consider combining this into one function and use a parameter to control whether to set or clear visual alarm, and set the default parameter value to "clear" or `false`.

cjyuan (commented):

illicitonion:

PR #1120, opened 2026-03-27
Manchester | 26-ITP-Jan | Abdu Hassen | Sprint 1| Data-Grouping

Abduhasen:
## Learners, PR Template

Self checklist

## Changelist

-fixing and implementing of data on Javascript using loop,if statement. 
-testing code using jest.

## Questions
N/A

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -10,7 +10,7 @@ function calculateMedian(list) {
     return null;
   }
 
-  list = list.filter((value) => typeof value === "number");
+  list = list.filter((value) => Number.isFinite(value));
 
   if (list.length === 0) {
     return null;
@@ -23,9 +23,7 @@ function calculateMedian(list) {
   if (list.length % 2 === 0) {
     return (list[middleIndex - 1] + list[middleIndex]) / 2;
   } else {
-    const middleIndex = Math.floor(list.length / 2);
-    const median = list.splice(middleIndex, 1)[0];
-    return median;
+    return list[middleIndex];
   }
 }
 
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -6,9 +6,27 @@
 // or 'list' has mixed values (the function is expected to sort only numbers).
 
 function calculateMedian(list) {
+  if (!Array.isArray(list)) {
+    return null;
+  }
+
+  list = list.filter((value) => typeof value === "number");
+
+  if (list.length === 0) {
+    return null;
+  }
+
+  list.sort((a, b) => a - b);
+
   const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+
+  if (list.length % 2 === 0) {
+    return (list[middleIndex - 1] + list[middleIndex]) / 2;
+  } else {
+    const middleIndex = Math.floor(list.length / 2);
+    const median = list.splice(middleIndex, 1)[0];
+    return median;
+  }
 }
 
 module.exports = calculateMedian;
--- a/Sprint-1/implement/dedupe.js
+++ b/Sprint-1/implement/dedupe.js
@@ -6,7 +6,7 @@ function dedupe(arr) {
     const cleanedArray = [...new Set(arr)];
     return cleanedArray;
   } else {
-    return arr;
+    return [...arr];
   }
 }
 
--- a/Sprint-1/implement/dedupe.js
+++ b/Sprint-1/implement/dedupe.js
@@ -1,1 +1,13 @@
-function dedupe() {}
+function dedupe(arr) {
+  if (arr.length === 0) {
+    return [];
+  }
+  if (arr.length !== new Set(arr).size) {
+    const cleanedArray = [...new Set(arr)];
+    return cleanedArray;
+  } else {
+    return arr;
+  }
+}
+
+module.exports = dedupe;
--- a/Sprint-1/implement/dedupe.js
+++ b/Sprint-1/implement/dedupe.js
@@ -2,9 +2,9 @@ function dedupe(arr) {
   if (arr.length === 0) {
     return [];
   }
-  if (arr.length !== new Set(arr).size) {
-    const cleanedArray = [...new Set(arr)];
-    return cleanedArray;
+  const uniqueSet = new Set(arr);
+  if (arr.length !== uniqueSet.size) {
+    return [...uniqueSet];
   } else {
     return [...arr];
   }
--- a/Sprint-1/implement/dedupe.test.js
+++ b/Sprint-1/implement/dedupe.test.js
@@ -27,13 +27,13 @@ test("The function should return a copy of the original array if the array conta
   const original1 = [2, 4, 5, 6, 8];
   expect(original1.length).toBe(new Set(original1).size);
   const result1 = dedupe(original1);
-  expect(result1).toEqual("[2, 4, 5, 6, 8]");
+  expect(result1).toEqual([2, 4, 5, 6, 8]);
   expect(result1).not.toBe(original1);
 
   const original2 = [3, 9, 25, 1, 12, 42];
   expect(original2.length).toBe(new Set(original2).size);
   const result2 = dedupe(original2);
-  expect(result2).toEqual("[3, 9, 25, 1, 12, 42]");
+  expect(result2).toEqual([3, 9, 25, 1, 12, 42]);
   expect(result2).not.toBe(original2);
 });
 // Given an array with strings or numbers
--- a/Sprint-1/implement/dedupe.test.js
+++ b/Sprint-1/implement/dedupe.test.js
@@ -27,13 +27,13 @@ test("The function should return a copy of the original array if the array conta
   const original1 = [2, 4, 5, 6, 8];
   expect(original1.length).toBe(new Set(original1).size);
   const result1 = dedupe(original1);
-  expect(result1).toEqual(original1);
+  expect(result1).toEqual("[2, 4, 5, 6, 8]");
   expect(result1).not.toBe(original1);
 
   const original2 = [3, 9, 25, 1, 12, 42];
   expect(original2.length).toBe(new Set(original2).size);
   const result2 = dedupe(original2);
-  expect(result2).toEqual(original2);
+  expect(result2).toEqual("[3, 9, 25, 1, 12, 42]");
   expect(result2).not.toBe(original2);
 });
 // Given an array with strings or numbers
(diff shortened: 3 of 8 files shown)


Sprint-1/fix/median.js
```diff
@@ -6,9 +6,27 @@
 // or 'list' has mixed values (the function is expected to sort only numbers).
 
 function calculateMedian(list) {
+  if (!Array.isArray(list)) {
+    return null;
+  }
+
+  list = list.filter((value) => typeof value === "number");
```

Sprint-1/fix/median.js
```diff
@@ -6,9 +6,27 @@
...
+  list.sort((a, b) => a - b);
+
   const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+
+  if (list.length % 2 === 0) {
+    return (list[middleIndex - 1] + list[middleIndex]) / 2;
+  } else {
+    const middleIndex = Math.floor(list.length / 2);
```
cjyuan: Why not just use the `middleIndex` declared on line 21?

Sprint-1/fix/median.js
```diff
@@ -6,9 +6,27 @@
...
+    const median = list.splice(middleIndex, 1)[0];
```
cjyuan: Why not use the same syntax you were using on line 24 to access an array element?

Sprint-1/implement/dedupe.test.js
```diff
@@ -6,23 +6,47 @@ Dedupe Array
...
+test("The function should return a copy of the original array if the array contains no duplicates", () => {
+  expect(dedupe([2, 4, 5, 6, 8])).toEqual([2, 4, 5, 6, 8]);
+  expect(dedupe([3, 7, 5, 8, 14, 19])).toEqual([3, 7, 5, 8, 14, 19]);
+  expect(dedupe(["apple", "banana", "milk", "egg"])).toEqual([
+    "apple",
+    "banana",
+    "milk",
+    "egg",
+  ]);
+});
```
cjyuan, line 38: This test should fail if the function returns the original array (instead of a **copy** of the original array).
Note: Currently your function is not returning a copy of the original array.

Sprint-1/implement/max.test.js
```diff
@@ -16,28 +16,57 @@ const findMax = require("./max.js");
...
+});
 
 // Given an array with non-number values
 // When passed to the max function
 // Then it should return the max and ignore non-numeric values
+test("The function should return the closest one to zero when an negative number array is passed", () => {
+  expect(findMax([3.5, "Leeds", "London", 3.7])).toBe(3.7);
+  expect(findMax(["italy", 0.8, 0.88, 0.89, "base"])).toBe(0.89);
+  expect(findMax([0.27, "jerry", 0.32, "apple", 0.78, "fly"])).toBe(0.78);
+});
```

Sprint-1/implement/sum.js
```diff
@@ -1,4 +1,17 @@
...
+    return undefined;
+  }
+  let sum = 0;
+  for (let i = 0; i < number.length; i++) {
+    sum += number[i];
+  }
 
+  return sum;
+}
+console.log(sum([]));
```
cjyuan: Should delete all unnecessary code to keep the code clean.

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,54 @@ const sum = require("./sum.js");
...
 // Then it should return the correct total sum
-
+test("The function should return the correct total value when decimal/float numbers array is passed", () => {
+  expect(sum([7.8045, 1.273, 3.19])).toEqual(12.2675);
+  expect(sum([4.6, 1.8, 3.18, 5.4])).toEqual(14.98);
+  expect(sum([11 / 12, 4 / 6, 1 / 2, 6 / 20])).toEqual(2.383333333333333);
+  expect(sum([1 / 2, 2 / 3, 3 / 4, 4 / 5, 5 / 6, 6 / 7])).toEqual(
+    4.4071428571428575
+  );
+});
```
cjyuan, line 49: Decimal numbers in most programming languages (including JS) are internally represented in "floating point number" format. Floating point arithmetic is not exact. For example, the result of `46.5678 - 46 === 0.5678` is false because `46.5678 - 46` only yield a value that is very close to `0.5678`. Even changing the order in which the program add/subtract numbers can yield different values.

So the following could happen
```javascript

```

Suggestion: Look up 

Sprint-1/fix/median.js
Abduhasen: to handle all type of input i changed the filter , using built in method of `Number.isFinite(value)` can handle all type of values.

Sprint-1/fix/median.js
Abduhasen: i added it mistakenly

Sprint-1/fix/median.js
Abduhasen: i was writing the codes each one by one that why i forgot to combine them and some times little bit confusing putting line of codes together if you have any recommendation of books or techniques that i can read that would be helpful. thank you.

Sprint-1/implement/dedupe.test.js
Abduhasen: i used `.not.toBe` to test if it is copy of the array or not.

Sprint-1/implement/dedupe.js
```diff
@@ -1 +1,13 @@
...
+function dedupe(arr) {
+  if (arr.length === 0) {
+    return [];
+  }
+  if (arr.length !== new Set(arr).size) {
+    const cleanedArray = [...new Set(arr)];
+    return cleanedArray;
+  } else {
+    return [...arr];
+  }
```
cjyuan: There is no need to create a Set object out of `arr` twice.

Sprint-1/implement/dedupe.test.js
```diff
@@ -6,23 +6,42 @@ Dedupe Array
...
 // Given an array with no duplicates
 // When passed to the dedupe function
 // Then it should return a copy of the original array
-
-// Given an array of strings or numbers
+test("The function should return a copy of the original array if the array contains no duplicates", () => {
+  const original = [2, 4, 5, 6, 8];
+  expect(dedupe(original)).toEqual(original);
+  expect(dedupe(original)).not.toBe(original);
+});
```
cjyuan, line 38: There is a chance that, even though the return value has incorrect elements (for example, `[2, 2, 2, 2, 2]`), 

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,55 @@ const sum = require("./sum.js");
...
 // Then it should return the correct total sum
-
+test("The function should return the correct total value when decimal/float numbers array is passed", () => {
+  expect(sum([7.8045, 1.273, 3.19])).toBeCloseTo(12.2675, 4);
+  expect(sum([4.6, 1.8, 3.18, 5.4])).toBeCloseTo(14.98);
+  expect(sum([11 / 12, 4 / 6, 1 / 2, 6 / 20])).toBeCloseTo(2.383333, 6);
+  expect(sum([1 / 2, 2 / 3, 3 / 4, 4 / 5, 5 / 6, 6 / 7])).toBeCloseTo(
+    4.407142857,
+    9
+  );
```
cjyuan, line 48: Why specify different "number of decimal places" in your tests?

Sprint-1/fix/median.js
cjyuan: Why not apply the change also to `findMax()` and `sum()`?

Sprint-1/fix/median.js
cjyuan: After the implemented function is working correctly, we should refactor our code to further improve the code. The process involves reviewing the code carefully, and reading code is a skill that takes time to develop (through reading and writing more code).

If you want something to read, you can try https://javascript.info/

Sprint-1/implement/sum.test.js
Abduhasen: Because the more decimal places I expected value has, the stricter my test needs to be.

Sprint-1/implement/dedupe.test.js
Abduhasen: i did some fixation now it is improved.

cjyuan: It is not about how many sets of data you are testing, the issue in "how" you carry out the test.

Don't think how you implement your function. Think about under what condition a buggy function that returns `[2, 2, 2, 2, 2]` can possibly pass the test.

Sprint-1/implement/sum.test.js
cjyuan: Why not just enforce the strictest tolerable difference?

If the function can pass this test  
  `expect(sum([7.8045, 1.273, 3.19])).toBeCloseTo(12.2675, 9);`
then the function will also pass  
  `expect(sum([7.8045, 1.273, 3.19])).toBeCloseTo(12.2675, 4);`

For adding only a few numbers, the rounding error will be very small. So it's ok to be strict.

Abduhasen: ok i will implement that in fact in this test i did that the test go through all the decimal numbers until the end that's why i specified each decimal number to test.

Sprint-1/implement/dedupe.test.js
```diff
@@ -25,29 +25,16 @@ test("given an empty array, it returns an empty array", () => {
 // Then it should return a copy of the original array
 test("The function should return a copy of the original array if the array contains no duplicates", () => {
   const original1 = [2, 4, 5, 6, 8];
+  expect(original1.length).toBe(new Set(original1).size);
   const result1 = dedupe(original1);
   expect(result1).toEqual(original1);
   expect(result1).not.toBe(original1);
```
cjyuan, line 31: - If their length is different, `toEqual()` would also fail. 

- Objects (including arrays) are passed to a function by reference, meaning that the they have a chance to be mutated by the function. So on line 30, it is safer to write `expect(reuslt1).toEqual('[2,4,5,6,8]')`.

cjyuan (commented):
You have all the info to further improve the code. I will mark this PR as complete first.

Sprint-1/implement/dedupe.test.js
Abduhasen: This means we should write the original value it self on this type of testing case.

cjyuan: Since you didn't notice the typo I made in the code I wrote, I am guessing you didn't run Jest test to check the change you made. Can you spot the typo?

Abduhasen: i think that the typo is putting the array in a single quotation marks the test treat the data as string.

illicitonion:

PR #1121, opened 2026-03-27
Manchester | 26-ITP-Jan | Abdu Hassen | Sprint 2| Data-Grouping

Abduhasen:
## Learners, PR Template

Self checklist

## Changelist

-debugging of codes
-implementing of codes
-interpreting of codes 

## Questions

N/A

### DIFF
--- a/Sprint-2/debug/author.js
+++ b/Sprint-2/debug/author.js
@@ -11,6 +11,10 @@ const author = {
   alive: true,
 };
 
-for (const value of author) {
-  console.log(value);
+for (const property in author) {
+  console.log(`${property}:${author[property]}`);
 }
+
+/*In this code the problem is that the method for...of is used for array 
+not for object type instead we use for...in this will work through the object
+ and list the key and for value We use author[property] */
--- a/Sprint-2/debug/author.js
+++ b/Sprint-2/debug/author.js
@@ -11,8 +11,8 @@ const author = {
   alive: true,
 };
 
-for (const property in author) {
-  console.log(`${property}:${author[property]}`);
+for (const prop in author) {
+  console.log(`${prop}:${author[prop]}`);
 }
 
 /*In this code the problem is that the method for...of is used for array 
--- a/Sprint-2/debug/recipe.js
+++ b/Sprint-2/debug/recipe.js
@@ -10,10 +10,8 @@ const recipe = {
   ingredients: ["olive oil", "tomatoes", "salt", "pepper"],
 };
 console.log(`${recipe.title} serves ${recipe.serves} 
-ingredients:`);
-for (const element of recipe.ingredients) {
-  console.log(element);
-}
+ingredients: 
+${recipe.ingredients.join("\n")}`);
 
 /* In this code will log the recipe.title and recipe.serves but recipe.ingredient is not 
 represented well so it can't log the array ,to access and to get all the array index i used  
--- a/Sprint-2/debug/recipe.js
+++ b/Sprint-2/debug/recipe.js
@@ -9,7 +9,12 @@ const recipe = {
   serves: 2,
   ingredients: ["olive oil", "tomatoes", "salt", "pepper"],
 };
+console.log(`${recipe.title} serves ${recipe.serves} 
+ingredients:`);
+for (const element of recipe.ingredients) {
+  console.log(element);
+}
 
-console.log(`${recipe.title} serves ${recipe.serves}
-  ingredients:
-${recipe}`);
+/* In this code will log the recipe.title and recipe.serves but recipe.ingredient is not 
+represented well so it can't log the array ,to access and to get all the array index i used  
+for...of loop to log all the ingredient in the array*/
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,19 +1,11 @@
 function contains(targetObject, searchTerm) {
   if (
     typeof targetObject !== "object" ||
-    Object.keys(targetObject).length === 0 ||
+    targetObject === null ||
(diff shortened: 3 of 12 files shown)


Sprint-2/debug/author.js
```diff
@@ -11,6 +11,10 @@ const author = {
   alive: true,
 };
 
-for (const value of author) {
-  console.log(value);
+for (const property in author) {
```
cjyuan: Can also name the variable `key` -- shorter name.

Sprint-2/debug/recipe.js
```diff
@@ -9,7 +9,12 @@ const recipe = {
   serves: 2,
   ingredients: ["olive oil", "tomatoes", "salt", "pepper"],
 };
+console.log(`${recipe.title} serves ${recipe.serves} 
+ingredients:`);
+for (const element of recipe.ingredients) {
+  console.log(element);
+}
```

Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,19 @@
-function contains() {}
+function contains(targetObject, searchTerm) {
+  if (
+    typeof targetObject !== "object" ||
+    Object.keys(targetObject).length === 0 ||
+    Array.isArray(targetObject)
+  ) {
+    return false;
+  }
```
cjyuan, line 8: What would happen when `targetObject` is `null`?

The check on line 4 is optional. If `searchTerm` is not a key of the object, the remaining code would also return `false`.

Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,19 @@
...
+
+  for (const property in targetObject) {
+    if (targetObject[property] === searchTerm || property === searchTerm) {
+      return true;
+    }
+  }
```
cjyuan: - The function is supposed to check only if it contains `searchTerm` as its key. There is no need to compare its value.

- Can also explore `Object.hasOwn()`.

Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,19 @@
...
+  return false;
+}
+console.log(contains(["a", "b", "c"], "d"));
```
cjyuan: Should delete all unnecessary code to keep the code clean.

Sprint-2/implement/contains.test.js
```diff
@@ -20,16 +20,37 @@ as the object doesn't contains a key of 'c'
...
+  );
+});
 
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+test("the function should return false when invalid parameter like an array is passed", () => {
+  expect(contains([10, 23, 34, 45], 45)).toBe(false);
+  expect(contains("hello", "hi")).toBe(false);
+  expect(contains(["green", "yellow", "rad"], "green")).toBe(false);
```
This is because `contains([10, 23, 34, 45], 45)` could also return `false` simply because `45` is not a key of the array.

Arrays are objects, with their **indices acting as keys.** A proper test should use a valid key to ensure the function returns `false` specifically because the input is an array, not because the key is missing.

Sprint-2/implement/querystring.js
```diff
@@ -6,11 +6,16 @@ function parseQueryString(queryString) {
...
+    if (equalityPosition == -1) {
+      continue;
+    }
+    const key = pair.slice(0, equalityPosition);
+    const value = pair.slice(equalityPosition + 1);
     queryParams[key] = value;
   }
 
   return queryParams;
 }
```
cjyuan, line 21: These arguments are also valid query strings:

```javascript
parseQueryString("key=")
parseQueryString("=value")
```

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,18 @@
-function tally() {}
+function tally(myArray) {
+  if (myArray.length === 0) {
+    return {};
+  } else if (typeof myArray === "string") {
+    throw Error;
+  }
```
cjyuan: - Why only throw an error when `myArray` is a string? Why not do the same for other types of data?

- Why not throw an error when `myArray` is an empty string?

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,18 @@
...
+  let tallyObject = {};
```
```
```

Sprint-2/implement/contains.test.js
```diff
@@ -20,16 +20,37 @@ as the object doesn't contains a key of 'c'
...
+    false
+  );
+});
 
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+test("the function should return false when invalid parameter like an array is passed", () => {
+  expect(contains([10, 23, 34, 45], "3")).toBe(false);
+  expect(contains("hello", "hi")).toBe(false);
```
cjyuan: Note: String also have their indices acting as keys. 

`Object.hasOwn("hello", "0")` evaluates to true.

cjyuan (commented):
Changes look good.

Please note that in CYF courses, the recommended way to inform the reviewer of your changes is to do both of the following:
  - **Reply to their feedback.**
    - In the responses, clarify how each piece of feedback was addressed to demonstrate that you've carefully reviewed the suggestions.
      - You may find the suggestions in this [PR Guide](https://curriculum.codeyourfuture.io/guides/reviewing/trainee-pr-guide/#addressing-reviewer-feedback) useful.
    - Your response may trigger a notification (depending on the reviewer's settings), helping ensure they’re aware of the updates you’ve made. 
  - **Replace the "Reviewed" label by a "Needs review" label** (which you have done -- great!)
    - Without this label, the reviewer would not know if your changes is ready to be reviewed.

illicitonion:

PR #1124, opened 2026-03-27
Manchester | 26-ITP-Jan | Abdu Hassen | Sprint 3| Alarm Clock App

Abduhasen:
## Learners, PR Template

Self checklist

## Changelist

-write and implement alarm clock 

## Questions

N/A

### DIFF
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -1,8 +1,13 @@
+let myInterval;
 function setAlarm() {
+  clearInterval(myInterval);
   const userInput = document.querySelector("#alarmSet").value;
-  let timeRemaining = userInput;
-
-  const myInterval = setInterval(function () {
+  let timeRemaining = Number(userInput);
+  if (isNaN(timeRemaining) || timeRemaining <= 0) {
+    alert("Please enter a valid number.");
+    return;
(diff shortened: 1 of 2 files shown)


Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,22 @@
-function setAlarm() {}
+function setAlarm() {
```
Luro91, line 2: What happens when the user sets an alarm when the countdown is already running?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,22 @@
...
+  const userInput = document.querySelector("#alarmSet").value;
+  let timeRemaining = userInput;
```
Luro91: What happens when the user enters a negative number or a string value?

Luro91 (commented):
Well done. 

A note on commit messages. It's hard to know for me what happens in a commit if the message is not descriptive. A commit should describe what changed in the code. It's also good practice to split commits into logical blocks (for example for each task or each review comment). This makes it easier to understand and additionally the commits could also be undone more easily if needed.
I like this article about commit messages: https://chris.beams.io/git-commit

illicitonion:

PR #1126, opened 2026-03-27
London | 26-ITP-Jan | Miriam Jorna | Sprint 1 | Data Groups

miriamjorna:
Self checklist

## Changelist

Have done the exercises for the Sprint-1 homework.
Thank you for your time reviewing!

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -1,3 +1,4 @@
+
 // Fix this implementation
 // Start by running the tests for this function
 // If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
@@ -6,9 +7,25 @@
 // or 'list' has mixed values (the function is expected to sort only numbers).
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  // filter only numbers from the list
+  const numbers = list.filter(item => typeof item === "number");
+  
+  // return null if no numbers found
+  if (numbers.length === 0) {
+    return null;
+  }
+  
+  // sort the numbers
+  numbers.sort((a, b) => a - b);
+  
+  const middleIndex = Math.floor(numbers.length / 2);
+  
+  // if even number of elements, average the two middle values
+  if (numbers.length % 2 === 0) {
+    return (numbers[middleIndex - 1] + numbers[middleIndex]) / 2;
+  }
+  
+  return numbers[middleIndex];
 }
 
 module.exports = calculateMedian;
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -1,14 +1,20 @@
 
 // Fix this implementation
 // Start by running the tests for this function
-// If you're in the Sprint-1 directory, you can run `npm test -- fix` to run the tests in the fix directory
+// If you're in the Sprint-1 directory, you can run `npm test -- fix`
+// to run the tests in the fix directory
 
 // Hint: Please consider scenarios when 'list' doesn't have numbers (the function is expected to return null)
 // or 'list' has mixed values (the function is expected to sort only numbers).
 
 function calculateMedian(list) {
+  // return null if input is not an array
+  if (!Array.isArray(list)) {
+    return null;
+  }
+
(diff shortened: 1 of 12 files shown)


Sprint-1/fix/median.js
```diff
@@ -6,9 +7,25 @@
 // or 'list' has mixed values (the function is expected to sort only numbers).
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  // filter only numbers from the list
+  const numbers = list.filter(item => typeof item === "number");
```

Sprint-1/implement/dedupe.test.js
```diff
@@ -16,7 +16,22 @@ E.g. dedupe([1, 2, 1]) returns [1, 2]
...
 // Then it should return an empty array
-test.todo("given an empty array, it returns an empty array");
+
+test("given an empty array, it returns an empty array", () => {
+  expect(dedupe([])).toEqual([]);
+});
+
+test("given an array with no duplicates, it returns a copy of the original array", () => {
+  expect(dedupe([1, 2, 3])).toEqual([1, 2, 3]);
+});
```
cjyuan, line 28: Your function implementation is correct. However, this test could be improved to better ensure
that any future changes continue to align with the expected behavior:

Sprint-1/implement/max.test.js
```diff
@@ -12,11 +12,33 @@ We have set things up already so that this file can see your function from the o
...
+  expect(findMax([-5, -1, -10])).toEqual(-1);
+});
+
+test("given an array with decimal numbers, returns the largest decimal", () => {
+  expect(findMax([1.5, 2.7, 0.3])).toEqual(2.7);
+});
+
+test("given an array with non-number values, returns the max and ignores non-numeric values", () => {
+  expect(findMax(['hey', 10, 'hi', 60, 10])).toEqual(60);
+});
```
cjyuan, line 37: When a string representing a valid numeric literal (for example, `"300"`) is compared to a number, 

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,28 @@ const sum = require("./sum.js");
...
+test("given an array containing negative numbers, returns the correct total sum", () => {
+  expect(sum([10, -5, 20])).toEqual(25);
+});
 
-// Given an array containing non-number values
-// When passed to the sum function
-// Then it should ignore the non-numerical values and return the sum of the numerical elements
+test("given an array with decimal numbers, returns the correct total sum", () => {
+  expect(sum([1.5, 2.5, 3.0])).toEqual(7);
+});
```
cjyuan, line 30: Decimal numbers in most programming languages (including JS) are internally represented in "floating point number" format. Floating point arithmetic is not exact. For example, the result of `46.5678 - 46 === 0.5678` is false because `46.5678 - 46` only yield a value that is very close to `0.5678`. Even changing the order in which the program add/subtract numbers can yield different values.

So the following could happen
```javascript

```

Suggestion: Look up 

cjyuan (commented):
You missed updating `Sprint-1/refactor/includes.js`.

miriamjorna:
Hm hm hmmm I had a nagging feeling I did not grasp the subject entirely and this does prove I need to put more time into rehashing it. Many thanks. It will take a bit of time to resubmit.

Sprint-1/fix/median.js
miriamjorna: Thank you - sorted in the update

Sprint-1/implement/dedupe.test.js
miriamjorna: Previously I understood the difference but not the method. Have now found the existential answer in .not.toBe

Sprint-1/implement/sum.test.js
miriamjorna: Insightful! Have changed.

Sprint-1/implement/max.test.js
miriamjorna: Thank you, have added.

miriamjorna:
Apologies for taking so long. Finally ready for review, including the initially forgotten refactoring exercise.

cjyuan:
Changes look good. Well done.

illicitonion:

PR #1129, opened 2026-03-28
London | 26-ITP-Jan | Angela McLeary | Sprint 1 | Sprint 1

AngelaMcLeary:
## Learners, PR Template

Self checklist

## Changelist

This PR is about fixing median.js, Implementing dedupe.js, max.js, sum.js and refactoring include.js.


Sprint-1/fix/median.js
```diff
@@ -6,9 +6,35 @@
...
 
 function calculateMedian(list) {
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
-}
-
+  if (!Array.isArray(list) || list.length === 0) return null;
+  //If the items are not an array or if the array is empty return null
+  const numbers = list.filter((num) => typeof num === "number");
```

Sprint-1/fix/median.js
```diff
@@ -6,9 +6,35 @@
...
+  if (numbers.length === 0) return null;
+  //checks the array before sorting and removing anything that is not a valid item ie not a number.
+  const sortedList = [...numbers].sort((a, b) => a - b);
```
cjyuan, line 29: Is it necessary to create a copy of `numbers` first before sorting in this function?

Sprint-1/implement/dedupe.test.js
```diff
@@ -16,13 +16,27 @@ E.g. dedupe([1, 2, 1]) returns [1, 2]
...
+  expect(dedupe([])).toEqual([]);
+});
 
 // Given an array with no duplicates
 // When passed to the dedupe function
 // Then it should return a copy of the original array
+test("given an array has no duplicates, it returns a copy of the original array", () => {
+  expect(dedupe([1, 2, 3])).toEqual([1, 2, 3]);
+  expect(dedupe([5, 1, 4])).toEqual([5, 1, 4]);
+});
```
cjyuan, line 31: This test should fail if the function returns the original array (instead of a **copy** of the original array).

Sprint-1/implement/max.test.js
```diff
@@ -16,28 +16,51 @@ const findMax = require("./max.js");
...
+});
+
 // Given an array with non-number values
 // When passed to the max function
 // Then it should return the max and ignore non-numeric values
 
+test("ignores non-number values and returns the max number", () => {
+  expect(findMax(["Not", "A", "Number", 75, 85, 105])).toEqual(105);
+});
+
```
cjyuan, line 60: When a string representing a valid numeric literal (for example, `"300"`) is compared to a number, 

Sprint-1/implement/sum.test.js
```diff
@@ -13,24 +13,40 @@ const sum = require("./sum.js");
...
+  expect(sum([5, 2, -3])).toEqual(4);
+});
 
 // Given an array with decimal/float numbers
 // When passed to the sum function
 // Then it should return the correct total sum
-
+test("given an array with decimal float numbers, it should return the total sum", () => {
+  expect(sum([1.5, 2.5, 3.5])).toEqual(7.5);
+});
```

So the following could happen
```javascript

```

Suggestion: Look up 

Sprint-1/fix/median.js
AngelaMcLeary: Hi @cjyuan, It is not strictly necessary, but it is good practice to do this to avoid modifying the original array and keeping it in tact.

Sprint-1/fix/median.js
AngelaMcLeary: Hi @cjyuan, Thank you for your feedback. I intend to. I have updated the code to accordingly. It should now pass the test correctly.

Sprint-1/implement/dedupe.test.js
AngelaMcLeary: Hi @cjyuan, Thank you for the feedback. I added an additional assertion to ensure the dedupe function returns a new array instance, not the original array. The previous test only checked for value equality using .toEqual(), which would still pass even if the function simply returned the input array.To properly validate the acceptance criteria, the test is now fortified.

Sprint-1/implement/max.test.js
AngelaMcLeary: Hi @cjyuan, Thank you for your feedback. I have updated the test to include an edge case for numerical string.

Sprint-1/implement/sum.test.js
AngelaMcLeary: Hi @cjyuan, Thank you for your feedback. I understand now why decimal values need special handling. toBeCloseTo() is a better choice because it checks that the result is close enough within a small margin, which avoids issues caused by floating‑point rounding.

Sprint-1/implement/dedupe.test.js
```diff
@@ -16,13 +16,29 @@ E.g. dedupe([1, 2, 1]) returns [1, 2]
...
 
 // Given an array with no duplicates
 // When passed to the dedupe function
 // Then it should return a copy of the original array
+test("given an array has no duplicates, it returns a copy of the original array", () => {
+ const originalArray = [1, 2, 3];
+ const result = dedupe(originalArray);
+  expect(result).toEqual(originalArray);
+  expect(result).not.toBe(originalArray);
```
cjyuan, line 30: This test can verify the function can return a different array.

However, there is a chance that, even though `result` has incorrect elements (for example, `[1, 1, 1]`), 
the test on line 29 could still pass. Can you figure out why, and then improve the test accordingly?

cjyuan (commented):
Other changes are good. Just one issue left.

Sprint-1/implement/sum.test.js
cjyuan: Note: The default value of the second parameter of `toBeCloseTo()` is only 2   
See: https://jestjs.io/docs/expect#tobeclosetonumber-numdigits

For the default value 2, the test criterion is `Math.abs(expected - received) < 0.005`, which may not be precise enough in some applications.

Sprint-1/implement/dedupe.test.js
AngelaMcLeary: Hi @cjyuan, Thanks for your feedback. I have done some research and understand that this only checks value equality, so it could still pass even if the function returns the same array or mutates it. Added checks to make sure a new array is returned and the original input isn't changed. I have added additional test.

cjyuan:
Changes look good. Well done.

illicitonion:

PR #1131, opened 2026-03-28
Sheffield | 26-ITP-Jan | Mahammad Osman | Sprint 2 | Data Groups

Darkidd77:
## Learners, PR Template

Self checklist

## Changelist:
Completed the tasks per instructions.

### DIFF
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,5 +1,5 @@
 function contains(obj, prop) {
-  if (Array.isArray(obj) || obj === undefined || obj === null) {
+  if (typeof obj !== "object" || obj === null || Array.isArray(obj)) {
     return false;
   }
   return obj.hasOwnProperty(prop);
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,4 +1,7 @@
 function contains(obj, prop) {
+  if (Array.isArray(obj) || obj === undefined || obj === null) {
+    return false;
+  }
   return obj.hasOwnProperty(prop);
 }
 
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,3 +1,5 @@
-function contains() {}
+function contains(obj, prop) {
+  return obj.hasOwnProperty(prop);
+}
 
 module.exports = contains;
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -51,4 +51,14 @@ describe("contains", () => {
     const arr = [1, 2, 3];
     expect(contains(arr, "a")).toEqual(false);
   });
+
+  test("contains returns false for arrays, even if the index exists", () => {
+    const arr = ["a", "b", "c"];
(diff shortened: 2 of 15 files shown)


Sprint-2/implement/contains.test.js
```diff
@@ -11,25 +11,44 @@ E.g. contains({a: 1, b: 2}, 'c') // returns false
...
+    expect(contains(obj, "c")).toEqual(false);
+  });
+  // Given invalid parameters like an array
+  // When passed to contains
+  // Then it should return false or throw an error
+  test("contains returns false for invalid parameters", () => {
+    const arr = [1, 2, 3];
+    expect(contains(arr, "a")).toEqual(false);
+  });
+});
```
cjyuan, line 71: This test cannot not yet confirm that the function correctly returns false when the first argument is an array. 

In addition, does your function behave as expected when the first argument is `null` or `undefined`?

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,19 @@
...
+
+  const counts = {};
+
+  for (const item of list) {
+    if (counts[item]) {
+      counts[item] = counts[item] + 1;
+    } else {
+      counts[item] = 1;
+    }
+  }
```
```
```

Sprint-2/implement/contains.test.js
Darkidd77: Yes, I now understand more what my function was exactly doing, I missed the fact that arrays are objects. as for the `null` and `undefined` it will def throw an error. I did try adding the tests cases you mentioned and it did brill fail. 
To fix I added more conditions to the function to make sure `null` and `undefined` and arrays as well to return `false`
Thank you

Sprint-2/implement/tally.js
Darkidd77: I looked it up, and now I realise that the `const counts = {};` is not actually an empty object completely it does inherit the properties and methods of Object prototype. Indeed the function call example you mentioned does not return the expected results.
to fix I used `Object.create(null)` to create an empty object with no inherited properties and added the test case you mentioned. Thank you

Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,8 @@
-function contains() {}
+function contains(obj, prop) {
+  if (Array.isArray(obj) || obj === undefined || obj === null) {
+    return false;
+  }
+  return obj.hasOwnProperty(prop);
+}
```
cjyuan, line 6: Can you check if `contains("abc", "0")` returns what you expect?

Darkidd77: hmmm indeed it doesn't? So after Looking into it I realised that JS considers not only arrays to be objects, but strings are actually arrays in disguise too, so  the string `"abc"` behind the scene is actually: 
`{
  "0": "a",
  "1": "b",
  "2": "c",
  "length": 3
}`

meaning `"abc".hasOwnProperty("0")` evaluates to `true`

I've updated the guard clauses to include `typeof obj !== 'object'` to immediately filter out strings, numbers, and booleans. Thanks

cjyuan:
All good now. Well done.

illicitonion:

PR #1132, opened 2026-03-28
London | 26-ITP-Jan | Angela McLeary | Sprint 2 | Sprint 2

AngelaMcLeary:
## Learners, PR Template

Self checklist

## Changelist

This PR is about debugging, implementing and interpreting code.

### DIFF
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -51,7 +51,7 @@ test("returns false when object does not contain the property", () => {
 // When passed to contains
 // Then it should return false or throw an error
 test("returns false for invalid inputs", () => {
-  expect(contains([], "a")).toBe(false);
+  expect(contains(["a"], 0)).toBe(false); //no arrays as objects.
   expect(contains(2, "a")).toBe(false);
   expect(contains("notANumber", "a")).toBe(false);
   expect(contains(null, "a")).toBe(false);
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -53,6 +53,6 @@ test("returns false when object does not contain the property", () => {
 test("returns false for invalid inputs", () => {
   expect(contains([], "a")).toBe(false);
   expect(contains(2, "a")).toBe(false);
-  expect(contains("notAnumber", "a")).toBe(false);
+  expect(contains("notANumber", "a")).toBe(false);
   expect(contains(null, "a")).toBe(false);
 });
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -16,43 +16,43 @@ as the object doesn't contains a key of 'c'
 // Given a contains function
 // When passed an object and a property name
 // Then it should return true if the object contains the property, false otherwise
-  test("When object contains the property, it returns true", () => {
-    const obj = { a: 1, b: 2 };
-    expect(contains(obj, 'a')).toBe(true);
-    expect(contains(obj, 'b')).toBe(true);
-  });
+test("When object contains the property, it returns true", () => {
+  const obj = { a: 1, b: 2 };
+  expect(contains(obj, "a")).toBe(true);
+  expect(contains(obj, "b")).toBe(true);
+});
 
 // Given an empty object
 // When passed to contains
 // Then it should return false
 test("When given an empty object, it returns false", () => {
-    expect(contains({}, 'a')).toBe(false);
-  });
+  expect(contains({}, "a")).toBe(false);
+});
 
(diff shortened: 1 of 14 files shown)


Sprint-2/implement/contains.test.js
```diff
@@ -16,20 +16,43 @@ as the object doesn't contains a key of 'c'
...
+test("returns false when object does not contain the property", () => {
+  const obj = { a: 1, b: 2 };
+  expect(contains(obj, "c")).toBe(false);
+});
+
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+test("returns false for invalid inputs", () => {
+  expect(contains([], "a")).toBe(false);
```

Sprint-2/implement/querystring.js
```diff
@@ -1,16 +1,28 @@
...
+  // loop through each part
 
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
+    // Find the first "=" sign
+    const index = pair.indexOf("=");
+    // throw error if "=" is missing
+  if (index === -1) {
+    throw new Error("Invalid query string: missing '='");
+  }// Everything before "=" is the key
```
cjyuan, line 18: FYI: These arguments are considered valid query strings:
```
parseQueryString("key=")
parseQueryString("=value")
```

No change required.

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,24 @@
-function tally() {}
+function tally(list) {
+  //This object will store each item as a key and how many times it appears as a value. First declare an empty array
+  var counts = {};
```
```
```

Sprint-2/interpret/invert.js
```diff
@@ -10,20 +10,34 @@ function invert(obj) {
   const invertedObj = {};
 
   for (const [key, value] of Object.entries(obj)) {
-    invertedObj.key = value;
+    //create new keys  and values to make them dynamic
+    let newKey = value.toString();
+    let newValue = key.toString();
+    invertedObj[newKey] = newValue;
   }
```
cjyuan, line 17: What do you expect from the following function calls?

`invert({ a: null })`
`invert({ a: undefined })`

Sprint-2/implement/contains.test.js
AngelaMcLeary: Hi @cjyuan, Thank you for your feedback. I have corrected the test and updated the implementation the explicitly check that the input is a plain object, not an array.

Sprint-2/implement/querystring.js
AngelaMcLeary: Hi @cjyuan, Thank you for the feedback and clarification. I understand now. I’ll keep in mind that more flexible parsers would need to handle empty keys, empty values, and consecutive separators etc.

Sprint-2/implement/tally.js
AngelaMcLeary: Hi @cjyuan, Thanks for the feedback. It does not return the value I expected, the output is { toString: 'function toString() { [native code] }11' } but I expected toString:2. 

I see the issue now — using {} means the object inherits built‑in keys like toString, so tallying the string 'toString' gives incorrect results. Using Object.create(null) creates an object with no inherited properties, so every string (including 'toString') is counted correctly.”

I have updated the implementation.

Sprint-2/interpret/invert.js
AngelaMcLeary: Hi @cjyuan, Thanks for the feedback. I thought that it would return { null: 'a' } and { undefined: 'a' }.
However it returned nothing for both.

I see the problem clearly. When the function calls toString it is crashing. I've updated the code to use String(value) and String(key) instead, which safely converts all values (including null and undefined) without throwing an error. The function now works correctly for all the test cases.”

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,25 @@
-function tally() {}
+function tally(list) {
+  //This object will store each item as a key and how many times it appears as a value. First declare an empty array
+  var counts = Object.create(null);
```
cjyuan, line 3: Should use `let` instead of `var`. `var`, when not used properly, can introduce unintended side effect.

cjyuan (commented):
Changes look good. Well done.

Sprint-2/interpret/invert.js
cjyuan: Note: 
- When a value (except symbol) is used as key, it is **implicitly converted** to its equivalent string value. 
- Key is always a string (except when it is a symbol).

So you don't actually have to explicitly convert the old key or value to strings.
Just `invertedObj[value] = key;` will do.

illicitonion:

PR #1133, opened 2026-03-28
Sheffield | 26-Jan-ITP | Martha Ogunbiyi| Sprint 2| Coursework/sprint 2

marthak1:
## Learners, PR Template

Self checklist

## Changelist

Sprint 2 coursework done

### DIFF
--- a/Sprint-2/debug/author.js
+++ b/Sprint-2/debug/author.js
@@ -12,5 +12,5 @@ const author = {
 };
 
 for (const value in author) {
-  console.log(value);
+  console.log(author[value]);
 }
--- a/Sprint-2/debug/author.js
+++ b/Sprint-2/debug/author.js
@@ -1,5 +1,5 @@
 // Predict and explain first...
-
+// for...of is trying to iterate a non iterable object
 // This program attempts to log out all the property values in the object.
 // But it isn't working. Explain why first and then fix the problem
 
@@ -11,6 +11,6 @@ const author = {
   alive: true,
 };
 
-for (const value of author) {
+for (const value in author) {
   console.log(value);
 }
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,3 +1,8 @@
-function contains() {}
+function contains(obj, targetKey) {
+  if (obj.constructor !== Object) {
+    throw new Error("Invalid Parameter");
+  }
+  return Object.keys(obj).includes(targetKey);
+}
 
 module.exports = contains;
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,5 +1,10 @@
 function contains(obj, targetKey) {
-  if (obj.constructor !== Object) {
+  if (
+    obj === null ||
+    typeof obj !== "object" ||
+    Array.isArray(obj) ||
+    Object.getPrototypeOf(obj) !== Object.prototype
+  ) {
     throw new Error("Invalid Parameter");
   }
   return Object.hasOwn(obj, targetKey);
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -2,7 +2,7 @@ function contains(obj, targetKey) {
   if (obj.constructor !== Object) {
     throw new Error("Invalid Parameter");
   }
-  return Object.keys(obj).includes(targetKey);
+  return Object.hasOwn(obj, targetKey);
 }
 
 module.exports = contains;
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -2,12 +2,10 @@ function contains(obj, targetKey) {
   if (
     obj === null ||
     typeof obj !== "object" ||
-    Array.isArray(obj) ||
-    Object.getPrototypeOf(obj) !== Object.prototype
+    Array.isArray(obj)
   ) {
     throw new Error("Invalid Parameter");
   }
   return Object.hasOwn(obj, targetKey);
 }
-
 module.exports = contains;
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -20,16 +20,43 @@ as the object doesn't contains a key of 'c'
 // Given an empty object
 // When passed to contains
 // Then it should return false
-test.todo("contains on empty object returns false");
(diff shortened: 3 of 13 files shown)


Sprint-2/debug/author.js
```diff
@@ -11,6 +11,6 @@ const author = {
   alive: true,
 };
 
-for (const value of author) {
+for (const value in author) {
   console.log(value);
 }
```
cjyuan, line 16: Does this code output the property values such as "Zadie", "Smith", ...?

Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,8 @@
-function contains() {}
+function contains(obj, targetKey) {
+  if (obj.constructor !== Object) {
+    throw new Error("Invalid Parameter");
+  }
+  return Object.keys(obj).includes(targetKey);
```
cjyuan: Can also use `Object.hasOwn()` -- probably a more efficient approach.

Sprint-2/implement/contains.test.js
```diff
@@ -20,16 +20,43 @@ as the object doesn't contains a key of 'c'
...
 
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+test("given an invalid parameters, it should return false or throw an error", () => {
+    const input = ['a', 'b', 'c'];
+    const target = 'a'
+    expect(() => contains(input, target)).toThrow();
+    
+});
```
cjyuan, line 61: This test does not yet confirm that the function correctly returns false when the first argument is an array. 
This is because `contains(['a', 'b', 'c'], "a")` could also return `false` simply because "a" is not a key of the array.

Arrays are objects, with their indices acting as keys. A proper test should use an array along with a valid key to ensure the function returns `false` specifically because the input is an array, not because the key is missing.

Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,8 @@
-function contains() {}
+function contains(obj, targetKey) {
+  if (obj.constructor !== Object) {
+    throw new Error("Invalid Parameter");
+  }
```
cjyuan: For each of the following function calls, does your function return the value you expect?

```
const obj = new Map();  // Map object is an object 
obj.foo = 1;
contains(obj, "foo"); 

// According to the spec in the test file, the function should also reject 
// array (even though array is a kind of objects in JS)
contains([0, 0, 0], "1"); 

contains(null, "1");
contains(undefined, "1");
```

Sprint-2/implement/querystring.js
```diff
@@ -6,7 +6,14 @@ function parseQueryString(queryString) {
...
+    const index = pair.indexOf("=");
+
+    if (index === -1) {
+      queryParams[pair] = "";
+      continue;
+    }
+    const key = pair.slice(0, index);
+    const value = pair.slice(index + 1);
     queryParams[key] = value;
   }
```
cjyuan, line 19: For the following function call, does your function return the value you expect?

```
```

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,14 @@
-function tally() {}
+function tally(list) {
+    const uniqueItems = {};
+    if(list.constructor === String){
```
cjyuan: Why only treat string as invalid argument? What about other data types: `null`, `undefined`, `123`, `true`, etc.

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,14 @@
...
+        throw new Error ("Invalid Input")
+    }
+    for (const frequency of list){
+    uniqueItems[frequency] = (uniqueItems[frequency] || 0) + 1;
+    }
```

as recommended in https://github.com/CodeYourFuture/Module-Structuring-and-Testing-Data/blob/main/readme.md ?

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,14 @@
-function tally() {}
+function tally(list) {
+    const uniqueItems = {};
```
```
```

Sprint-2/debug/author.js
marthak1: Thanks for the feedback, i have update to output the values rather the keys

Sprint-2/implement/contains.js
marthak1: Thanks for the feedback. I have now updated the function implementation.

Sprint-2/implement/contains.test.js
marthak1: Thanks for the feedback. Test case now updated to correctly fail, given an invalid parameter

Sprint-2/implement/contains.js
marthak1: Thanks for the feedback. Validation checks has now been updated

Sprint-2/implement/querystring.js
marthak1: Thanks for the feedback, function updated to skip empty entries

Sprint-2/implement/tally.js
marthak1: Thanks for the feedback. all update now implemented .

Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,13 @@
-function contains() {}
+function contains(obj, targetKey) {
+  if (
+    obj === null ||
+    typeof obj !== "object" ||
+    Array.isArray(obj) ||
+    Object.getPrototypeOf(obj) !== Object.prototype
```
cjyuan: What is the purpose of the condition on line 6?

cjyuan (commented):
Changes look good. Just one more question.

Sprint-2/implement/contains.js
marthak1: Thanks for feedback. Line 6 ensures the function also reject non plain objects like Map, Set , Date etc.

cjyuan: What about an empty object created from `Object.create(null)`? Do you plan to reject that kind of object as well?

Note: Prototype is a relatively advanced concept in JS programming. If you have to use it, use it with care.

cjyuan:
All good.

Sprint-2/implement/contains.js
marthak1: Thanks for the feedback,  I never thought about about objects created with no prototypes at all on this case. function has now been updated.

illicitonion:

PR #1140, opened 2026-03-28
London | 26-ITP-Jan | Carlos Abreu |Sprint 3 | Alarm Clock

carlosyabreu:
## Learners, PR Template

Self checklist

## Changelist
Alarm clock app PR for Sprint 3 module

### DIFF
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -1,4 +1,52 @@
-function setAlarm() {}
+let timerInterval = null;
+let currentTime = 0;
+
+function setAlarm() {
+  // Clear any existing timer
+  if (timerInterval) {
+    clearInterval(timerInterval);
+    timerInterval = null;
+  }
+  
+  // Get the input value
+  const input = document.getElementById("alarmSet");
+  let totalSeconds = parseInt(input.value);
+  
+  // Validate input (if empty or invalid, default to 0)
+  if (isNaN(totalSeconds)) {
+    totalSeconds = 0;
+  }
+  
+  currentTime = totalSeconds;
+  
+  // Update the display
+  updateTimeDisplay(currentTime);
+  
+  // Start the countdown
+  timerInterval = setInterval(() => {
+    if (currentTime > 0) {
+      currentTime--;
+      updateTimeDisplay(currentTime);
+      
+      // Check if timer reached zero
+      if (currentTime === 0) {
+        clearInterval(timerInterval);
+        timerInterval = null;
+        playAlarm();
+        // Change background color
+        document.body.style.backgroundColor = "red";
+      }
+    }
+  }, 1000);
+}
+
+function updateTimeDisplay(totalSeconds) {
+  const minutes = Math.floor(totalSeconds / 60);
+  const seconds = totalSeconds % 60;
+  const formattedTime = `${minutes.toString().padStart(2, "0")}:${seconds.toString().padStart(2, "0")}`;
+  const heading = document.getElementById("timeRemaining");
+  heading.textContent = `Time Remaining: ${formattedTime}`;
+}
 
 // DO NOT EDIT BELOW HERE
 
@@ -16,10 +64,25 @@ function setup() {
 
 function playAlarm() {
   audio.play();
+  // Play continuously by looping
+  audio.loop = true;
 }
 
 function pauseAlarm() {
   audio.pause();
+  audio.currentTime = 0;
+  // Reset background color when alarm is stopped
+  document.body.style.backgroundColor = "";
+  
+  // Clear any ongoing timer
+  if (timerInterval) {
+    clearInterval(timerInterval);
+    timerInterval = null;
+  }
+  
+  // Reset display to 00:00
+  currentTime = 0;
+  updateTimeDisplay(0);
 }
 
 window.onload = setup;
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,4 @@ node_modules
 **/.DS_Store
 .idea
 package-lock.json
-package.json
+**package.json
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
-node_modules
+node_modules/
 .DS_Store
-.vscode
-**/.DS_Store
-.idea
+.vscode/
+.idea/
 package-lock.json
+package.json
 **/package.json
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ node_modules
 **/.DS_Store
 .idea
 package-lock.json
+package.json
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,4 @@ node_modules
 **/.DS_Store
 .idea
 package-lock.json
-**package.json
+**/package.json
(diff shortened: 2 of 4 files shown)


Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,52 @@
-function setAlarm() {}
+let timerInterval = null;
+let currentTime = 0;
```
cjyuan, line 2: Why not declare `currentTime` as a local variable inside `setAlarm()`?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,52 @@
...
+
+function setAlarm() {
+  // Clear any existing timer
+  if (timerInterval) {
+    clearInterval(timerInterval);
+    timerInterval = null;
+  }
+
```
cjyuan, line 10: Currently when starting a new countdown, the application does not always return to a clean initial state, 

Can also consider introducing a dedicated reset function to return the app to a clean initial state to help ensure consistency. There are few places in this script you can call this reset function instead of repeating the reset logic.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,52 @@
...
+  }
+  
+  // Get the input value
+  const input = document.getElementById("alarmSet");
+  let totalSeconds = parseInt(input.value);
+  
+  // Validate input (if empty or invalid, default to 0)
+  if (isNaN(totalSeconds)) {
+    totalSeconds = 0;
+  }
```
cjyuan, line 18: What range of integers are acceptable? (Currently some input can make the app behave abnormally)

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,52 @@
...
+      if (currentTime === 0) {
+        clearInterval(timerInterval);
+        timerInterval = null;
+        playAlarm();
+        // Change background color
+        document.body.style.backgroundColor = "red";
+      }
+    }
+  }, 1000);
+}
```
cjyuan, line 41: If input is 0, the timer interval will remain active even though we won't notice anything visually.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -16,10 +64,25 @@ function setup() {
...
   audio.play();
+  // Play continuously by looping
+  audio.loop = true;
 }
 
 function pauseAlarm() {
   audio.pause();
+  audio.currentTime = 0;
+  // Reset background color when alarm is stopped
+  document.body.style.backgroundColor = "";
```
cjyuan, line 75: To better separate presentation logic from application logic, you can consider defining a CSS class, and use [`classList.toggle()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/classList) to apply/remove the style. For example,
```
```

Sprint-3/alarmclock/alarmclock.js
carlosyabreu: Sorry @cjyuan for the delay.
I've been busy and stressed working on last and final module (Moduel Data Flows) and the Project-TV-Show with a colleague and working in a collaborative way with Git and GitHub has been a hell to understand its workflow. 
It's been challenging but rewarding to finally get a glimpse how  it works in a shared environment.
As I'm almost done with Module Data Flows and Project TV Show (I'm at level 500 just to finish today later) I'm back to Module Data Group.

Going back to your query. 
The reason currentTime is declared outside setAlarm() (at the module level) is that it needs to be shared and changed by multiple functions:
setAlarm() initializes currentTime and starts the interval that decrements it.
The interval callback inside setAlarm() updates currentTime every second.
updateTimeDisplay() reads currentTime to format and display the remaining time.
pauseAlarm() resets currentTime to 0 and updates the display.

If currentTime were declared as a local variable inside setAlarm(), it would be inaccessible to the interval callback (unless captured via closure, but then it couldn't be modified by pauseAlarm() or reset across multiple alarm sets). 
Each call to setAlarm() would create a new, independent local variable, breaking the ability to stop or reset the timer from outside. By making it a module level variable, all relevant functions share a single source of true for the remaining time, allowing consistent updates and control.

Sprint-3/alarmclock/alarmclock.js
carlosyabreu: Thanks for pointing that out to me.

Sprint-3/alarmclock/alarmclock.js
carlosyabreu: The acceptable range of integers for the alarm clock is 0 or any positive integer (i.e., totalSeconds >= 0).
Negative integers cause abnormal behavior because:
The updateTimeDisplay function will produce malformed strings like -1:-1 (since Math.floor(-1/60) = -1 and -1 % 60 = -1).
The countdown interval checks if (currentTime > 0), which is false for negative values, so the timer never counts down and never triggers the alarm.
Zero is acceptable: it displays 00:00 and does not play the alarm (as expected for an already‑expired timer). Non‑integer numbers are truncated by parseInt, so only the integer part matters.

Sprint-3/alarmclock/alarmclock.js
carlosyabreu: Thanks for pointing that out it makes the understanding more clear to me.
Looking again to the code I don't think so, I think it is not correct to have the timer interval remain active when the input is 0.
When totalSeconds = 0, the interval starts running but the condition if (currentTime > 0) is never true, so the interval never clears itself. This results in a useless, permanent interval that consumes computer resources unnecessarily.
The expected behavior should be: 
if the input is 0, the display should show 00:00 and no interval should be started (and no alarm should play).

Sprint-3/alarmclock/alarmclock.js
carlosyabreu: Thank you @cjyuan.
Your observation adds value to my understanding.

Sprint-3/alarmclock/alarmclock.js
cjyuan: I just noticed your check on line 27. 

Different issue now. When input is 0. the alarm won't sound but the interval won't stop.
Visually the user may not notice anything but the active interval will silently consume (very tiny amount) of CPU.

Sprint-3/alarmclock/alarmclock.js
cjyuan: Why not just treat negative input as if they are `NaN` or just return immediately?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -16,10 +64,25 @@ function setup() {
...
+  
+  // Clear any ongoing timer
+  if (timerInterval) {
+    clearInterval(timerInterval);
+    timerInterval = null;
+  }
+  
+  // Reset display to 00:00
+  currentTime = 0;
```
cjyuan, line 84: Delete line 84 and then move the declaration of `currentTime` from line 2 to inside `setAlarm()`, and your code will still work. 

No change required. You can take time to find out why later. It will help your understanding of the JS language better.

Sprint-3/alarmclock/alarmclock.js
cjyuan: Have you figured out what else need to be reset before a new countdown begins?

cjyuan:
Your app is working but the code has some room to improve. 

I will mark this PR as complete first. You can address the comments later.

illicitonion:

PR #1143, opened 2026-03-28
 Cape Town | 2026-ITP-Jan | Isaac Abodunrin | Sprint 2 | Grouping data: Objects

bytesandroses:
## Learners, PR Template

Self checklist

## Changelist
I completed the following together with tests where applicable:

- Debug (address.js, author.js, recipe.js)
- Implement (contains.js, lookup.js, querystring.js, and tally.js)
- Interpret (invert.js)

### DIFF
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -36,7 +36,7 @@ test("should return false when object does not contain passed property name", ()
       expect(contains(obj, prop)).toEqual(false);
     } catch (error) {
       throw new Error(
-        `Failed to return false when ${prop} is not present in obj: ${error.message}`
+        `Failed to return false when ${ob} is not present in obj: ${error.message}`
       );
     }
   });
@@ -54,4 +54,24 @@ test("should return false when object is empty", () => {
 });
 
 // Case 4: Should throw an error if a non-object is passed
-test("should throw error when non-object is passed");
+test("should throw error when non-object is passed", () => {
+  const nonObjects = [
+    null,
+    undefined,
+    42,
+    "The Curse",
+    true,
+    Infinity,
+    ["string"],
+  ];
+
+  nonObjects.forEach((nonObj) => {
+    try {
+      expect(() => contains(nonObj, "prop")).toThrow();
+    } catch (error) {
+      throw new Error(
+        `Failed to throw error when non-object ${nonObj} is passed: ${error.message}`
+      );
+    }
+  });
+});
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -1,77 +1,79 @@
 const contains = require("./contains.js");
 
-// Case 1: Should return true if the property exists in object.
-test("should return true when object contains passed property name", () => {
-  const objsWithValidProps = [
-    [{ a: 1, b: 2 }, "b"],
-    [{ name: "John", age: 30 }, "name"],
-    [{ nested: { key: "value" } }, "nested"],
-    [{ id: 123, status: "active", language: "JavaScript" }, "status"],
-    [{ data: [], items: null }, "data"],
-  ];
+describe("contains", () => {
+  // Case 1: Should return true if the property exists in object.
+  test("should return true when object contains passed property name", () => {
+    const objsWithValidProps = [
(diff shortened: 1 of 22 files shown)


Sprint-2/implement/contains.test.js
```diff
@@ -1,35 +1,70 @@
...
-// When passed to contains
-// Then it should return false
-test.todo("contains on empty object returns false");
+  // Case 4: Should return false for properties that only exist in the prototype chain
+  test("should return false for properties in prototype chain", () => {
+    const objsWithProtoProps = [
+      [{ a: 1 }, "toString"],
+      [{ name: "John", age: 30 }, "hasOwnProperty"],
+      [{ nested: { key: "value" } }, "isPrototypeOf"],
+    ];
```
cjyuan, line 45: Good idea!

Sprint-2/implement/contains.test.js
```diff
@@ -1,35 +1,70 @@
...
+      Infinity,
+      ["string"],
+    ];
 
-// Given invalid parameters like an array
-// When passed to contains
-// Then it should return false or throw an error
+    nonObjects.forEach((nonObj) => {
+      expect(() => contains(nonObj, "prop")).toThrow(
+        "First argument must be an object in the form { key: value }"
```
cjyuan: Your function implementation is correct.

However, we should write tests not only to verify our current implementation, but also to ensure that future changes do not alter the function's expected behavior.

This test cannot yet confirm the function can correctly return `false` when the first argument is an array. 
This is because `contains(["string"], "prop")` could also return `false` simply because "prop" is not a key of the array.

Arrays are objects, with their indices acting as keys. A proper test should use a valid key to ensure the function returns `false` specifically because the input is an array, not because the key is missing.

Sprint-2/implement/querystring.js
```diff
@@ -1,12 +1,31 @@
...
+        "String must contain both a key and a value separated by '='"
+      );
+    }
+
+    const key = pair.slice(0, separatorIndex).trim();
+    const value = pair.slice(separatorIndex + 1).trim();
+
+    if (key.length === 0 || value.length === 0) {
+      throw new Error("Neither the key nor value can be purely whitespace");
+    }
```
cjyuan, line 27: **No change required.**

Notes: These arguments are also valid query strings:

```javascript
parseQueryString("key=")
parseQueryString("=value")
```

In addition, in real query string, both `key` and `value` are ***percent-encoded*** or ***URL encoded*** in the URL. For example, the string "5%" is encoded as "5%25". So to get the actual value of "5%25"

And in practice, we can use `URLSearchParams` API to parse query string.

Sprint-2/interpret/invert.js
```diff
@@ -1,29 +1,35 @@
...
+);
+
 // a) What is the current return value when invert is called with { a : 1 }
+// { key: 1 }
 
 // b) What is the current return value when invert is called with { a: 1, b: 2 }
+// { key: 2 }
 
 // c) What is the target return value when invert is called with {a : 1, b: 2}
+// { key: 2 }
```
cjyuan: I think "target value" refers to the expected return value after the code is fixed.

cjyuan (commented):
Very comprehensive tests! Impressive.

Sprint-2/implement/contains.test.js
bytesandroses: Thanks CJ, you're absolutely right.

I incorrectly thought that hasOwnProperty() was available only to {} objects, but it's available to all object types, including arrays. So something like `["string"].hasOwnProperty(0)` evaluates to true. This means that future implementations could return true in such a scenario, but my current test fails to test for that by using `["string"].hasOwnProperty("prop")`  since "prop" isn't an index.

To fix this, I've gone ahead and created a separate test case for arrays using `expect(() => contains(["string"], 0)).toThrow()`.

Sprint-2/interpret/invert.js
bytesandroses: Sorry it should be { '1': 'a', '2': 'b' } -- I'd mistakenly copy-pasted the return from (b)

cjyuan:
All good. Well done.

Note: I assume the PR is ready to be re-reviewed.

illicitonion:

PR #1154, opened 2026-03-31
London | 26-ITP-Jan | Kris Oldrini | Sprint 2 | Coursework

XiaoQuark:
## Learners, PR Template

Self checklist

### Debug

- Wrote a prediction of the expected outcome before running the code
- Ran the code and compared the result with the prediction
- Identified and fixed the bugs in the files

### Implement

- Wrote functions based on the specified requirements
- Tested the functions with different inputs
- Added unit tests to confirm the functions behave as expected

### Interpret

- Read and interpreted the program in invert.js
- Answered the required questions about the program
- Fixed the implementation and added tests to confirm the behaviour

## Questions

1. In `contains`, I validated the input object with:

```js
 typeof object !== "object" || object === null || Array.isArray(object)
```
instead of checking for a plain object using:

```js
Object.getPrototypeOf(object) !== Object.prototype
```
Would you recommend the stricter plain-object check here, or is my current validation appropriate for this task?

2. In `contains.test.js`, if the `property` argument is omitted (e.g. `contains({ a: 1 })`), the function currently returns `false`. I did not add a test for this case because the instructions did not mention validating the `property` parameter. Should I have included a test anyway?

3. In lookup, I used Object.fromEntries() because the input was already an array of [key, value] pairs. Is that an appropriate approach for this exercise, or would you prefer an explicit loop implementation to demonstrate the transformation?

### DIFF
--- a/Sprint-2/debug/author.js
+++ b/Sprint-2/debug/author.js
@@ -4,6 +4,9 @@
 // I'm expecting an error
 // For Objects, use for.. in loops
 
+// Update after feedback:
+// I completely forgot `value` on it's own only accesses the property's key. To access the actual value I still need to use bracket notation, author[value]
+
 // This program attempts to log out all the property values in the object.
 // But it isn't working. Explain why first and then fix the problem
 
@@ -16,5 +19,5 @@ const author = {
 };
 
 for (const value in author) {
-  console.log(value);
+  console.log(author[value]);
 }
--- a/Sprint-2/debug/author.js
+++ b/Sprint-2/debug/author.js
@@ -1,5 +1,9 @@
 // Predict and explain first...
 
+// for.. of loops are used with Arrays, not Objects.
+// I'm expecting an error
+// For Objects, use for.. in loops
+
 // This program attempts to log out all the property values in the object.
 // But it isn't working. Explain why first and then fix the problem
 
@@ -11,6 +15,6 @@ const author = {
   alive: true,
 };
 
-for (const value of author) {
+for (const value in author) {
   console.log(value);
 }
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -45,7 +45,7 @@ test("given an object that does not contain the target property, return false", 
 // Then it should return false or throw an error
 
 test("given an array instead of a valid object, throw an error", () => {
-  expect(() => contains(["a", "b", "c"], "3")).toThrow();
+  expect(() => contains(["a", "b", "c"], "2")).toThrow();
 });
 
 test("given null instead of a valid object, throw an error", () => {
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -36,7 +36,7 @@ test("given an object containing the target property, return true", () => {
 // When passed to contains with a non-existent property name
 // Then it should return false
 
-test("given an object that does not contains the target property, return false", () => {
+test("given an object that does not contain the target property, return false", () => {
   expect(contains({ a: 1, b: 2 }, "c")).toBe(false);
 });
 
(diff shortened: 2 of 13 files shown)


Sprint-2/debug/author.js
```diff
@@ -11,6 +15,6 @@ const author = {
   alive: true,
 };
 
-for (const value of author) {
+for (const value in author) {
   console.log(value);
 }
```
cjyuan, line 23: Have you ran your code to ensure it outputs the value of the properties? The values are `'Zadie', 'Smith', 'writer', 40, true`.

Sprint-2/implement/contains.test.js
```diff
@@ -20,16 +20,38 @@ as the object doesn't contains a key of 'c'
...
+  expect(contains({ a: 1, b: 2 }, "c")).toBe(false);
+});
+
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+
+test("given an array instead of a valid object, throw an error", () => {
+  expect(() => contains(["a", "b", "c"], "3")).toThrow();
+});
```
cjyuan, line 49: This test does not yet confirm that the function correctly returns false when the first argument is an array. 
This is because `contains(["a", "b", "c"], "3")` could also return `false` simply because `"3"` is not a key of the array.

Arrays are objects, with their **indices acting as keys.** A proper test should use a valid key to ensure the function returns `false` specifically because the input is an array, not because the key is missing.

Sprint-2/implement/querystring.js
```diff
@@ -6,7 +6,9 @@ function parseQueryString(queryString) {
   const keyValuePairs = queryString.split("&");
 
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
+    if (pair === "") continue;
+    const [key, ...rest] = pair.split("=");
+    const value = rest.join("=");
     queryParams[key] = value;
```
cjyuan, line 12: These arguments are also valid query strings:

```javascript
parseQueryString("key=")
parseQueryString("=value")
```

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,16 @@
-function tally() {}
+function tally(items) {
+  if (!Array.isArray(items)) {
+    throw new Error("Invalid input: items must be an array");
+  }
+  const counts = {};
```
```
```

cjyuan:
For the questions you raised in the PR description, have you tried asking AI?

Sprint-2/debug/author.js
XiaoQuark: This has been fixed and my comments in the file edited

Sprint-2/implement/contains.test.js
XiaoQuark: I changed the value of `property` to `"2"`, for this test. This should now make the test work correctly.

Sprint-2/implement/querystring.js
XiaoQuark: I've added tests for the additional valid query string formats. 
Some of those cases were already covered by existing tests, so I avoided duplicating equivalent scenarios. 

I chose not to add the case `parseQueryString("key1=value1&key2")`, because I am already testing for `parseQueryString("name=&position=developer")` and it seemed redundant. If that's not the case I will add it too.

Sprint-2/implement/tally.js
XiaoQuark: I added a new test and fixed the function as requested.

XiaoQuark:
> For the questions you raised in the PR description, have you tried asking AI?

I did. It's because `Object.getPrototypeOf(object) !== Object.prototype` is a stricter method that would also return true for Map() and Class(), as their prototype would be different. 
But this is why I'm wondering if it wouldn't actually be better than the other, less strict method.

Sprint-2/debug/author.js
cjyuan: Could consider naming the variable `key` instead of `value`.

cjyuan:
1. **object** -- a kind of data with certain characteristics (you can look up its definition). 
2. When `Object.getPrototype(obj) == Object.prototype"` is true, it only means `obj` is an object created using `Object` as it prototype/template, but not all objects meet this requirement.
    - You can ask AI "What objects do not meet the condition: Object.getPrototype(obj) == Object.prototype in JS?"

In this exercise, I think the term "object" simply refers to the general object, which can be checked via the condition `typeof object == "object" && object != null`. The concept of "prototype" might be too advanced at this stage.

cjyuan:
Changes look good.

illicitonion:

PR #1155, opened 2026-03-31
Sheffield | 26-ITP-Jan | Mahmoud Shaabo | Sprint 3 | Alarm Clock App

mahmoudshaabo1984:
Hello Reviewers / CodeYourFuture Team,

I have completed the Alarm Clock App project for Sprint 3. 

Acceptance Criteria Met:
- Updated the title in `index.html` to "Alarm clock app".
- Implemented `setAlarm()` in `alarmclock.js` to read the input and display the time in `mm:ss` format.
- Used `setInterval()` to decrement the timer every 1 second.
- When the timer reaches `00:00`, the interval is cleared, the background changes to red, and the alarm audio plays.
- The "Stop Alarm" button clears the interval, resets the background, and pauses the audio.
- All Jest tests are passing locally.

Thank you for your review!

### DIFF
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -10,12 +10,18 @@ function updateTimeDisplay(totalSeconds) {
     "Time Remaining: " + mins + ":" + secs;
 }
 
+// Helper function: show or clear the feedback message in a separate element
+function showFeedback(message) {
+  document.getElementById("feedback").innerText = message;
+}
+
 function setAlarm() {
   // Clear any previously running countdown
   clearInterval(interval);
 
-  // Reset background to default
-  document.body.style.backgroundColor = "";
+  // Remove alarm style and clear feedback message
+  document.body.classList.toggle("alarm-activated", false);
+  showFeedback("");
 
   // Stop any currently playing alarm sound before starting a new countdown
   // (the user may not click "Stop" first before setting a new alarm)
@@ -26,15 +32,14 @@ function setAlarm() {
 
   // Validate input: reject NaN, negative numbers, or empty field
   if (isNaN(totalSeconds) || totalSeconds < 0) {
-    document.getElementById("timeRemaining").innerText =
-      "Please enter a valid number of seconds (0 or above).";
+    showFeedback("Please enter a valid number of seconds (0 or above).");
     return;
   }
 
   // If input is 0, trigger the alarm immediately instead of waiting 1 second
   if (totalSeconds === 0) {
     updateTimeDisplay(0);
-    document.body.style.backgroundColor = "red";
+    document.body.classList.toggle("alarm-activated", true);
     playAlarm();
     return;
   }
@@ -52,8 +57,8 @@ function setAlarm() {
 
       updateTimeDisplay(0);
 
-      // Change background colour to signal the alarm
-      document.body.style.backgroundColor = "red";
+      // Apply alarm style using a CSS class (separates presentation from logic)
+      document.body.classList.toggle("alarm-activated", true);
 
       // Trigger the alarm sound (defined below the DO NOT EDIT line)
       playAlarm();
@@ -71,7 +76,7 @@ function setAlarm() {
 document.addEventListener("DOMContentLoaded", function () {
   document.getElementById("stop").addEventListener("click", function () {
     clearInterval(interval);
-    document.body.style.backgroundColor = "";
+    document.body.classList.toggle("alarm-activated", false);
   });
 });
 
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -1,6 +1,15 @@
 // Stores the interval ID so we can clear it later
 let interval = null;
 
+// Helper function: format seconds as mm:ss and update the display
+// This keeps the time-display logic in ONE place (CJ feedback: avoid repetition)
+function updateTimeDisplay(totalSeconds) {
+  const mins = String(Math.floor(totalSeconds / 60)).padStart(2, "0");
+  const secs = String(totalSeconds % 60).padStart(2, "0");
+  document.getElementById("timeRemaining").innerText =
+    "Time Remaining: " + mins + ":" + secs;
+}
+
 function setAlarm() {
   // Clear any previously running countdown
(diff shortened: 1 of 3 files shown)


mahmoudshaabo1984:
Hi @cifarquhar,

Here is my Pull Request for the Alarm Clock App. I successfully implemented the timer logic using `setInterval` and `clearInterval`, and handled the math to format the remaining seconds into a standard `mm:ss` display. All Jest tests passed successfully.

Looking forward to your feedback!

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,52 @@
...
+  let totalSeconds = parseInt(document.getElementById("alarmSet").value);
+
+  // --- Format and display the initial time ---
+  // Math.floor(119 / 60) = 1  --> minutes
+  // 119 % 60 = 59             --> leftover seconds
+  // padStart(2, "0") ensures single digits become "01", "09", etc.
+  const mins = String(Math.floor(totalSeconds / 60)).padStart(2, "0");
+  const secs = String(totalSeconds % 60).padStart(2, "0");
+  document.getElementById("timeRemaining").innerText =
+    "Time Remaining: " + mins + ":" + secs;
```
cjyuan: Instead of keeping several copies of the code for displaying time (lines 18-21, lines 44-47, lines 32-33),
it is better to define a function to update the time display and call the function to update the time display.

Advantages:
- Keep the logic for updating time display in one place
- Less code

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,52 @@
...
+
+function setAlarm() {
+  // Clear any previously running countdown
+  clearInterval(interval);
+
+  // Reset background to default
+  document.body.style.backgroundColor = "";
+
+  // Read the input value and convert it to a plain integer (total seconds)
+  let totalSeconds = parseInt(document.getElementById("alarmSet").value);
```
cjyuan, line 31: Some integer values or unusual input can still make your app behave abnormally. Can you figure out a way to prevent that from happening?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,52 @@
...
+// Stores the interval ID so we can clear it later
+let interval = null;
+
+function setAlarm() {
+  // Clear any previously running countdown
+  clearInterval(interval);
+
+  // Reset background to default
+  document.body.style.backgroundColor = "";
+
```
cjyuan, line 25: Currently when starting a new countdown, the application does not always return to a clean initial state, 
which can lead to inconsistent behaviour between runs. In addition to resetting the countdown and the background color, what else should also be reset? (Hint: a user may not click the "Stop" button first before starting a new count down.)

cjyuan (commented):
- Can you make the "Stop" button to also stop the count down? You can add another "click" event listener to the button to avoid modifying the original code.

- When input is 1, the alarm is played one second later -- an expected behavior.
However, when input is 0, the alarm is also played one second later (instead of being played immediately). Can you make the app's behaviour more consistent?

mahmoudshaabo1984:
Hi CJ,

Thank you for the excellent feedback! I have just pushed an update to address all your points:

1. Created a helper function to format and display the time, avoiding code duplication (DRY principle).
2. Made sure the "Stop" button also clears the countdown interval.
3. Updated the logic so if the input is 0, the alarm triggers immediately.
4. Added validation to reject empty inputs or negative numbers.
5. Ensured `pauseAlarm()` is called at the start of a new alarm to maintain a clean state.

All Jest tests are still passing successfully. Please let me know if everything looks good to go!

Best regards,
Mahmoud

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,79 @@
...
+  // (the user may not click "Stop" first before setting a new alarm)
+  pauseAlarm();
+
+  // Read the input value and convert it to a plain integer (total seconds)
+  let totalSeconds = parseInt(document.getElementById("alarmSet").value);
+
+  // Validate input: reject NaN, negative numbers, or empty field
+  if (isNaN(totalSeconds) || totalSeconds < 0) {
+    document.getElementById("timeRemaining").innerText =
+      "Please enter a valid number of seconds (0 or above).";
```
cjyuan: It is better to introduce a separate element to show feedback message, and leave "#timeRemaining" for only displaying time.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,79 @@
...
+    totalSeconds--; // subtract 1 second
+
+    if (totalSeconds <= 0) {
+      // Stop the interval when time runs out
+      clearInterval(interval);
+
+      updateTimeDisplay(0);
+
+      // Change background colour to signal the alarm
+      document.body.style.backgroundColor = "red";
```
```
```

cjyuan (commented):
Previous changes are good.

I just have two more suggestions for you to consider.

mahmoudshaabo1984:
Hi CJ,

Thank you for the excellent and detailed feedback! I have just pushed a new update to address your additional suggestions:

1. Added a separate `<p id="feedback">` element in HTML to handle error messages, keeping `#timeRemaining` strictly for the timer.
2. Refactored the background color logic to use `document.body.classList.toggle("alarm-activated")`, completely separating the presentation logic (CSS) from the application logic (JS).

All Jest tests are passing perfectly. I really appreciate these advanced tips! Please let me know if it is good to go.

Best regards,
Mahmoud

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,84 @@
...
+  // Clear any previously running countdown
+  clearInterval(interval);
+
+  // Remove alarm style and clear feedback message
+  document.body.classList.toggle("alarm-activated", false);
+  showFeedback("");
+
+  // Stop any currently playing alarm sound before starting a new countdown
+  // (the user may not click "Stop" first before setting a new alarm)
+  pauseAlarm();
```
cjyuan, line 28: As the resetting logic is becoming more complicated, you can consider placing them in a function.

cjyuan (commented):
Changes look good. Well done.

illicitonion:

PR #1162, opened 2026-04-02
Sheffield | ITP-Jan-26 | Mahammad Osman | Sprint 3 | Alarm Clock

Darkidd77:
## Learners, PR Template

Self checklist

## Changelist

- Connected the HTML inputs to the `setAlarm` function.
- Implemented the countdown engine using `setInterval` (1000ms delay).
- Added logic to mathematically split raw seconds into minutes and seconds, formatted to `MM:SS` using `.padStart(2, '0')`.
- Added safety checks to `clearInterval` both when the timer hits zero, and if the user clicks "Set Alarm" while a timer is already running.

### DIFF
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -1,4 +1,71 @@
-function setAlarm() {}
+let timerInterval;
+let flashInterval;
+
+function setAlarm() {
+  const inputElement = document.getElementById("alarmSet");
+  const headingElement = document.getElementById("timeRemaining");
+
+  if (!inputElement.value) {
+    return;
+  }
+
+  if (timerInterval) {
+    clearInterval(timerInterval);
+  }
+  if (flashInterval) {
+    clearInterval(flashInterval);
+  }
+
+  let timeInSeconds = Number(inputElement.value);
+
+  function updateScreen(secondsRemaining) {
+    const minutes = Math.floor(secondsRemaining / 60);
+
+    const seconds = secondsRemaining % 60;
+
+    const formattedMinutes = String(minutes).padStart(2, "0");
+    const formattedSeconds = String(seconds).padStart(2, "0");
+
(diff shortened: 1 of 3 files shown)


Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,71 @@
-function setAlarm() {}
+let timerInterval;
+let flashInterval;
+
+function setAlarm() {
+  const inputElement = document.getElementById("alarmSet");
+  const headingElement = document.getElementById("timeRemaining");
+
+  if (!inputElement.value) {
+    return;
```
Luro91, line 9: It's important to validate user input. 

What happens if a user enters a negative value?
How could you give feedback to the user about the input not being valid?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,71 @@
...
+    // when we hit zero?
+    if (timeInSeconds <= 0) {
+      clearInterval(timerInterval);
+      playAlarm();
+      flashScreen();
+    }
+  }, 1000);
+}
+
+function flashScreen() {
```
Luro91, line 56: Nice bonus implementation of the flashing screen

Darkidd77: Thank you

Sprint-3/alarmclock/alarmclock.js
Darkidd77: Yes, I realised that I have to validate the input to avoid unexpected actions from the code, as if it's negative values it trigger the alarm as its less than zero, i did added a check after parsing the input to make sure its not negative and not a NaN and return a message informing the user to input a valid value. Thank you

Luro91: You could remove the condition for checking "!inputElement.value" now and save some line of codes.

Darkidd77: Noted! Thank you

illicitonion:

PR #1163, opened 2026-04-02
Sheffield | 26-ITP-Jan | Martha Ogunbiyi | Sprint 3| Sprint3 /Quote generator

marthak1:
## Learners, PR Template

Self checklist

## Changelist

Built quote generator app as per assignment requirements

### DIFF
--- a/Sprint-3/quote-generator/quotes.js
+++ b/Sprint-3/quote-generator/quotes.js
@@ -489,7 +489,8 @@ const quotes = [
     author: "Zig Ziglar",
   },
 ];
-
+const title = document.querySelector("title")
+title.textContent = "Quote generator app";
 // call pickFromArray with the quotes array to check you get a random quote
 const quoteEl = document.querySelector("#quote");
 const authorEl = document.querySelector("#author");
--- a/Sprint-3/quote-generator/quotes.js
+++ b/Sprint-3/quote-generator/quotes.js
@@ -489,8 +489,7 @@ const quotes = [
     author: "Zig Ziglar",
   },
 ];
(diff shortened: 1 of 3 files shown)


Sprint-3/quote-generator/quotes.js
```diff
@@ -489,5 +489,22 @@ const quotes = [
     author: "Zig Ziglar",
   },
 ];
-
+const title = document.querySelector("title")
+title.textContent = "Quote generator app";
```
cjyuan: Why not just set the title in HTML?

Sprint-3/quote-generator/quotes.js
```diff
@@ -489,5 +489,22 @@ const quotes = [
...
+
+  quoteEl.textContent = randomQuote.quote;
+  authorEl.textContent = randomQuote.author;
+}
+
+// Show initial quote on page load
+displayQuote();
+
+// Change quote when button is clicked
+button.addEventListener("click", displayQuote);
```
Would be even better to place the code 
inside a function to make it clearer that "this is what runs when the page loads."

For examples,

```javascript
function setup() {
}

```
or
```javascript
});
```

Sprint-3/quote-generator/quotes.js
marthak1: Thanks for the feedback , title has now been set in html file.

Sprint-3/quote-generator/quotes.js
marthak1: thanks for the feed, onl page load function added

cjyuan:
All good!

illicitonion:

PR #1172, opened 2026-04-03
Sheffield | 26-ITP-Jan | Seti Mussa | Sprint 2 | Data Group

Seti-Jemal:
Self checklist

## Changelist
I have answered the task as required.


Sprint-2/debug/recipe.js
```diff
@@ -12,4 +13,4 @@ const recipe = {
 
 console.log(`${recipe.title} serves ${recipe.serves}
   ingredients:
-${recipe}`);
+${recipe.ingredients.join('\n  ')}`);
```
cjyuan, line 16: Why place two extra space in the separator?

Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,9 @@
-function contains() {}
+function contains(obj, prop) {
+  if (typeof obj !== "object" || obj === null || Array.isArray(obj)) {
+    return false;
+  }
+
+  return Object.prototype.hasOwnProperty.call(obj, prop);
```
cjyuan: Can also explore `Object.hasOwn()`.

Sprint-2/implement/contains.test.js
```diff
@@ -20,16 +20,26 @@ as the object doesn't contains a key of 'c'
...
-
+test("return false when property does not exit", () => {
+  expect(contains({ a: 1, b: 2 }, "c")).to(fasle);
+});
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+test("return false for invalid input like an array", () => {
+  expect(contains([], "a")).toBe(false);
+});
```
cjyuan, line 45: This test does not yet confirm that the function correctly returns false when the first argument is an array. 

Sprint-2/implement/querystring.js
```diff
@@ -1,12 +1,20 @@
...
+
+  const queryParams = {};
   const keyValuePairs = queryString.split("&");
 
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
+    if (!pair) continue;
+
+    const index = pair.indexOf("=");
+    if (index === -1) continue;
```
cjyuan, line 13: FYI: `"key1&key2"` is also a valid query string.

No change required.

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,15 @@
...
+function tally(items) {
+  if (!Array.isArray(items)) {
+    throw new Error("Expected an array");
+  }
+
+  const counts = {};
+
+  for (const item of items) {
+    counts[item] = (counts[item] || 0) + 1;
+  }
```
cjyuan, line 10: Does the following function call return the value you expect?
```
```

Sprint-2/stretch/mode.js
cjyuan, line 1: Have you tried running this script or the tests in `mode.test.js`?

Sprint-2/stretch/till.js
```diff
@@ -23,9 +23,38 @@ const till = {
...
+  };
+
+  expect(totalTill(till)).toEqual("£4.40");
+});
+//then fix the implementation of totalTill
+function totalTill(till) {
+  let total = 0;
+
+  for (const [coin, quantity] of Object.entries(till)) {
+    const value = parseInt(coin);
```
cjyuan, line 55: Can you explain why you choose to use `parseInt(coin)` instead of `Number(coin)`?

Sprint-2/implement/querystring.js
Seti-Jemal: Thanks for the note.

Sprint-2/debug/recipe.js
Seti-Jemal: Because I thought it would error but it didn't , it showed "[ object Object]" so I used join() to put ingredients on new lines.

Sprint-2/implement/contains.js
Seti-Jemal: I have applied it.

Sprint-2/implement/tally.js
Seti-Jemal: I have looked up. I changed to object.create (null) so there is no inherited properties.

Sprint-2/stretch/till.js
Seti-Jemal: I used parseInt because it gets the number from a string like "50", but Number() would give NaN.

Sprint-2/debug/recipe.js
cjyuan: I mean, why use `recipe.ingredients.join('\n  ')` and not `recipe.ingredients.join('\n')`?

Sprint-2/stretch/till.js
cjyuan: Both `parseInt("50")` and `Number("50")`produce `50` though. Can you elaborate?

Sprint-2/debug/recipe.js
Seti-Jemal: It just from writing mistake.

Sprint-2/stretch/till.js
Seti-Jemal: Yeah that works for `50`, both parseInt and Number give 50. But my coins are like `50p`, so pareseInt works because it read the number at the start and ignores the 'p'. Number(`50p`) would return NaN because it tries to convert the whole string and fails.

cjyuan:
All good. Well done.

illicitonion:

PR #1178, opened 2026-04-04
Sheffield | 26-ITP-Jan | Seti Mussa | Sprint 3 | Alarm Clock

Seti-Jemal:
Self checklist

## Changelist

I have completed sprint 3 alarm clock.

### DIFF
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -1,5 +1,6 @@
 let timeRemaining = 0;
 let timerId = null;
+let isPaused = false;
 
 const audio = new Audio("alarmsound.mp3");
 
@@ -14,10 +15,10 @@ function formatTime(seconds) {
 // UPDATE DISPLAY
 function updateDisplay(time) {
   const heading = document.getElementById("timeRemaining");
-  heading.textContent = "Time Remaining: " + formatTime(time);
+  heading.textContent = formatTime(Math.max(0, time));
 }
 
-// REQUIRED BY TESTS
+// PLAY ALARM
 function playAlarm() {
   audio.loop = true;
   audio.currentTime = 0;
@@ -30,28 +31,36 @@ function triggerAlarm() {
   playAlarm();
 }
 
-// STOP / RESET ALARM STATE
+// STOP / PAUSE / RESUME
 function stopAlarm() {
-  if (typeof audio.pause === "function") {
-    try {
-      audio.pause();
-    } catch (error) {
-      // ignore jsdom audio pause errors
-    }
+  // 1. If timer is running → pause
+  if (timerId !== null) {
+    clearInterval(timerId);
+    timerId = null;
+    isPaused = true;
+    return;
   }
 
-  audio.currentTime = 0;
-  audio.loop = false;
+  // 2. If paused → resume
+  if (isPaused && timeRemaining > 0) {
+    startTimer();
+    isPaused = false;
+    return;
+  }
 
-  clearInterval(timerId);
-  timerId = null;
+  // 3. If alarm is playing → stop it
+  if (timeRemaining <= 0) {
+    audio.pause();
+    audio.currentTime = 0;
+    audio.loop = false;
 
-  document.body.classList.toggle("alarm-activated", false);
+    document.body.classList.toggle("alarm-activated", false);
+  }
 }
 
 // START TIMER
 function startTimer() {
-  stopAlarm();
+  clearInterval(timerId);
 
   timerId = setInterval(() => {
     timeRemaining = timeRemaining - 1;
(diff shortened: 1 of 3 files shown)


Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,25 +1,93 @@
...
+}
 
-  document.getElementById("stop").addEventListener("click", () => {
-    pauseAlarm();
-  });
+// UPDATE DISPLAY (MUST MATCH TEST EXACTLY)
+function updateDisplay() {
+  const heading = document.getElementById("timeRemaining");
+  heading.textContent = "Time Remaining: " + formatTime(timeRemaining);
 }
```
cjyuan: Why not include a parameter to
- make this function reusable (e.g. call `updateDisplay(0)` to reset timer display)
- avoid using a global variable `timeRemaining`

?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,25 +1,93 @@
...
+
+  clearInterval(timerId);
+  timerId = null;
+
+  document.body.style.backgroundColor = "";
+}
+
+// TRIGGER ALARM
+function triggerAlarm() {
+  document.body.style.backgroundColor = "red";
```
```
```

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,25 +1,93 @@
...
+  playAlarm();
+}
+
+// START TIMER
+function startTimer() {
+  clearInterval(timerId);
```

cjyuan (commented):
- Given a countdown is still active, when the users click the "Set Alarm" button with input equals to 0, then what do you expect the app to behave?

- While a countdown is active, does the timer display look normal to you?

---
Note: We should respect instructions like `DO NOT EDIT BELOW HERE`; it is usually there for a reason. If you are curious about why, you can ask AI `Why should programmers respect "DO NOT EDIT BELOW HERE" instruction in a file?` (You have made some major changes to the original code, so don't worry about that this time).

Sprint-3/alarmclock/alarmclock.js
Seti-Jemal: Thanks for the feedback. I have updated updateDisplay(0) to take a parameter so it's reusable and doesn't rely on the global variable.

Sprint-3/alarmclock/alarmclock.js
Seti-Jemal: I have replaced the inline background style with CSS class and used classList.togg() to separate presentation from logic.

Sprint-3/alarmclock/alarmclock.js
Seti-Jemal: I changed the time logic so starting a new countdown resets the app to clean the state before running again.

cjyuan:
No new commits on this branch on GitHub yet.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,25 +1,90 @@
...
+  return `${String(mins).padStart(2, "0")}:${String(secs).padStart(2, "0")}`;
+}
 
-  document.getElementById("stop").addEventListener("click", () => {
-    pauseAlarm();
-  });
+// UPDATE DISPLAY
+function updateDisplay(time) {
+  const heading = document.getElementById("timeRemaining");
+  heading.textContent = "Time Remaining: " + formatTime(time);
```
cjyuan: Does the display of countdown looks "normal" to you?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,25 +1,90 @@
...
+    timeRemaining--;
+
+    updateDisplay(timeRemaining);
+
+    if (timeRemaining <= 0) {
+      clearInterval(timerId);
+      timerId = null;
+
+      timeRemaining = 0;
+      updateDisplay(timeRemaining);
```
cjyuan: When the condition at line 54 becomes true, what value is `timeRemaining`? 
Is there any need to perform the tasks on line 58-59?

Seti-Jemal: No, those lines are not needed because timeRemaining is already on 0.

Sprint-3/alarmclock/alarmclock.js
cjyuan: Didn't you notice anything odd on the timer display?

If you cannot see it, can you capture a screenshot of your app and then post in here?

Sprint-3/alarmclock/alarmclock.js
cjyuan: If they are not needed, why not remove them?

Sprint-3/alarmclock/alarmclock.js
Seti-Jemal: Sorry, I made the changes on the old branch. Now I have update the app it display correctly.

cjyuan (commented):
Looks good.

illicitonion:

PR #1181, opened 2026-04-05
 Cape Town | 2026-ITP-Jan | Isaac Abodunrin | Sprint 3 | Grouping data: Quote generator

bytesandroses:
## Self checklist

## Changelist
- Created a responsive quote generator webpage styled using `quote_generator_example.png` as a guide.
- When the page loads (or the New quotes button pressed), a quote is randomly selected and the following is displayed:
  - Quote
  - Author
  - New quote button

### DIFF
--- a/Sprint-3/quote-generator/quotes.js
+++ b/Sprint-3/quote-generator/quotes.js
@@ -1,3 +1,17 @@
+const newQuoteButton = document.querySelector("#new-quote");
+
+function displayQuote() {
+  const randomQuote = pickFromArray(quotes);
+  const quote = document.querySelector("#quote");
+  const author = document.querySelector("#author");
+
+  quote.textContent = randomQuote.quote;
+  author.textContent = `- ${randomQuote.author}`;
+}
+
+newQuoteButton.addEventListener("click", displayQuote);
+window.addEventListener("load", displayQuote);
+
 // DO NOT EDIT BELOW HERE
 
 // pickFromArray is a function which will return one item, at
--- a/Sprint-3/quote-generator/quotes.js
+++ b/Sprint-3/quote-generator/quotes.js
@@ -1,5 +1,3 @@
-const newQuoteButton = document.querySelector("#new-quote");
-
 function displayQuote() {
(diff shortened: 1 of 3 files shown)


Sprint-3/quote-generator/quotes.js
```diff
@@ -1,3 +1,17 @@
+const newQuoteButton = document.querySelector("#new-quote");
+
+function displayQuote() {
+  const randomQuote = pickFromArray(quotes);
+  const quote = document.querySelector("#quote");
+  const author = document.querySelector("#author");
+
+  quote.textContent = randomQuote.quote;
+  author.textContent = `- ${randomQuote.author}`;
```
cjyuan: The leading `-` appears to be for styling purposes. Keeping it in the CSS or in HTML could allow us to adjust the view without changing any JavaScript code.

Sprint-3/quote-generator/quotes.js
```diff
@@ -1,3 +1,17 @@
...
+}
+
+newQuoteButton.addEventListener("click", displayQuote);
+window.addEventListener("load", displayQuote);
```
cjyuan: The code that setup the event listener once is also part of the "run on load" code. 
You can consider including all the "run on load" code in a function to make it clearer 
that "this is what runs when the page loads."

For examples,

```javascript
function setup() {
}

```
or
```javascript
});
```

bytesandroses: Thanks for the suggestion CJ. I've gone ahead and wrapped the code needed to run on loading into a setup function.

Sprint-3/quote-generator/quotes.js
bytesandroses: You're absolutely right -- removing the `-` from JavaScript would be better for separation of concerns.  I've moved the `-` into the CSS. I've also moved the leading `“` in front of each quote from the HTML into the CSS so it's more consistent.

cjyuan:
Looks good. Well done.

illicitonion:

PR #1182, opened 2026-04-06
 Cape Town | 2026-ITP-Jan | Isaac Abodunrin | Sprint 3 | Grouping data: Alarm Clock

bytesandroses:
## Self checklist

## Changelist
Implemented alarm clock functionality and countdown timer logic to

- set alarm 
- format time correctly
- display the remaining time as it counts down

### DIFF
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -2,9 +2,13 @@ let timeRemaining;
 let timerInterval;
 
 function formatTime(time) {
-  const minutes = Math.floor(time / 60);
-  const seconds = time % 60;
-  return `${minutes}:${seconds < 10 ? "0" : ""}${seconds}`;
+  const minutes =
+    Math.floor(time / 60) < 10
+      ? "0" + Math.floor(time / 60)
+      : Math.floor(time / 60);
+  const seconds = time % 60 < 10 ? "0" + (time % 60) : time % 60;
+
+  return `${minutes}:${seconds}`;
 }
 
 function displayAlarm(time) {
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -2,11 +2,8 @@ let timeRemaining;
 let timerInterval;
 
 function formatTime(time) {
-  const minutes =
-    Math.floor(time / 60) < 10
-      ? "0" + Math.floor(time / 60)
-      : Math.floor(time / 60);
-  const seconds = time % 60 < 10 ? "0" + (time % 60) : time % 60;
+  const minutes = String(Math.floor(time / 60)).padStart(2, "0");
+  const seconds = String(time % 60).padStart(2, "0");
 
   return `${minutes}:${seconds}`;
 }
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -14,15 +14,15 @@ function displayAlarm(time) {
 }
 
 function decreaseAlarmTime() {
+  timeRemaining--;
+  displayAlarm(timeRemaining);
+
   if (timeRemaining <= 0) {
     clearInterval(timerInterval);
     timerInterval = null;
     playAlarm();
     return;
   }
-
-  timeRemaining--;
-  displayAlarm(timeRemaining);
 }
 
 function resetAlarm() {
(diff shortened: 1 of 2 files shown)


Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,6 +1,42 @@
...
+    return;
+  }
+
+  timeRemaining--;
+  displayAlarm(timeRemaining);
+}
+
+function setAlarm() {
+  const setTime = document.getElementById("alarmSet").value;
+  timeRemaining = parseInt(setTime, 10);
```
cjyuan: Using user input without proper validation and sanitisation can be dangeours.

Currently, some unusual input values can make your app behave abnormally. Can you add code to sanitise them?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,6 +1,42 @@
...
+
+  if (timerInterval) {
+    clearInterval(timerInterval);
+  }
```
cjyuan: The countdown interval is not the only state to be reset.

You can also consider introducing a dedicated reset function to return the app to a clean initial state to help ensure consistency.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,6 +1,42 @@
...
+    clearInterval(timerInterval);
+    timerInterval = null;
+    timeRemaining = 0;
+    playAlarm();
+    return;
+  }
+
+  timeRemaining--;
+  displayAlarm(timeRemaining);
+}
```
cjyuan, line 26: When the countdown reaches 00:00, there is a one second delay before the alarm sound is played. Is this by design?

At line 25, when `timeRemaining` changes from 1 to 0, the app only changes the display to 00:00. It's only in the next interval, the code on lines 17-22 is executed.

Sprint-3/alarmclock/alarmclock.js
bytesandroses: You're absolutely right -- the timer behaves oddly if certain inputs are automatically allowed without sanitizing. If a user enters a negative number, a negative timer is displayed which can't be counted down. And if a number isn't entered, it simply displays NaNs. 

To fix this, I've added two validation checks:

1. Ensure the input is a number
2. Ensure the number is positive

Sprint-3/alarmclock/alarmclock.js
bytesandroses: I see what you mean. The `clearInterval(timerInterval)` only stops the timer counting down -- it doesn't remove it. So it creates a situation where multiple timers can run at the same time.

I've taken your suggestion and handled it in s single `reset()` function.

Sprint-3/alarmclock/alarmclock.js
bytesandroses: No this definitely a bug -- it wasn't by design. 

I've applied a fix by changing the order of execution: `timeRemaining` is first decremented so that the alarm sound can be played at exactly 00:00 without needing to start a new loop.

cjyuan (commented):
**When input is 0**, there is a delay of one second before the alarm is sounded, and the timer display will mess up.

Note: In `setAlarm()`
```javascript

   // This only setups decreaseAlarmTime() to be called one second later (and every second thereafter)
   timerInterval = setInterval(decreaseAlarmTime, 1000);  

  displayAlarm(timeRemaining);  // This line is executed first before decreaseAlarmTim() is called
```

Sprint-3/alarmclock/alarmclock.js
cjyuan: I meant to point out "alarm not reset", which you have handled now in the `reset()` function.

In JS, calling `clearInterval(timerInterval)` is enough to remove the active interval.  
Performing `timerInterval = null;` is optional but a good practice.

bytesandroses:
> **When input is 0**, there is a delay of one second before the alarm is sounded, and the timer display will mess up.
> 
> Note: In `setAlarm()`
> 
> ```js
>    // This only setups decreaseAlarmTime() to be called one second later (and every second thereafter)
>    timerInterval = setInterval(decreaseAlarmTime, 1000);  
> 
>   displayAlarm(timeRemaining);  // This line is executed first before decreaseAlarmTim() is called
> ```

Thank CJ, you're absolutely correct. There's a bug when a user enters a value of 0. 

To fix it, I've added an extra step in `playAlarm()` to check whether the user has inputted a value of 0. If it's 0, then the alarm sound is automatically played without the 1 second delay. Otherwise the countdown is run as usual.

cjyuan:
Changes look good.

illicitonion:

PR #1183, opened 2026-04-06
London | 26-ITP-Jan | Karla Grajales | Sprint 3 | Alarm Clock App

Grajales-K:
## Learners, PR Template

Self checklist

## Changelist

This application features a dynamic countdown engine built with JavaScript that accurately converts numerical user input into a ` mm:ss` display format. It integrates a multisensory alert system that triggers both a continuous audio track and a synchronised disco background effect using random RGB generation once the alarm reaches zero.

The core logic was optimised for test-driven development, implementing a repetition counter to prevent infinite loops and ensure all Jest unit tests pass successfully. The project includes an input validation to handle non-numeric values, ensuring a stable and functional application.


Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,8 +1,59 @@
-function setAlarm() {}
+let countdown;
```
cjyuan: Consider renaming `countdown` to `countdownInterval` or `intervalId` since this stores the interval ID, not the countdown value.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,8 +1,59 @@
...
+
+function setAlarm() {
+  clearInterval(countdown);
+  document.body.style.backgroundColor = "white";
```
which can lead to inconsistent behavior between runs.

You can also consider introducing a dedicated reset function to return the app to a clean initial state to help ensure consistency. There are few places in this script you can call this reset function instead of repeating the reset logic.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,8 +1,59 @@
...
+
+  let inputTime = Number(document.getElementById("alarmSet").value);
+
+  if (isNaN(inputTime) || inputTime <= 0) {
+    alert("Please type or select your time 👇⏰");
+    return;
+  }
```
cjyuan: What type of numbers should `inputTime` be? (Currently some input value can corrupt the timer display)

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,8 +1,59 @@
...
+  }, 1000);
+}
+
+function updateDisplay(seconds) {
+  let minutes = Math.floor(seconds / 60);
+  let remainingSeconds = seconds % 60;
+
+  let display = `Time Remaining: ${minutes
+    .toString()
+    .padStart(2, "0")}:${remainingSeconds.toString().padStart(2, "0")}`;
```
cjyuan: Having a long expression within a template string literal can make the code harder to read.
Better to use an inline function to perform the repeated task, or to store the formatted
values in variables first.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,8 +1,59 @@
...
+      countdown = setInterval(() => {
+        document.body.style.backgroundColor = `rgb(
+        ${Math.floor(Math.random() * 256)},
+        ${Math.floor(Math.random() * 256)}, 
+        ${Math.floor(Math.random() * 256)})`;
+        repetitions++;
+
+        if (repetitions > 100) {
+          clearInterval(countdown);
+        }
```
cjyuan: It is best practice to separate presentation logic from application logic.
Can you either implementing the background animation in CSS or move the code that perform the animation into a separate function?

Note: CSS random() function is still in experimental stage. With pure CSS, we can only cycle through a pre-defined set of colors (e.g. https://codepen.io/beben-koben/pen/eYPNew)

If you are implementing this purely in CSS, you can consider defining a CSS class, and use [`classList.toggle()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/classList) to apply/remove the style. For example,
```
```

Sprint-3/alarmclock/alarmclock.js
Grajales-K: Thanks @cjyuan, for the feedback. I changed the variable name to countdownInterval to clarify that it holds the interval ID, not the numeric time.

Sprint-3/alarmclock/alarmclock.js
Grajales-K: Thanks for the feedback! I have introduced a resetApp() function to ensure the application returns to a clean state every time a new alarm is set or stopped. This function handles clearing the interval, resetting the background color, and stopping the audio, avoiding inconsistent behavior if the 'Stop' button isn't clicked.

Sprint-3/alarmclock/alarmclock.js
Grajales-K: Thanks @cjyuan, I've updated the validation to use Number.isInteger(). This ensures the app only accepts positive whole numbers, preventing decimal values (like 9.5) from causing visual glitches or inconsistent countdowns.

Sprint-3/alarmclock/alarmclock.js
Grajales-K: Thanks! I updated the code to use a ternary operator to format the zeros. I also stored the values in variables first to make the template string cleaner and easier to read

Sprint-3/alarmclock/alarmclock.js
Grajales-K: Thanks for the feedback! I've separated the presentation logic by moving the background animation code into its own function (startBackgroundAnimation). This keeps the setAlarm function focused on the core logic while making the animation code easier to manage independently.

Grajales-K:
Apologies for the delay, @cjyuan. I hope the changes are good. Thank you for your time reviewing my work.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,8 +1,84 @@
...
+function setAlarm() {
+  resetApp();
+
+  const rawValue = document.getElementById("alarmSet").value;
+  const inputTime = Number(rawValue);
+
+  if (rawValue === "" || !Number.isInteger(inputTime) || inputTime <= 0) {
+    alert("Please type or select a valid positive time in seconds 👇⏰");
+    return;
+  }
```
cjyuan: One of the checks is unnecessary.

cjyuan (commented):
Changes look good.

Sprint-3/alarmclock/alarmclock.js
Grajales-K: I removed the rawValue === "" check since inputTime <= 0 already handles empty inputs. Thank you!

illicitonion:

PR #1188, opened 2026-04-08
London | 26-ATP-Jan | Boualem Larbi Djebbour | Sprint 2 | coursework

djebsoft:
## Learners, PR Template

Self checklist

## Changelist

completed sprint-2 exercies in data groups module.

## Questions

### DIFF
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -45,8 +45,9 @@ test("given an object with properties, returns false when passed to contains wit
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+// "length" is perfect for the test because it's real property on the array itself
 test("given invalid parameters (array), returns false or throws an error", () => {
-  expect(contains(["gitName", "age", "position"], "cohort")).toEqual(false);
+  expect(contains(["gitName", "age", "position"], "length")).toBe(false);
 });
 
 test("given invalid parameters (like null or undefined), returns false or throws an error", () => {
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -20,16 +20,31 @@ as the object doesn't contains a key of 'c'
 // Given an empty object
 // When passed to contains
 // Then it should return false
-test.todo("contains on empty object returns false");
+test("given an empty object, returns false", () => {
+  expect(contains({}, "x")).toEqual(false);
+});
 
 // Given an object with properties
 // When passed to contains with an existing property name
 // Then it should return true
+test("given an object with properties, returns true when passed to contains with an existing property name", () => {
+  expect(
+    contains({ gitName: "djebsoft", position: "cyf trainee" }, "gitName")
+  ).toEqual(true);
+});
 
 // Given an object with properties
 // When passed to contains with a non-existent property name
 // Then it should return false
+test("given an object with properties, returns false when passed to contains with a non-existent property name", () => {
+  expect(
+    contains({ gitName: "djebsoft", position: "cyf trainee" }, "age")
+  ).toEqual(false);
+});
 
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+test("given invalid parameters, returns false or throws an error", () => {
+  expect(contains(["gitName", "position"], "gitName")).toEqual(false);
+});
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -45,6 +45,10 @@ test("given an object with properties, returns false when passed to contains wit
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
-test("given invalid parameters, returns false or throws an error", () => {
+test("given invalid parameters (array), returns false or throws an error", () => {
   expect(contains(["gitName", "position"], "gitName")).toEqual(false);
+});
+
+test("given invalid parameters (like null or undefined), returns false or throws an error", () => {
+  expect(contains(null, "gitName")).toEqual(false);
 });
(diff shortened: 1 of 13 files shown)


Sprint-2/implement/contains.test.js
```diff
@@ -20,16 +20,31 @@ as the object doesn't contains a key of 'c'
...
+    contains({ gitName: "djebsoft", position: "cyf trainee" }, "age")
+  ).toEqual(false);
+});
 
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+test("given invalid parameters, returns false or throws an error", () => {
+  expect(contains(["gitName", "position"], "gitName")).toEqual(false);
+});
```
cjyuan, line 51: This test does not yet confirm that the function correctly returns false when the first argument is an array. 
This is because `contains(["gitName", "position"], "gitName")` could also return `false` simply because "gitName" is not a key of the array.

**Arrays are objects, with their indices acting as keys.** A proper test should use a valid

After you fixed this test, make sure you also run the test to check your function.

Note: When testing invalid type of data, `undefined` and `null` are usually good candidates to test -- many functions fail because they could not handle `undefined` or `null`.

Sprint-2/implement/lookup.js
```diff
@@ -1,5 +1,24 @@
-function createLookup() {
-  // implementation here
+// the first element is the country code
+// the second array is the currency code of the country
+// the function should return an object
+// the keys of the object should be the first element of each inner array (country code)
+// the values of the object should be the second element of each inner array (currency code)
+
+function Lookup(coCuPairs) {
```
cjyuan: - In JS naming convention, both function and variable names usually begin with a lowercase letter (i.e. camelCase). Names starting with an uppercase letter are used for built-in and custom data types (e.g., `Math`).

- `countryCurrencyPairs` is probably more readable than `coCuPairs`.

Sprint-2/implement/lookup.js
```diff
@@ -1,5 +1,24 @@
...
+    coCuPairs.every((pair) => Array.isArray(pair)) &&
+    // check if each inner array has two elements.
+    coCuPairs.every((pair) => pair.length === 2)
+    // we can also use index to check the pair is an array as follows:
+    // for (let i = 0; i < coCuPairs.length; i++)
+    //  return Array.isArray(coCuPairs[i])
+  ) {
+    return Object.fromEntries(coCuPairs);
+  }
+  return "Invalid input: expected an array of arrays";
```
cjyuan: Why not throw an error instead so that the caller don't have to check the type of the return value before using it?

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,12 @@
...
-
+function tally(arr) {
+  if (!Array.isArray(arr)) {
+    throw new Error("invalid input");
+  }
+  let result = {};
+  for (let i = 0; i <= arr.length - 1; i++) {
+    let item = arr[i];
+    result[item] = (result[item] || 0) + 1;
+  }
```
```
```

Sprint-2/implement/contains.test.js
djebsoft: thank you for the feedback
I just realised that the false returned cases should be the exception, so I changed the if condition to the other way around.
I removed the case of empty object (object length equals to zero) as it does not add any plus to the code. 
I also added the case where the input is an array to return false.
and run the test to check the function.
added a test with null input for invalid type of data.

Sprint-2/implement/lookup.js
djebsoft: thank you for the suggestion
i changed the parameter name to countrycurrencyPairs.

Sprint-2/implement/lookup.js
djebsoft: thanks for the review
I changed the if condition to the other way around.
I changed the code to throw an error when the data type of the passed argument to the function is invalid.

Sprint-2/implement/tally.js
djebsoft: thank you for this
it actually didn't return the value I expected which is: { toString: 2 }
its part of prototype polution bug 
so, I needed to (and at any time keys come from user input like strings)
to avoid normal {} objects

Sprint-2/implement/contains.test.js
cjyuan: Why not also update the test (the one that tests array)?

Your function is correct, but we write tests not only to verify our current implementation, but also to ensure that future changes do not alter the function's expected behavior.

Sprint-2/implement/lookup.js
cjyuan: What about the function name?

Sprint-2/implement/contains.test.js
djebsoft: thanks for the spot
sorry that I got confused about this point
i think the perfect example test is to test tricky arguments where the first argument looks like an object by having two elements and the second argument is string that match one the elements in the first argument
therefore, if the first argument wasn't an array than the test should return true. but because it an array, the function returns false as invalid input.
could you enlighten me about what I'm mistaken here ?

cjyuan: `expect(contains(["gitName", "position"], "gitName")).toEqual(false);`

The following function (changed in the future by someone else) could also pass the above test:
```
function contains(obj, key) {
}
```

cjyuan: This test is supposed to fail when the function is implemented as:
```javascript
function contains(obj, key) {
}
```

The new test still behaves the same as the previous one.

**Arrays are objects, with their indices acting as keys.** You should use a **valid
key** to ensure the function returns false specifically because the input is an array, not because the key is missing.

illicitonion:

PR #1190, opened 2026-04-10
London | 26-ITP-Jan | Miriam Jorna | Sprint 2 | Data Groups

miriamjorna:
Self checklist

## Changelist

Coursework for Sprint 2. 
My mojo went MIA but I've found it back!

### DIFF
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -20,16 +20,28 @@ as the object doesn't contains a key of 'c'
 // Given an empty object
 // When passed to contains
 // Then it should return false
-test.todo("contains on empty object returns false");
+
+test("contains an empty object returns false", () => {
+  expect(contains({}, "a")).toEqual(false);
+});
 
 // Given an object with properties
 // When passed to contains with an existing property name
 // Then it should return true
+test("given an object with properties, returns true for existing property", () => {
+  expect(contains({a: 1, b: 2}, "a")).toEqual(true);
+});
 
 // Given an object with properties
 // When passed to contains with a non-existent property name
 // Then it should return false
+test("given an object with properties, returns false for non-existent property", () => {
+  expect(contains({a: 1, b: 2}, "c")).toEqual(false);
+});
 
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+test("given invalid parameters like an array, returns false", () => {
+  expect(contains([1, 2, 3], "a")).toEqual(false);
+});
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -43,5 +43,5 @@ test("given an object with properties, returns false for non-existent property",
 // When passed to contains
 // Then it should return false or throw an error
 test("given invalid parameters like an array, returns false", () => {
-  expect(contains([1, 2, 3], "a")).toEqual(false);
+  expect(contains([1, 2, 3], "0")).toEqual(false);
 });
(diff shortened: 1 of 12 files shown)


Sprint-2/implement/contains.test.js
```diff
@@ -20,16 +20,28 @@ as the object doesn't contains a key of 'c'
...
+test("given an object with properties, returns false for non-existent property", () => {
+  expect(contains({a: 1, b: 2}, "c")).toEqual(false);
+});
 
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+test("given invalid parameters like an array, returns false", () => {
+  expect(contains([1, 2, 3], "a")).toEqual(false);
+});
```
cjyuan, line 47: Your function is correct, but we should prepare tests not only to verify our current implementation, but also to ensure that future changes do not alter the function's expected behavior.

This test cannot yet confirm that the function correctly returns false when the first argument is an array. 

**Arrays are objects, with their indices acting as keys.** A proper test should **use a valid key** to ensure the function returns `false` specifically because the input is an array, not because the key is missing.

Sprint-2/implement/querystring.js
```diff
@@ -6,7 +6,9 @@ function parseQueryString(queryString) {
   const keyValuePairs = queryString.split("&");
 
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
+    const index = pair.indexOf("=");
+    const key = pair.slice(0, index);
+    const value = pair.slice(index + 1);
```
cjyuan, line 12: For each of the following function calls, does your function return the value you expect?
Note: These arguments are valid query strings.

```javascript
parseQueryString("key=")                                   
parseQueryString("=value")
```

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,16 @@
...
+    throw new Error("Input must be an array");
+  }
+  const result = {};
+  for (const item of arr) {
+    if (result[item]) {
+      result[item]++;
+    } else {
+      result[item] = 1;
+    }
+  }
```
cjyuan, line 12: Does the following function call returns the value you expect?
```
```

Suggestion:
- use `Object.hasOwn()` to check if `result` contains a specific key/item

Sprint-2/implement/contains.test.js
miriamjorna: Changed "a" into "0" because "0" exists as a key in the array, so if 
contains 
returns false it proves the function is correctly rejecting arrays as invalid input.

Sprint-2/implement/querystring.js
miriamjorna: Hmmm, that gets messy quickly. I've added measures to ignore pairs with no = and empty keys.

Sprint-2/implement/tally.js
miriamjorna: I chose Object.hasOwn()
Thanks again for your time!

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,16 @@
...
+  if (!Array.isArray(arr)) {
+    throw new Error("Input must be an array");
+  }
+  const result = {};
+  for (const item of arr) {
+   if (Object.hasOwn(result, item)) {
+  result[item]++;
+} else {
+  result[item] = 1;
+}
```

Have you installed the **prettier** VSCode extension and **enabled "Format on save/paste"** on VSCode,
as recommended in 
?

If you have successfully setup VSCode, you won't need to worry about manual indentation or trailing space.

cjyuan (commented):
Other changes are good.

Sprint-2/implement/tally.js
miriamjorna: Thank you, this one is now also Prettified :)

cjyuan:
All good now. Well done.

illicitonion:

PR #1192, opened 2026-04-11
London | 26-ITP-Jan | Miriam Jorna | Sprint 3 | Alarmclock

miriamjorna:
Self checklist

## Changelist

Made a working alarm clock. 
The page changes colour when the alarm goes off.
I've made a colour reset provision for when the alarm is paused (=what the user perceives as turning it off).
Noticed input could be negative; have made sure only positive integers can now be used to set the time.


Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,30 @@
-function setAlarm() {}
+function setAlarm() {
+  let seconds = parseInt(document.getElementById("alarmSet").value);
+  
+  if (seconds <= 0) {
+  alert("The number of seconds must be more than 0 please");
+  return;
+}
```
cjyuan: Indentation is off.

as recommended in 
?

---

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,30 @@
-function setAlarm() {}
+function setAlarm() {
+  let seconds = parseInt(document.getElementById("alarmSet").value);
+  
+  if (seconds <= 0) {
```
cjyuan: Some invalid input can make the app behave abnormally.

Note: Why requiring min to be `0` here but set the "min" attribute in HTML to 1?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,30 @@
...
+    
+    if (seconds <= 0) {
+      clearInterval(countdown);
+      playAlarm();
+      document.body.style.backgroundColor = "darkorange";
+    }
+  }, 1000);
+}
 
 // DO NOT EDIT BELOW HERE
```
cjyuan, line 45: To respect the `// DO NOT EDIT BELOW HERE` marker, we could reset the background in a separate callback as:

```
window.addEventListener("load", function() {
    // code to reset background
    ...
  });
});
```

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,30 @@
...
+  updateDisplay(); // update immediately on click
+
+  const countdown = setInterval(() => {
+    seconds--;
+    updateDisplay();
+    
+    if (seconds <= 0) {
+      clearInterval(countdown);
+      playAlarm();
+      document.body.style.backgroundColor = "darkorange";
```
```
```

cjyuan (commented):
Currently, when starting a new countdown, the application does not always return to a clean initial state, which can lead to inconsistent behavior between runs.

Consider introducing a dedicated reset function to return the app to a clean initial state, and call this function before starting a new countdown to help ensure consistency.

Sprint-3/alarmclock/alarmclock.js
miriamjorna: Oh, my apologies, I inserted that and did not prettify it after doing so. Tbh I would normally have done that by hand quickly. Just forgot.

Sprint-3/alarmclock/alarmclock.js
miriamjorna: Ah - this would then also remove the need to do the above recommendation on resetting the background colour, did I see that correctly? (the bit that is currently not respecting the  // DO NOT EDIT BELOW HERE marker)

cjyuan: You can still change background color, just specify a default background color for the body element in `style.css`, and then define a class, says `.alarm-activated` with a different background color.

By adding and removing the class from body, you can then switch between two background colors.

Note: What you want to do (besides stopping the alarm) when the user clicks the "Stop" button is up to you.

Sprint-3/alarmclock/alarmclock.js
miriamjorna: To be honest I'm a bit puzzled by your note. 
When I noticed that the little arrows on the input box could give negative values I changed the requirements to be >0 by putting in this 
   if (seconds <= 0) 
     alert("The number of seconds must be more than 0 please");
and making sure the input required integers.

It has to be minimal 1 because otherwise the clock can't start. 
What am I missing that makes you state the min is 0 here? Because when it is 0 the warning comes up?

miriamjorna: Oh wait, a different notation would work for both things you said.  
Have changed it to 
   (!seconds || seconds < 1)

Sprint-3/alarmclock/alarmclock.js
miriamjorna: Many thanks. All sorted, I believe. Was a bit of a rollercoaster in my perception, from easy to rather complicated to easy to... etc. But got there in the end :)

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,46 @@
...
+  if (!seconds || seconds < 1) {
+    alert("The number of seconds must be more than 0 please");
+    return;
+  }
+  function updateDisplay() {
+    const minutes = Math.floor(seconds / 60);
+    const remainingSeconds = seconds % 60;
+    document.getElementById("timeRemaining").textContent =
+      `Time Remaining: ${String(minutes).padStart(2, "0")}:${String(remainingSeconds).padStart(2, "0")}`;
+  }
```
cjyuan: Would be better to design the function to take `seconds` through a parameter, and move the function to the outer scope.
 
This way, you can call `updateDisplay(0)` on line 7 -- keep the display logic in one place.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,46 @@
-function setAlarm() {}
+let countdown;
+
+// reset before starting new countdown
+function resetAlarm() {
+  clearInterval(countdown);
+  audio.pause();
+  document.getElementById("timeRemaining").textContent = "Time Remaining: 00:00";
+  document.body.classList.toggle("alarm-activated", false);
+}
```
cjyuan, line 18: This function is never called.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,46 @@
...
+      `Time Remaining: ${String(minutes).padStart(2, "0")}:${String(remainingSeconds).padStart(2, "0")}`;
+  }
+
+  updateDisplay(); // update immediately on click
+
+  // code to reset background
+    function pauseAlarm() {
+      audio.pause();
+      document.body.classList.toggle("alarm-activated", false);
+    }
```
cjyuan: This function is also never called.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,46 @@
-function setAlarm() {}
+let countdown;
```
cjyuan: Consider renaming `countdown` to `countdownInterval` or `intervalId` since this stores the interval ID, not the countdown value.

Sprint-3/alarmclock/alarmclock.js
miriamjorna: Apologies. Due to time constraints I gave up on the idea of making a pause button, but I didn't clean up my code accordingly

Sprint-3/alarmclock/alarmclock.js
miriamjorna: Apologies. Due to time constraints I gave up on the idea of making a pause button, but I didn't clean up my code accordingly

Sprint-3/alarmclock/alarmclock.js
miriamjorna: Check. Thank you.

Sprint-3/alarmclock/alarmclock.js
miriamjorna: Have refactored this. Looks nice :)

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -18,8 +57,4 @@ function playAlarm() {
   audio.play();
 }
 
-function pauseAlarm() {
-  audio.pause();
-}
-
```
cjyuan, line 24: Why delete `pauseAlarm()`?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,43 @@
...
+    document.getElementById("timeRemaining").textContent =
+      `Time Remaining: ${String(minutes).padStart(2, "0")}:${String(remainingSeconds).padStart(2, "0")}`;
+  }
+
+// reset before starting new countdown
+function resetAlarm() {
+  clearInterval(countdownId);
+  updateDisplay(0);  // replaces the manual textContent line
+  document.body.classList.toggle("alarm-activated", false);
+}
```
cjyuan, line 18: Shouldn't this function, `resetAlarm()` be called in `setAlarm()`?

You should also reset the audio because the user may not click the Stop button before starting a new countdown.

cjyuan (commented):
Indentation is a bit off. Manual indentation is a good habit but enabling auto-format can save time and help clean up trailing space. The latter is easy to miss when we format code manually.

Sprint-3/alarmclock/alarmclock.js
miriamjorna: Oh no! Have I been careless cleaning up the pause button issue? Not the first time I've confused the script term "pause alarm" vs. the human term "stop alarm", apologies!

Sprint-3/alarmclock/alarmclock.js
miriamjorna: Implemented. Thank you for your time!

Sprint-3/alarmclock/alarmclock.js
cjyuan: We can double check what we have changed in the PR by clicking the "Files changed" tab.

image

cjyuan:
Changes look good.

illicitonion:

PR #1193, opened 2026-04-11
Cape Town | 2026-ITP-Jan | Pretty Taruvinga | Sprint 3 | Alarm clock

Pretty548:
## Learners, PR Template

Self checklist

## Changelist

Fixed my code and removed whats not supposed to be there.

### DIFF
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -23,6 +23,7 @@ function setAlarm() {
 
   if (timer) {
     clearInterval(timer);
+    timer = null;
   }
 
   updateDisplay(timeLeft);
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -26,6 +26,8 @@ function setAlarm() {
     timer = null;
   }
 
+  stopAlarm();
+
   updateDisplay(timeLeft);
 
   timer = setInterval(() => {
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -18,7 +18,6 @@ function setAlarm() {
   timeLeft = parseInt(input, 10);
 
   if (isNaN(timeLeft) || timeLeft <= 0) {
-    alert("Please enter a valid number of seconds.");
     return;
   }
 
@@ -29,13 +28,14 @@ function setAlarm() {
   updateDisplay(timeLeft);
 
   timer = setInterval(() => {
-    if (timeLeft > 0) {
-      timeLeft--;
-      updateDisplay(timeLeft);
-    } else {
+    if (timeLeft <= 0) {
       clearInterval(timer);
       playAlarm();
+      return;
     }
+
+    timeLeft--;
+    updateDisplay(timeLeft);
   }, 1000);
 }
 
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -1,4 +1,3 @@
-let timeLeft = 0;
 let timer = null;
 
 function formatTime(seconds) {
@@ -13,6 +12,7 @@ function updateDisplay(timeLeft) {
 }
 
 function setAlarm() {
+  let timeLeft = 0;
   const input = document.getElementById("alarmSet").value;
 
   timeLeft = parseInt(input, 10);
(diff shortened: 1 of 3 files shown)


Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,38 @@
...
+  const display = document.getElementById("timeRemaining");
+
+  timeLeft = parseInt(input, 10);
+
+  if (isNaN(timeLeft) || timeLeft <= 0) {
+    alert("Please enter a valid number of seconds.");
+    return;
+  }
+
+  display.textContent = `Time Remaining: ${formatTime(timeLeft)}`;
```
cjyuan: Could consider implementing a function to display time even if it contains only one statement. Doing so can keep the logic of displaying time in one place.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,38 @@
-function setAlarm() {}
+let timeLeft = 0;
```
cjyuan: Could declare `timeLeft` inside `setAlarm()` (since it is only used within that function).

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,38 @@
...
+  if (isNaN(timeLeft) || timeLeft <= 0) {
+    alert("Please enter a valid number of seconds.");
+    return;
+  }
+
+  display.textContent = `Time Remaining: ${formatTime(timeLeft)}`;
+
+  if (timer) {
+    clearInterval(timer);
+  }
```
cjyuan, line 27: What elsse should also be reset before a new countdown begin?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,38 @@
...
+  timer = setInterval(() => {
+    if (timeLeft > 0) {
+      timeLeft--;
+      display.textContent = `Time Remaining: ${formatTime(timeLeft)}`;
+    } else {
+      clearInterval(timer);
+      playAlarm();
+    }
+  }, 1000);
+}
```
cjyuan, line 43: When the countdown reaches 00:00, there is a one second delay before the alarm sound is played. Is this by design?

cjyuan (commented):

Sprint-3/alarmclock/alarmclock.js
Pretty548: Refactored display logic into a reusable function and fixed timer behaviour to ensure correct initial rendering and countdown timing.

Sprint-3/alarmclock/alarmclock.js
Pretty548: Thanks for the suggestion! I’ve moved timeLeft inside setAlarm to limit its scope and keep the global space cleaner.

Sprint-3/alarmclock/alarmclock.js
Pretty548: Good point  I’ve reset the timer reference to null after clearing it to ensure a clean state before starting a new countdown.

Sprint-3/alarmclock/alarmclock.js
Pretty548: Good catch — I’ve updated the logic so the alarm triggers immediately when the countdown reaches 00:00, removing the delay

Sprint-3/alarmclock/alarmclock.js
cjyuan: It's still declared in the outermost scope.

Sprint-3/alarmclock/alarmclock.js
cjyuan: - Setting timer to `null` is unnecessary in JS but it is a good practice.

- What does the "Stop" button do? What would happen when a countdown reaches zero, and the user immediately set another countdown?

Sprint-3/alarmclock/alarmclock.js
Pretty548: Good catch — I had left the global timeLeft in by mistake. I’ve removed it and kept it scoped inside setAlarm() 👍

Sprint-3/alarmclock/alarmclock.js
Pretty548: The Stop button clears the timer and stops the alarm. If a countdown reaches zero and the user immediately starts another one, the alarm may still be playing. So we should also reset/stop the alarm when starting a new countdown to keep the behavior consistent.

cjyuan (commented):
Changes look good. Well done.

illicitonion:

PR #1200, opened 2026-04-15
London | 26-ATP-Jan | Boualem Larbi Djebbour | Sprint 3 | Quote generator

djebsoft:
## Learners, PR Template

Self checklist

## Changelist
creating a quote generator application

### DIFF
--- a/Sprint-3/quote-generator/index.html
+++ b/Sprint-3/quote-generator/index.html
@@ -7,7 +7,7 @@
     <script defer src="quotes.js"></script>
     <link rel="stylesheet" href="style.css" />
   </head>
-  <body onload="setup()">
+  <body>
     <div class="container">
       <p id="quote"></p>
       <p id="author"></p>
--- a/Sprint-3/quote-generator/index.html
+++ b/Sprint-3/quote-generator/index.html
@@ -8,7 +8,7 @@
     <link rel="stylesheet" href="style.css" />
   </head>
   <body>
-    <h1>hello there</h1>
+    <h1>Welcome to the Quote Generator</h1>
     <p id="quote"></p>
     <p id="author"></p>
     <button type="button" id="new-quote">New quote</button>
--- a/Sprint-3/quote-generator/index.html
+++ b/Sprint-3/quote-generator/index.html
@@ -1,10 +1,11 @@
-<!DOCTYPE html>
+<!doctype html>
 <html lang="en">
   <head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <title>Title here</title>
+    <title>Quote generator app</title>
     <script defer src="quotes.js"></script>
+    <link rel="stylesheet" href="style.css" />
   </head>
   <body>
     <h1>hello there</h1>
--- a/Sprint-3/quote-generator/index.html
+++ b/Sprint-3/quote-generator/index.html
@@ -7,10 +7,14 @@
     <script defer src="quotes.js"></script>
     <link rel="stylesheet" href="style.css" />
   </head>
-  <body>
+  <body onload="main()">
     <h1>Welcome to the Quote Generator</h1>
-    <p id="quote"></p>
-    <p id="author"></p>
-    <button type="button" id="new-quote">New quote</button>
+    <div class="container">
+      <p id="quote"></p>
+      <p id="author"></p>
+      <div id="new-quote">
+        <button type="button">New quote</button>
+      </div>
+    </div>
   </body>
 </html>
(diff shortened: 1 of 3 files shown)


Sprint-3/quote-generator/quotes.js
```diff
@@ -1,3 +1,10 @@
+function main() {
+  document.querySelector("#quote").innerText = pickFromArray(quotes).quote;
+  document.querySelector("#author").innerText = pickFromArray(quotes).author;
+}
```
cjyuan: The quote and the author will not match because they are from two randomly selected objects.

Sprint-3/quote-generator/quotes.js
```diff
@@ -1,3 +1,10 @@
...
+
+document.getElementById("new-quote").addEventListener("click", main);
+
```
cjyuan: Best practice is to place all the "run on load" code in one place, ideally in a function,
to make it clear that "this is what runs when the page loads."
Then, add the function on page load as:

```javascript
function setup() {
  // 1. show initial quote
  // 2. setup on click event listener
}

```
or
```javascript
  // 1. show initial quote
  // 2. setup on click event listener
});
```

Sprint-3/quote-generator/quotes.js
```diff
@@ -1,3 +1,10 @@
+function main() {
```
cjyuan: Can consider rename this function to better match the task it is doing.

Sprint-3/quote-generator/index.html
```diff
@@ -1,15 +1,19 @@
...
+    <title>Quote generator app</title>
     <script defer src="quotes.js"></script>
+    <link rel="stylesheet" href="style.css" />
   </head>
-  <body>
-    <h1>hello there</h1>
-    <p id="quote"></p>
-    <p id="author"></p>
-    <button type="button" id="new-quote">New quote</button>
+  <body onload="main()">
```
cjyuan: Better to separate JS code from HTML, and setup the callback in JS code using `addEventListener()`.

Sprint-3/quote-generator/quotes.js
djebsoft: thank for spotting the bug,
I assigned the return value of the function pickFromArray to randomQuote variable to keep the same object for quote and author to match and avoid the random method running twice

Sprint-3/quote-generator/quotes.js
```diff
@@ -1,3 +1,10 @@
+function setup() {
+  const randomQuote = pickFromArray(quotes);
+  document.querySelector("#quote").innerText = randomQuote.quote;
+  document.querySelector("#author").innerText = randomQuote.author;
+  document.getElementById("new-quote").addEventListener("click", setup);
```
cjyuan: Line 5 should only be executed once on page load instead of being called every time when the button is clicked.

Sprint-3/quote-generator/quotes.js
```diff
@@ -1,3 +1,13 @@
+function setup() {
+  const randomQuote = pickFromArray(quotes);
+  document.querySelector("#quote").innerText = randomQuote.quote;
+  document.querySelector("#author").innerText = randomQuote.author;
+}
```
cjyuan: The function name does not quite match the task being performed.

Sprint-3/quote-generator/quotes.js
```diff
@@ -1,3 +1,13 @@
...
+function newQuote() {
+  document.getElementById("new-quote").addEventListener("click", setup);
+}
+newQuote();
```
cjyuan: Statements on lines 7 and 9 are the run-on-load tasks. They are the operations "setup()" should be performing.

Sprint-3/quote-generator/quotes.js
```diff
@@ -1,3 +1,13 @@
+function randomQuote() {
+  const randomQuote = pickFromArray(quotes);
+  document.querySelector("#quote").innerText = randomQuote.quote;
+  document.querySelector("#author").innerText = randomQuote.author;
+}
+function setup() {
+  document.getElementById("new-quote").addEventListener("click", randomQuote);
+}
+setup();
+window.addEventListener("load", randomQuote);
```
cjyuan: You can also consider write:
```javascript
 function setup() {
  // All tasks to be performed once on page load
  document.getElementById("new-quote").addEventListener("click", randomQuote);
  randomQuote();
}

// Setup "setup()" to be called on page load
window.addEventListener("load", setup);
```

cjyuan (commented):
All good.

Sprint-3/quote-generator/quotes.js
djebsoft: done, thanks for the clarification

illicitonion:

PR #1204, opened 2026-04-18
Glasgow| 2026-ITP-Jan | Tuan Nguyen| Sprint 2 | Objects

Jacknguyen4438:
## Learners, PR Template

Self checklist

## Changelist

For this PR I have finish all the required section need for PR submission.

## Questions

No question.

### DIFF
--- a/Sprint-2/debug/author.js
+++ b/Sprint-2/debug/author.js
@@ -11,6 +11,6 @@ const author = {
   alive: true,
 };
 
-for (const value of author) {
+for (const value in author) {
   console.log(value);
 }
--- a/Sprint-2/debug/author.js
+++ b/Sprint-2/debug/author.js
@@ -2,7 +2,7 @@
 
 // This program attempts to log out all the property values in the object.
 // But it isn't working. Explain why first and then fix the problem
-
+// this function have 2 bugs, one is that the since object is not iterable so by using for of loop is will not work. The second is that is only print out property name not property value
 const author = {
   firstName: "Zadie",
   lastName: "Smith",
@@ -12,5 +12,5 @@ const author = {
 };
 
 for (const value in author) {
-  console.log(value);
+  console.log(author[value]);
 }
--- a/Sprint-2/debug/recipe.js
+++ b/Sprint-2/debug/recipe.js
@@ -12,4 +12,4 @@ const recipe = {
 
 console.log(`${recipe.title} serves ${recipe.serves}
   ingredients:
-${recipe}`);
+${recipe.ingredients}`);
--- a/Sprint-2/debug/recipe.js
+++ b/Sprint-2/debug/recipe.js
@@ -12,4 +12,4 @@ const recipe = {
 
 console.log(`${recipe.title} serves ${recipe.serves}
   ingredients:
-${recipe.ingredients}`);
+${recipe.ingredients.join("\n")}`);
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -3,18 +3,18 @@ function contains(object, key) {
     return false;
   }
 
-  for (const checkItem in object){
-    let validation = object[checkItem]
-     if(Array.isArray(validation)){
+  for (const checkItem in object) {
+    let validation = object[checkItem];
+    if (Array.isArray(validation)) {
       return false;
-     }
+    }
   }
   for (const item in object) {
     if (item === key) {
       return true;
     }
   }
-   return false;
+  return false;
 }
 
 module.exports = contains;
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,19 +1,29 @@
 function contains(object, key) {
+  if (
+    object === null ||
+    object === undefined ||
+    typeof object !== "object" ||
+    Array.isArray(object)
+  ) {
+    return false;
+  }
+
   if (Object.keys(object).length === 0) {
     return false;
   }
 
-  for (const checkItem in object) {
-    let validation = object[checkItem];
-    if (Array.isArray(validation)) {
+  for (const item in object) {
+    if (Array.isArray(object[item])) {
       return false;
     }
   }
+
   for (const item in object) {
     if (item === key) {
       return true;
     }
   }
+
   return false;
 }
 
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,10 +1,5 @@
 function contains(object, key) {
-  if (
-    object === null ||
-    object === undefined ||
-    typeof object !== "object" ||
-    Array.isArray(object)
-  ) {
+  if (object === null || typeof object !== "object" || Array.isArray(object)) {
     return false;
   }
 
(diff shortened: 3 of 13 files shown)


Sprint-2/debug/author.js
```diff
@@ -11,6 +11,6 @@ const author = {
   alive: true,
 };
 
-for (const value of author) {
+for (const value in author) {
   console.log(value);
 }
```
cjyuan, line 16: Does this output the property names or property values ?

Sprint-2/debug/recipe.js
```diff
@@ -12,4 +12,4 @@ const recipe = {
 
 console.log(`${recipe.title} serves ${recipe.serves}
   ingredients:
-${recipe}`);
+${recipe.ingredients}`);
```
cjyuan: This does not quite output each ingredient on a new line?

Sprint-2/implement/contains.test.js
```diff
@@ -16,20 +16,42 @@ as the object doesn't contains a key of 'c'
...
+  expect(contains(goodsObject,"color")).toBe(false);
+  expect(contains(goodsObject,"country")).toBe(false);
+});
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+test("should return false if the input object do not have non existing key value", () => {
+  let mixObject = {a:2,b:"hello",c:[]};
+ expect(contains(mixObject,"c")).toBe(false);
+});
```
cjyuan: I think the description means checking if the first argument is an array or a type that's not object. That is, checking  

```javascript
expect(contains(["A", "B"], "1" )).toBe(false);
expect(contains(null, "1" )).toBe(false);
expect(contains(undefined, "1" )).toBe(false);
expect(contains("ABC", "1" )).toBe(false);
```

Can you check also if your function can pass these tests?

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,17 @@
...
+    throw new Error("Input musk be an array")
+  }
+  let totalSum = {};
+  for (const item of sumArray) {
+    if (totalSum[item] === undefined) {
+      totalSum[item] = 1;
+    } else {
+      totalSum[item] += 1;
+    }
+  }
```
cjyuan, line 15: Does the following function call returns the value you expect?
```
```

Suggestion:
- use `Object.hasOwn()`

cjyuan (commented):
Some of the code is not consistently formatted.

Have you installed the **prettier** VSCode extension and enabled "Format on save/paste" on VSCode, as recommended in 
?

Sprint-2/debug/author.js
Jacknguyen4438: Thank you for the feed back I see I have miss fixing this , the answer is is print out property name not property value

Sprint-2/debug/recipe.js
Jacknguyen4438: Thank you for the feed back I have making change and this is my solution:
  ingredients:
${recipe.ingredients.join("\n")}`);

Sprint-2/implement/contains.test.js
Jacknguyen4438: Thank you for the feed back here is my solution base on your feed back:

function contains(object, key) {
  if (
    object === null ||
    object === undefined ||
    typeof object !== "object" ||
    Array.isArray(object)
  ) {
    return false;
  }

  if (Object.keys(object).length === 0) {
    return false;
  }

  for (const item in object) {
    if (Array.isArray(object[item])) {
      return false;
    }
  }

  for (const item in object) {
    if (item === key) {
      return true;
    }
  }

  return false;
}

Sprint-2/implement/tally.js
Jacknguyen4438: Thank you for the feed back here is my solution:

function tally(sumArray) {
  if (sumArray.length === 0) return {};
  if (!Array.isArray(sumArray)) {
    throw new Error("Input musk be an array");
  }
  let totalSum = {};
  for (const item of sumArray) {
    if (!Object.hasOwn(totalSum, item)) {
      totalSum[item] = 1;
    } else {
      totalSum[item] += 1;
    }
  }
  return totalSum;
}

Jacknguyen4438:
@cjyuan Hello and thank you for the feed back I have make some change base on your feed back the brand is ready for your review again. If you read this and see the tag not a course work, sorry for the mis leading tags due to the PR validation system keep reject my request for re-review by remove the need review tags

Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,30 @@
-function contains() {}
+function contains(object, key) {
+  if (
+    object === null ||
+    object === undefined ||
+    typeof object !== "object" ||
+    Array.isArray(object)
```
cjyuan: One of these checks is unnecessary, can you delete it?

Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,30 @@
...
+
+  if (Object.keys(object).length === 0) {
+    return false;
+  }
+
+  for (const item in object) {
+    if (Array.isArray(object[item])) {
+      return false;
+    }
+  }
```
cjyuan: The function is supposed to check only if `object` is an object that is not an array, and whether the object has a property with the specified name. That is, the function should not need to check the property values.

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,16 @@
-function tally() {}
-
+function tally(sumArray) {
+  if (sumArray.length === 0) return {};
+  if (!Array.isArray(sumArray)) {
+    throw new Error("Input musk be an array");
+  }
```
cjyuan: If `sumArray` is `null` or `undefined`, line 2 will cause an error to be thrown.

cjyuan (commented):
Please revert the change to `Sprint-3/alarmclock/index.html` on this branch to keep it clean.

Sprint-2/implement/contains.js
Jacknguyen4438: Than you for the feed back , I have remove the object  === undefined since other 3 validation text can handle just fine

Sprint-2/implement/contains.js
Jacknguyen4438: THank you for your feed back. I have make some change  to make sure the function operate what is suppose to do:

function contains(object, key) {
  if (object === null ||typeof object !== "object"  || Array.isArray(object)) {
    return false;
  }

  if (Object.keys(object).length === 0) {
    return false;
  }

  for (const item in object) {
    if (item === key) {
      return true;
    }
  }
   return false;
}

Sprint-2/implement/tally.js
Jacknguyen4438: Thank you for the feed back here is my solution:

function tally(sumArray) {
  if (!Array.isArray(sumArray)) {
    throw new Error("Input musk be an array");
  }
  
  if (sumArray.length === 0) return {};

  let totalSum = {};
  for (const item of sumArray) {
    if (!Object.hasOwn(totalSum, item)) {
      totalSum[item] = 1;
    } else {
      totalSum[item] += 1;
    }
  }
  return totalSum;
}

Jacknguyen4438:
Hello @cjyuan I have make change and ready to be review again

cjyuan:
Looks good.

illicitonion:

PR #1209, opened 2026-04-21
Glasgow| 2026-ITP-Jan | Tuan Nguyen| Sprint 3 |  Alarm clock

Jacknguyen4438:
## Learners, PR Template

Self checklist

## Changelist

Hello I have the function alarm clock and is working correctly. If additional change need to be done I will fix it right away.

## Questions

If this PR have been done mark as completed, can I come back and do the additional task later?

### DIFF
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -16,26 +16,21 @@ function setAlarm() {
   const input = document.getElementById("alarmSet");
   let time = Number(input.value);
 
-  // Always reset first so old countdown stops
   resetAlarm();
 
-  // Validate input (reject decimals, negatives, NaN)
   if (isNaN(time) || time < 0 || !Number.isInteger(time)) {
     showError("Please enter a whole number of seconds.");
     return;
   }
 
-  // If time is 0, trigger alarm immediately
   if (time === 0) {
     updateHeading(0);
     playAlarm();
     return;
   }
 
-  // Show starting time
   updateHeading(time);
 
-  // Start countdown
   countdown = setInterval(() => {
     time--;
 
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -1,31 +1,47 @@
 let countdown = null;
 
+function resetAlarm() {
+  clearInterval(countdown);
+  countdown = null;
+  pauseAlarm();
+  updateHeading(0);
+}
+
 function setAlarm() {
   const input = document.getElementById("alarmSet");
   let time = Number(input.value);
 
-  const heading = document.getElementById("timeRemaining");
-
-  if (countdown) {
-    clearInterval(countdown);
+  if (isNaN(time) || time < 0) {
+    updateHeading(0);
+    return;
   }
 
-  updateHeading(time, heading);
+  resetAlarm();
+
+  if (time === 0) {
+    updateHeading(0);
+    playAlarm();
+    return;
+  }
+
+  updateHeading(time);
 
   countdown = setInterval(() => {
     time--;
 
     if (time <= 0) {
-      updateHeading(0, heading);
+      updateHeading(0);
       clearInterval(countdown);
       playAlarm();
     } else {
-      updateHeading(time, heading);
+      updateHeading(time);
     }
   }, 1000);
 }
 
-function updateHeading(time, heading) {
+function updateHeading(time) {
+  const heading = document.getElementById("timeRemaining");
+
   const minutes = Math.floor(time / 60);
   const seconds = time % 60;
 
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -4,11 +4,10 @@ function resetAlarm() {
   clearInterval(countdown);
   countdown = null;
   pauseAlarm();
-  updateHeading(0);
 }
 
 function showError(message) {
-  const heading = document.getElementById("timeRemaining");
+  const heading = document.getElementById("errorMessage");
   heading.innerText = message;
 }
 
@@ -22,6 +21,8 @@ function setAlarm() {
     showError("Please enter a whole number of seconds.");
     return;
   }
+
+  document.getElementById("errorMessage").innerText = "";
 
   if (time === 0) {
     updateHeading(0);
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -7,25 +7,35 @@ function resetAlarm() {
   updateHeading(0);
 }
 
+function showError(message) {
+  const heading = document.getElementById("timeRemaining");
+  heading.innerText = message;
+}
+
 function setAlarm() {
   const input = document.getElementById("alarmSet");
   let time = Number(input.value);
 
+  // Always reset first so old countdown stops
+  resetAlarm();
+
+  // Validate input (reject decimals, negatives, NaN)
   if (isNaN(time) || time < 0 || !Number.isInteger(time)) {
-    updateHeading(0);
+    showError("Please enter a whole number of seconds.");
     return;
   }
 
-  resetAlarm();
-
+  // If time is 0, trigger alarm immediately
   if (time === 0) {
     updateHeading(0);
     playAlarm();
     return;
   }
 
+  // Show starting time
   updateHeading(time);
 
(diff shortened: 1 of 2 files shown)


Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,5 +1,39 @@
-function setAlarm() {}
+let countdown = null;
 
+function setAlarm() {
+  const input = document.getElementById("alarmSet");
+  let time = Number(input.value);
+
+  const heading = document.getElementById("timeRemaining");
```
cjyuan: `heading` is not used directly in this function. Can you explain why you choose to declare it here and passing it to `updateHeading()` on every function call? Why not just declare it in `updateHeading()` -- the function that actually uses it?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,5 +1,39 @@
-function setAlarm() {}
+let countdown = null;
 
+function setAlarm() {
+  const input = document.getElementById("alarmSet");
+  let time = Number(input.value);
```
cjyuan, line 16: What type and range of number should `time` be? Invalid user input could cause the app to behave abnormally.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,5 +1,39 @@
...
+
+  const heading = document.getElementById("timeRemaining");
+
+  if (countdown) {
+    clearInterval(countdown);
+  }
```

Hint: a user may not click the "Stop" button first before starting a new countdown. 

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,5 +1,39 @@
...
+    time--;
+
+    if (time <= 0) {
+      updateHeading(0, heading);
+      clearInterval(countdown);
+      playAlarm();
+    } else {
+      updateHeading(time, heading);
+    }
+  }, 1000);
```
cjyuan: When input is 1, the alarm is played one second later -- an expected behavior.

However, when input is 0, the alarm is also played one second later (instead of being played immediately).
Can you make the app's behaviour more consistent?

Jacknguyen4438:
Hello @cjyuan, thank you for your review my code is ready to be review again if you need me to max more change please let know.

Sprint-3/alarmclock/alarmclock.js
Jacknguyen4438: Thank you your feedback I was write that way because I try to make the code more easier to read not only to my self but to other reviewer as well here is my solution:

function setAlarm() {
  const input = document.getElementById("alarmSet");
  let time = Number(input.value);


  if (isNaN(time) || time < 0) {
    updateHeading(0);
    return;
  }


  resetAlarm();

  if (time === 0) {
    updateHeading(0);
    playAlarm();
    return;
  }

Sprint-3/alarmclock/alarmclock.js
Jacknguyen4438: Thank you for the feed back I have add validation cod to help with this here is my validation code below:

if (isNaN(time) || time < 0) {
    updateHeading(0);
    return;
  }

Sprint-3/alarmclock/alarmclock.js
Jacknguyen4438: Thank you for the feed back I see that this way of writing can cause some inconsistent behaviour between each runs so  here is mu solution:

function resetAlarm() {
  clearInterval(countdown);
  countdown = null;
  pauseAlarm();
  updateHeading(0);
}

Sprint-3/alarmclock/alarmclock.js
Jacknguyen4438: Thank you for the feed back. I have make change so that it will give the audio when 0 is put in.

Sprint-3/alarmclock/alarmclock.js
cjyuan: What do you expect the app to behave when the user enters `3.14` as input?

Sprint-3/alarmclock/alarmclock.js
cjyuan: Your original approach has its merit, and the new approach (placing 
`const heading = document.getElementById("timeRemaining");` 
inside `updateHeading()` has its drawback.

Can you figure out the pros and cons between these two approaches?

Jacknguyen4438: Thank you for the feed back. Yes I would like to point out pros and cons of these 2 approach.

for const input = document.getElementById("alarmSet"); 
The pros:
- The DOM lookup happens only once, so it’s more efficient.

- The function is more flexible because it can update any heading passed to it.

- Dependencies are explicit because the heading element is provided as an argument.

Cons:

- Slightly more boilerplate code.

 -I must pass the heading every time I call the function or the heading side in HTML will not work as intent.
 
 For putting the  const input = document.getElementById("alarmSet");  inside the function;
 
The pros:

 - Simpler function signature (updateHeading(time)), easier to read.

- The function is self‑contained and finds what it needs internally.

- Less parameters → cleaner API.

The cons
 
- It performs a DOM lookup every second, which is less efficient.

- It is less flexible because it always updates the same element.

- The dependency on the DOM ID is hidden inside the function.

Sprint-3/alarmclock/alarmclock.js
Jacknguyen4438: Thank you for the feed back here is my explain when decimal is pass down:

If the user enters 3.14, Number() converts it to a floating‑point number.
The countdown will decrease by 1 each second (3.14 → 2.14 → 1.14 → 0.14 → alarm), but the display will show whole seconds because of the formatting.

const minutes = Math.floor(time / 60);
const seconds = time % 60;

This behaviour is technically valid but not intuitive for users, since countdown timers usually expect whole seconds.

A more predictable behaviour would be to either:

reject decimal input as invalid, or

convert it to an integer using Math.floor() or Math.round().

This would make the app’s behaviour clearer and more consistent.

Thank you for the reading and I will some change base on this explanation to make sure the function work properly.

Jacknguyen4438:
Hello @cjyuan and thank you for the feed back I make small change so that the function can correctly handle float number. Also I have answer some of the question you would like me to answer.  I hope to hear from you soon thank you

Sprint-3/alarmclock/alarmclock.js
illicitonion: In general if you're just ignoring user input (as you do if someone puts in a float), it's good to show an error message telling them what's wrong, rather than just ignore it.

Also, if you already had an alarm counting down, then submitted 1.2, it will set to 00:00 but then shortly after start counting down again.

Can you fix up both of these things?

Jacknguyen4438: Hello @illicitonion thank you for your feed back I have understand what your like to ask me to do so here is my validation code that will show on the HTML page:

if (isNaN(time) || time < 0 || !Number.isInteger(time)) {
    showError("Please enter a whole number of seconds.");
    return;
  }

Jacknguyen4438:
@illicitonion Hello and thank you for your time review my PR, I have make change to fit with you feed back if I need make change for the error message please let me know

cjyuan:
Changes look good.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,5 +1,60 @@
...
+function resetAlarm() {
+  clearInterval(countdown);
+  countdown = null;
+  pauseAlarm();
+  updateHeading(0);
+}
+
+function showError(message) {
+  const heading = document.getElementById("timeRemaining");
+  heading.innerText = message;
```
cjyuan: Would be better to introducing a separate element to showing message.

Jacknguyen4438: @cjyuan thank you for you reply I have making change that now will display the error without interfering with the heading element in HTML page. Here is my solution:

function showError(message) {
  const heading = document.getElementById("errorMessage");
  heading.innerText = message;
}

function setAlarm() {
  const input = document.getElementById("alarmSet");
  let time = Number(input.value);

  resetAlarm();

  if (isNaN(time) || time < 0 || !Number.isInteger(time)) {
    showError("Please enter a whole number of seconds.");
    return;
  }

  document.getElementById("errorMessage").innerText = "";
  
  for better checking I have make an commit with full code so you can check for better view of the whole code. Thank you.

illicitonion:

PR #1213, opened 2026-05-04
Glasgow| 26-ITP-May| Alasdair MacDonald| Sprint 2 | Exercises

MacDonald91:
## Learners, PR Template

Self checklist

## Changelist

resubmission
## Questions
n/aq

### DIFF
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -1,35 +1,31 @@
 const contains = require("./contains.js");
 
-/*
-Implement a function called contains that checks an object contains a
-particular property
-
-E.g. contains({a: 1, b: 2}, 'a') // returns true
-as the object contains a key of 'a'
-
-E.g. contains({a: 1, b: 2}, 'c') // returns false
-as the object doesn't contains a key of 'c'
-*/
-
-// Acceptance criteria:
-
-// Given a contains function
-// When passed an object and a property name
-// Then it should return true if the object contains the property, false otherwise
-
 // Given an empty object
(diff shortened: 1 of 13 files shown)


cjyuan (commented):
To provide fuller context for this code review, can you include the link to your previous reviewed PR in the PR description?

Sprint-2/implement/contains.test.js
```diff
@@ -1,35 +1,31 @@
...
+  expect(contains(obj, "c")).toBe(false);
+});
 
 // Given invalid parameters like an array
 // When passed to contains
-// Then it should return false or throw an error
+// Then it should return false
+test("returns false when passed invalid parameters like an array", () => {
+  expect(contains([1, 2, 3], "a")).toBe(false);
+});
```
cjyuan, line 31: Your function implementation is correct, but this test is not properly prepared and may fail to detect error when the implementation is modified in the future.

This test cannot yet confirm that the function correctly returns `false` when the first argument is an array. 

Arrays are objects, with their **indices acting as keys**. A proper test should use a valid

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,23 @@
...
+
+  for (const item of items) {
+
+    if (result[item]) {
+      result[item] += 1;
+    } else {
+      result[item] = 1;
+    }
+
+  }
```
cjyuan, line 17: Does the following function call returns the value you expect?
```
```

Suggestion:
- use `Object.hasOwn()`

PR #1215, opened 2026-05-07
 Glasgow| 2026-ITP-May | Tuan Nguyen| Sprint 1 | Array-and-Loop

Jacknguyen4438:
## Learners, PR Template

Self checklist

## Changelist

Hello and thank you for checking this PR. I have done this course before hand and it ready to be review if additional change need to be fix or add please let know thank.

## Questions

No Question.

### DIFF
--- a/Sprint-1/fix/median.js
+++ b/Sprint-1/fix/median.js
@@ -13,6 +13,7 @@ function calculateMedian(list) {
 
   // sorted copy for correct median calculation
   const sorted = [...list].sort((a, b) => a - b);
+  
 
   let correctMedian;
   if (sorted.length % 2 === 0) {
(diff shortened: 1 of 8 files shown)


Sprint-1/fix/median.js
```diff
@@ -6,9 +6,26 @@
...
-  const middleIndex = Math.floor(list.length / 2);
-  const median = list.splice(middleIndex, 1)[0];
-  return median;
+  // validation
+  if (!Array.isArray(list)) return null;
+  if (list.length === 0) return null;
+
+  // sorted copy for correct median calculation
+  const sorted = [...list]
+    .filter((items) => typeof items === "number" && !isNaN(items))
```
illicitonion, line 15: `filter` itself actually returns a copy, so you could skip the `[...list]` part if you wanted

Sprint-1/fix/median.js
```diff
@@ -6,9 +6,26 @@
...
+    .sort((a, b) => a - b);
+
+  if (sorted.length === 0) return null;
+
+  let correctMedian;
```
illicitonion, line 20: `median` feels like a more natural name for this variable than `correctMedian` - what do you think?

illicitonion (commented):
I will mark this as complete, but please think about the two comments

PR #1219, opened 2026-05-07
London | 26-ITP-May | Raihan Sharif | Sprint 1 | Exercises

RaihanSharif:
## Learners, PR Template

Self checklist

## Changelist
Completed exercises as per specs.

### DIFF
--- a/Sprint-1/implement/dedupe.test.js
+++ b/Sprint-1/implement/dedupe.test.js
@@ -16,12 +16,20 @@ E.g. dedupe([1, 2, 1]) target output: [1, 2]
 // Given an empty array
 // When passed to the dedupe function
 // Then it should return an empty array
-test.todo("given an empty array, it returns an empty array");
+test("given empty array, returns empty array", () => {
+  expect(dedupe([])).toEqual([]);
+});
 
 // Given an array with no duplicates
 // When passed to the dedupe function
 // Then it should return a copy of the original array
+test("array with no duplicates returns copy of original", () => {
+  expect(dedupe([1, 3, 4])).toEqual([1, 3, 4]);
+});
 
 // Given an array with strings or numbers
 // When passed to the dedupe function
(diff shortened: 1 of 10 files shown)


Sprint-1/implement/dedupe.test.js
```diff
@@ -16,13 +16,27 @@ E.g. dedupe([1, 2, 1]) returns [1, 2]
...
+test("given empty array, returns empty array", () => {
+  expect(dedupe([])).toEqual([]);
+});
 
 // Given an array with no duplicates
 // When passed to the dedupe function
 // Then it should return a copy of the original array
+test("array with no duplicates returns copy of original", () => {
+  expect(dedupe([1, 3, 4])).toEqual([1, 3, 4]);
+});
```
cjyuan, line 28: Your function implementation is correct. However, this test could be improved to better ensure
that any future changes continue to align with the expected behavior described on line 25:

Sprint-1/implement/dedupe.test.js
```diff
@@ -32,7 +32,9 @@ test("array with no duplicates returns copy of original", () => {
 // Then it should return a new array with duplicates removed while preserving the
 // first occurrence of each element from the original array.
 test("removes duplicates, while preserving first occurrance of each element", () => {
-  expect(dedupe([1, 3, 3, "as", "as", "df"])).toEqual([1, 3, "as", "df"]);
+  const inputArr = [1, 3, 3, "as", "as", "df"];
+  expect(dedupe(inputArr)).toEqual([1, 3, "as", "df"]);
+  expect(dedupe(inputArr)).not.toBe(inputArr);
```
cjyuan, line 37: Supposedly the check is expected in the previous test. Well, at least you know how to carry out the test.

cjyuan (commented):
All good now.

PR #1220, opened 2026-05-07
London | 26-ITP-May | Raihan Sharif | Sprint 2 | Exercises

RaihanSharif:
## Learners, PR Template

Self checklist

## Changelist
Done all exercises. For the invert exercises, made decision to prevent inverting where multiple keys have the same value. e.g. {a: "f", b: "f"} as this would produce {f: "b"}, thereby losing some information.

### DIFF
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -4,7 +4,7 @@ function isObject(item) {
 
 function contains(object, key) {
   if (isObject(object)) {
-    return Object.keys(object).includes(key);
+    return key in object;
   }
   throw new TypeError("Item is not plain object, Date, or Map");
 }
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,3 +1,12 @@
-function contains() {}
+function isObject(item) {
+  return typeof item === "object" && item !== null && !Array.isArray(item);
+}
+
+function contains(object, key) {
+  if (isObject(object)) {
+    return Object.keys(object).includes(key);
+  }
+  throw new TypeError("Item is not plain object, Date, or Map");
+}
 
 module.exports = contains;
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,12 +1,18 @@
 function isObject(item) {
-  return typeof item === "object" && item !== null && !Array.isArray(item);
+  return (
+    typeof item === "object" &&
+    item !== null &&
+    !Array.isArray(item) &&
+    !(item instanceof Date) &&
+    !(item instanceof Map)
+  );
 }
 
 function contains(object, key) {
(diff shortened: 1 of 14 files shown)


Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,18 @@
...
+    item !== null &&
+    !Array.isArray(item) &&
+    !(item instanceof Date) &&
+    !(item instanceof Map)
+  );
+}
+
+function contains(object, key) {
+  if (isObject(object)) {
+    return key in object;
```
```
```

Sprint-2/implement/querystring.js
```diff
@@ -1,13 +1,50 @@
...
+    like "toString", if the query string has a "toString=value".
+
+
+
+    if a kv pair already exists on queryParams, and value is the same as the incoming value, ignore. It's a duplicate
+    Otherwise, if it has a single value, change value to array containing the old value and new value.
+
+    if value at key is an array, and contains the incoming value, ignore. Else, push it to the array.
+    */
+    if (Object.prototype.hasOwnProperty.call(queryParams, key)) {
```
cjyuan, line 37: Could also use `Object.hasOwn()`

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,16 @@
...
+  }
+  return list.reduce((acc, curr) => {
+    if (typeof curr !== "string" && typeof curr !== "number") {
+      throw new TypeError(
+        `Invalid element: ${curr} must be a string or number`
+      );
+    }
+    acc[curr] = (acc[curr] ?? 0) + 1;
+    return acc;
+  }, {});
```
```
```

Suggestion:
- use `Object.hasOwn()`

Sprint-2/interpret/invert.js
```diff
@@ -7,23 +7,68 @@
...
+    // if a value is not a number or string, it can't be used as a key in the inverted version
+    if (typeof value !== "string" && typeof value !== "number") {
+      throw new TypeError(`Value ${JSON.stringify(value)} is not a valid key`);
+    }
+
+    // if inverted object already has this key, don't overwrite, throw error
+    if (Object.prototype.hasOwnProperty.call(invertedObj, String(value))) {
+      throw new Error(`Duplicate value "${value}" found — cannot invert`);
+    }
+    invertedObj[String(value)] = key;
```
cjyuan: All values (except Symbol) when used as keys, are converted implicitly to string. So a cast to string is optional.

Sprint-2/implement/contains.js
RaihanSharif: hasOwn excludes inherited properties. I hadn't considered that we'd want to exclude inherited properties.

I suppose in this case, we do want  to exlcude inherited properties.

cjyuan: The spec didn't specify. So as long as you are aware of the difference, that's enough.

PR #1221, opened 2026-05-08
London | 26-ITP-May | Raihan Sharif | Sprint 3 | Quote Generator

RaihanSharif:
## Learners, PR Template

Self checklist

## Changelist
On page load display random quote. On button click display random quote.

### DIFF
--- a/Sprint-3/quote-generator/quotes.js
+++ b/Sprint-3/quote-generator/quotes.js
@@ -491,3 +491,14 @@ const quotes = [
 ];
 
 // call pickFromArray with the quotes array to check you get a random quote
+
+const quoteText = document.getElementById("quote");
+const authorText = document.getElementById("author");
+
+const newQuoteBtn = document.getElementById("new-quote");
+
+newQuoteBtn.addEventListener("click", () => {
+  const selectedQuote = pickFromArray(quotes);
+  quoteText.textContent = selectedQuote.quote;
+  authorText.textContent = `-- ${selectedQuote.author}`;
+});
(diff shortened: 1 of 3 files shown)


Sprint-3/quote-generator/quotes.js
```diff
@@ -491,3 +491,16 @@ const quotes = [
...
+
+const quoteText = document.getElementById("quote");
+const authorText = document.getElementById("author");
+
+const newQuoteBtn = document.getElementById("new-quote");
+
+newQuoteBtn.addEventListener("click", () => {
+  const selectedQuote = pickFromArray(quotes);
+  quoteText.textContent = selectedQuote.quote;
+  authorText.textContent = `-- ${selectedQuote.author}`;
```
cjyuan: The leading `--` appears to be for styling purposes. Keeping it in the CSS or in HTML could allow us to adjust the view without changing any JavaScript code.

Sprint-3/quote-generator/quotes.js
```diff
@@ -491,3 +491,16 @@ const quotes = [
...
+});
+
+document.onload = newQuoteBtn.click();
```
cjyuan: Why assign the return value of `.click()` to `document.onload`?

cjyuan (commented):
Placing all the "run on load" code in one place can make it clearer that 
"this is what runs when the page loads." 

Consider organizing the code in the following manner:
```javascript
// Function definitions 
function showQuote() {
  ...
}
...

  // code to be executed on page load, including code that setup event listeners
});
```

Sprint-3/quote-generator/quotes.js
RaihanSharif: meant to do `window.onload = () => newQuoteBtn.click();`

Now uses function to generate quote and calls that instead.

PR #1223, opened 2026-05-08
London | 26-ITP-May | Raihan Sharif | Sprint 3 | Alarm Clock

RaihanSharif:
## Learners, PR Template

Self checklist

## Changelist
Implemented the basic feature using setInterval. I tried to use setInterval to poll Date.now() every 500ms, as I thought that would be more robust, but the test suit meant I couldn't do that.

### DIFF
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -1,4 +1,64 @@
-function setAlarm() {}
+// need to set interval outside of setAlarm() so that every time the user clicks set alarm
+// the previous interval is deleted. Otherwise there will be overlapping intervals
+// if a user clicks on set alarm while an countdown is already active.
+let intervalId = null;
+
+function setAlarm() {
+  const alarmClockInput = document.getElementById("alarmSet");
+  const clockDisplay = document.getElementById("timeRemaining");
+  let secondsRemaining = parseInt(alarmClockInput.value, 10);
+
+  if (!isValidInput(secondsRemaining)) {
+    alert("please enter a valid positive integer!");
+    return;
+  }
+
+  clearInterval(intervalId);
+
+  // display immediately on click;
+  const minutes = Math.floor(secondsRemaining / 60);
+  const seconds = secondsRemaining % 60;
+  displayTime(minutes, seconds, clockDisplay);
+
+  intervalId = setInterval(() => {
+    secondsRemaining--;
+
+    if (secondsRemaining <= 0) {
+      clearInterval(intervalId);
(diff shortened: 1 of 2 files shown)


Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,64 @@
...
+    alert("please enter a valid positive integer!");
+    return;
+  }
+
+  clearInterval(intervalId);
+
+  // display immediately on click;
+  const minutes = Math.floor(secondsRemaining / 60);
+  const seconds = secondsRemaining % 60;
+  displayTime(minutes, seconds, clockDisplay);
```
cjyuan: These three lines of code are repeated on lines 33-35. Wouldn't it be better to compute the minutes and seconds in the `displayTime()` function?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,64 @@
...
+  const clockDisplay = document.getElementById("timeRemaining");
+  let secondsRemaining = parseInt(alarmClockInput.value, 10);
+
+  if (!isValidInput(secondsRemaining)) {
+    alert("please enter a valid positive integer!");
+    return;
+  }
+
+  clearInterval(intervalId);
+
```
cjyuan, line 17: Currently when starting a new countdown, the application does not always return to a clean initial state, 

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,64 @@
...
+function isValidInput(input) {
+  if (input === "" || isNaN(input)) {
+    return false;
+  }
+
+  if (!Number.isInteger(input)) {
+    return false;
+  }
+
+  if (input < 0) {
```
cjyuan: Could consider changing the condition to `input <= 0` to align with the `min` value introduced in HTML.

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,68 @@
...
+// need to set interval outside of setAlarm() so that every time the user clicks set alarm
+// the previous interval is deleted. Otherwise there will be overlapping intervals
+// if a user clicks on set alarm while an countdown is already active.
+let intervalId = null;
+
+function setAlarm() {
+  const alarmClockInput = document.getElementById("alarmSet");
+  const clockDisplay = document.getElementById("timeRemaining");
+  let secondsRemaining = parseInt(alarmClockInput.value, 10);
+  const setBtn = document.getElementById("set");
```
cjyuan: `setBtn` is not used anywhere.

PR #1226, opened 2026-05-13
London | 26-ITP-May | Alex Jamshidi | Sprint 2 | Exercises

Alex-Jamshidi:
## Learners, PR Template

Self checklist

## Changelist

Completed exercises for Sprint 2

### DIFF
--- a/Sprint-2/debug/address.js
+++ b/Sprint-2/debug/address.js
@@ -1,4 +1,5 @@
 // Predict and explain first...
+// This will fail because to parse from an object requires a key, not an index value
 
 // This code should log out the houseNumber from the address object
 // but it isn't working...
@@ -12,4 +13,4 @@ const address = {
   postcode: "XYZ 123",
 };
 
-console.log(`My house number is ${address[0]}`);
+console.log(`My house number is ${address.street}, ${address.city}, ${address.country}, ${address.postcode}`);
--- a/Sprint-2/debug/address.js
+++ b/Sprint-2/debug/address.js
@@ -13,4 +13,6 @@ const address = {
   postcode: "XYZ 123",
 };
 
-console.log(`My house number is ${address.street}, ${address.city}, ${address.country}, ${address.postcode}`);
+console.log(
+  `My house number is ${address.houseNumber}, ${address.street}, ${address.city}, ${address.country}, ${address.postcode}`
+);
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,3 +1,5 @@
-function contains() {}
+function contains() {
+    return true
+}
 
 module.exports = contains;
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,5 +1,8 @@
-function contains() {
-    return true
+function contains(object, property) {
+    if (typeof object !== 'object' || Array.isArray(object)) {return false;}
+    if (typeof property !== 'string') {return false;}
+
+    return Object.hasOwn(object, property)
 }
 
 module.exports = contains;
--- a/Sprint-2/implement/contains.js
+++ b/Sprint-2/implement/contains.js
@@ -1,8 +1,12 @@
 function contains(object, property) {
-    if (typeof object !== 'object' || Array.isArray(object)) {return false;}
-    if (typeof property !== 'string') {return false;}
+  if (typeof object !== "object" || !object || Array.isArray(object)) {
+    return false;
+  }
+  if (typeof property !== "string" || !object) {
+    return false;
+  }
 
-    return Object.hasOwn(object, property)
+  return Object.hasOwn(object, property);
 }
 
 module.exports = contains;
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -16,7 +16,7 @@ as the object doesn't contains a key of 'c'
 // When passed an object and a property name
 // Then it should return true if the object contains the property, false otherwise
 
-describe("contains", () =>{
+describe("contains", () => {
 // Given an empty object
 // When passed to contains
 // Then it should return false
--- a/Sprint-2/implement/contains.test.js
+++ b/Sprint-2/implement/contains.test.js
@@ -12,24 +12,44 @@ as the object doesn't contains a key of 'c'
 */
 
 // Acceptance criteria:
-
 // Given a contains function
 // When passed an object and a property name
 // Then it should return true if the object contains the property, false otherwise
 
+describe("contains", () =>{
 // Given an empty object
 // When passed to contains
 // Then it should return false
-test.todo("contains on empty object returns false");
+it("contains on empty object returns false", () => {
+    const object = {};
+    const property = 'a'
+    expect(contains(object, property)).toEqual(false)
+});
 
 // Given an object with properties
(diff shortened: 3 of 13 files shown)


Sprint-2/debug/address.js
```diff
@@ -12,4 +13,4 @@ const address = {
   postcode: "XYZ 123",
 };
 
-console.log(`My house number is ${address[0]}`);
+console.log(`My house number is ${address.street}, ${address.city}, ${address.country}, ${address.postcode}`);
```
cjyuan: Your code outputs every parts of an address except the "house number".

Sprint-2/implement/contains.js
```diff
@@ -1,3 +1,8 @@
-function contains() {}
+function contains(object, property) {
+    if (typeof object !== 'object' || Array.isArray(object)) {return false;}
+    if (typeof property !== 'string') {return false;}
```
cjyuan: Line 2 cannot yet reject `null`.

Sprint-2/implement/contains.test.js
```diff
@@ -12,24 +12,44 @@ as the object doesn't contains a key of 'c'
...
+    expect(contains(object, property)).toEqual(false)
+});
 
 // Given invalid parameters like an array
 // When passed to contains
 // Then it should return false or throw an error
+it("given invalid parameter (an array) returns false or throws an error", () => {
+    const object = [];
+    const property = [];
+    expect(contains(object, property)).toEqual(false)
```
cjyuan: Your function can correctly return `false` when the first argument is an array. 

However, we write tests not only to verify our current implementation, but also to ensure that future changes do not alter the function's expected behavior.

The current test cannot yet confirm the function can correctly return `false` when the first argument is an array 
because `contains([], [])` could also return `false` simply because `[]` is not a key of the array.

Sprint-2/implement/lookup.js
```diff
@@ -1,5 +1,9 @@
-function createLookup() {
-  // implementation here
+function createLookup(arrayOfArrays) {
+  const lookup = {}
+  for (array of arrayOfArrays) {
```
cjyuan: `array` not yet declared.

Sprint-2/implement/querystring.js
```diff
@@ -3,14 +3,40 @@ function parseQueryString(queryString) {
...
   }
 
   return queryParams;
 }
 
+// this function below can have the object PEC expanded to include all percentage-encoded characters
+// currently without adding this database, an unknown PEC could cause the code to get stuck in an infinite loop
+
+function changePEC(string) {
+  const PEC = {"%24": "$", "%2F": "/"}
```
cjyuan: There are more than two percentage encoded characters. 

Consider using the built-in function to decode PEC's in the string instead.

Sprint-2/implement/querystring.js
```diff
@@ -3,14 +3,40 @@ function parseQueryString(queryString) {
...
 
   for (const pair of keyValuePairs) {
-    const [key, value] = pair.split("=");
+    if (pair !== "") {
+    let index = pair.indexOf("=")
+    if (!pair.includes("=")) {index = pair.length}
+    const [key, value] = [pair.slice(0, index), pair.slice(index + 1)]
     queryParams[key] = value;
+    }
+    else {};
```
cjyuan: Why leave a do-nothing else block in the function?

Sprint-2/implement/tally.js
```diff
@@ -1,3 +1,12 @@
-function tally() {}
+function tally(array) {
+    if (!Array.isArray(array)) {throw new Error("Invalid array");}
+    tally = {}
+    array.forEach((item) =>{
+        if (item in tally) {tally[item] += 1}
+        else tally[item] = 1;
+    })
+    return tally
```
```
```

Suggestion:
- use `Object.hasOwn()`

cjyuan (commented):
Some of the code is not consistently formatted or indented.

Have you installed the **prettier** VSCode extension and enabled "Format on save/paste" on VSCode, as recommended in 
?

Sprint-2/debug/address.js
Alex-Jamshidi: Added house number.

I had installed prettier and enabled the formatting, however it wasn't working because I hadn't set as my default formatter. I have formatted this code now. (and gone back through all my old PRs and formatted too!)

thanks :)

Sprint-2/implement/contains.js
Alex-Jamshidi: updated to reject null

written tests to confirm

Sprint-2/implement/contains.test.js
Alex-Jamshidi: Thanks for this, I understand.

I wrote several tests to cover all options for arrays (filled or empty).

Sprint-2/implement/lookup.js
Alex-Jamshidi: thanks, have declared with const

Sprint-2/implement/querystring.js
Alex-Jamshidi: I didn't know there was a function for this *facepalm*
I have updated to use the function, it was a lot easier :)

Thanks

Sprint-2/implement/querystring.js
Alex-Jamshidi: It's a good question that doesn't have a good answer, I have removed it.

Thanks

Sprint-2/implement/tally.js
Alex-Jamshidi: that was a tough one

Wrote test for case and updated code to pass

Sprint-2/implement/contains.test.js
```diff
@@ -12,24 +12,51 @@ as the object doesn't contains a key of 'c'
...
+  });
+
+  // Given invalid parameters like an array
+  // When passed to contains
+  // Then it should return false or throw an error
+  it("given invalid parameter (an array) returns false or throws an error", () => {
+    expect(contains([], [])).toEqual(false);
+    expect(contains(["a", 1], "a")).toEqual(false);
+    expect(contains({ a: 1, b: 2 }, ["a"])).toEqual(false);
+    expect(contains(["a", 1], ["a"])).toEqual(false);
```
cjyuan: `"a"` and `["a"]` are not keys of `["a", "1"]`, but `"0"` and `"1"` are. You should use either "0" or "1" as key to ensure the function returns false because the first parameter is an array.

Sprint-2/implement/querystring.js
```diff
@@ -3,11 +3,24 @@ function parseQueryString(queryString) {
   if (queryString.length === 0) {
     return queryParams;
   }
+
+  // Adds percentage-encoded characters
+  queryString = decodeURIComponent(queryString);
```
cjyuan: If you decode the characters so early and some the decoded characters are '=', '+', or '&', the query will be incorrectly parsed.

Sprint-2/implement/contains.test.js
Alex-Jamshidi: updated and rerun tests to pass

Sprint-2/implement/querystring.js
Alex-Jamshidi: updated this to resolve decoded chars at the end
refactored code to make more clear

cjyuan (commented):
Changes look good. Well done!

PR #1227, opened 2026-05-13
London | 26-ITP-May | Alex Jamshidi | Sprint 3 | Alarm Clock 

Alex-Jamshidi:
## Learners, PR Template

Self checklist

## Changelist

created alarm clock code

### DIFF
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -1,7 +1,7 @@
 function setAlarm() {
   timeRemaining = document.getElementById("timeRemaining")
-  alarmSet = document.getElementById("alarmSet").value
-
+  timeSet = document.getElementById("alarmSet").value
+  const time = convertTime(timeSet)
   
   
 
@@ -11,7 +11,7 @@ function setAlarm() {
 
 function countdown(time) {
   setInterval(() => {
-
+    
   }, 1000)
 }
 
@@ -22,6 +22,16 @@ function convertTime(timeSeconds) {
   console.log([minutes, seconds])
   return [minutes, seconds];
 }
+
+function printTime(time) {
+  const paddedHours = String(time[0]).padStart(2, "0");
+  const paddedSeconds = String(time[1]).padStart(2, "0");
+  return paddedHours + ":" + paddedSeconds;
+}
+
+printTime([50,40])
+printTime([50,4])
+printTime([5,40])
 
 convertTime(5000000)
 convertTime(50)
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -1,8 +1,8 @@
 let timer;
 
 function setAlarm() {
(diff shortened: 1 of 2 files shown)


Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,52 @@
-function setAlarm() {}
+let timer;
+
+function setAlarm() {
+  timeSet = document.getElementById("alarmSet").value;
```
illicitonion: What's the scope of this variable? Is there a better scope for it?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,52 @@
...
+  const time = countdown(timeSet);
+  if (time > 0) {
+    countdown(time);
+  }
```
illicitonion: Why do you have this `if` block? What does it do? When will it run?

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,52 @@
...
+    timeRemaining -= 1;
+    if (timeRemaining == 0) {
+      clearInterval(timer);
+      playAlarm();
+    }
+    UpdateShownTime(timeRemaining);
+  }, 1000);
+}
+
+function UpdateShownTime(timeRemaining) {
```
illicitonion: The name of this function doesn't follow the same pattern of the other function names - can you make them consistent?

illicitonion (commented):
This generally looks good, but I left a few things to think about / work on.

Good job extracting away the duplication of showing a time into a reusable function.

Sprint-3/alarmclock/alarmclock.js
Alex-Jamshidi: this would have been a global variable without declaring it
changed to 'const' to restrict scope to function

Sprint-3/alarmclock/alarmclock.js
Alex-Jamshidi: It doesn't, this was old code that I was using to stop the countdown, but later added into the setInterval function.

Redundant code - removed.

Sprint-3/alarmclock/alarmclock.js
Alex-Jamshidi: removed capitalisation at start

Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,4 +1,49 @@
-function setAlarm() {}
+let timer;
+
+function setAlarm() {
+  const timeSet = document.getElementById("alarmSet").value;
+  const time = countdown(timeSet);
```
illicitonion: What's the point of this `time` variable now?

Alex-Jamshidi: have removed variable assignment as after deleting code snippet before, it is no longer used. countdown(timeSet) is run without result assigned to variable.

PR #1229, opened 2026-05-14
London | 26-ITP-May | Alex Jamshidi | Sprint 3 | Quote Generator App 

Alex-Jamshidi:
## Learners, PR Template

Self checklist

## Changelist

Created quote generator app to reandomly generate quotes on button click

### DIFF
--- a/Sprint-3/quote-generator/quotes.js
+++ b/Sprint-3/quote-generator/quotes.js
@@ -491,3 +491,34 @@ const quotes = [
 ];
 
 // call pickFromArray with the quotes array to check you get a random quote
+
+function setup() {
+  updateQuote();
+  appendQuote();
+}
+
+const state = {
+  quote: "",
+  author: "",
+};
+
+function updateQuote() {
+  const chosenQuote = pickFromArray(quotes);
+  state.quote = chosenQuote.quote;
+  state.author = chosenQuote.author;
+}
+
+function appendQuote() {
+  const quoteElem = document.getElementById("quote");
+  const authorElem = document.getElementById("author");
+  quoteElem.textContent = "";
+  authorElem.textContent = "";
+  quoteElem.append(state.quote);
+  authorElem.append(state.author);
+}
+
+document.getElementById("new-quote").addEventListener("click", function () {
+  setup();
+});
+
+window.onload = setup;
--- a/Sprint-3/quote-generator/quotes.js
+++ b/Sprint-3/quote-generator/quotes.js
@@ -492,9 +492,9 @@ const quotes = [
 
 // call pickFromArray with the quotes array to check you get a random quote
 
-function setup() {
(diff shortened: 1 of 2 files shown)


Sprint-3/quote-generator/quotes.js
```diff
@@ -491,3 +491,34 @@ const quotes = [
...
+  author: "",
+};
+
+function updateQuote() {
+  const chosenQuote = pickFromArray(quotes);
+  state.quote = chosenQuote.quote;
+  state.author = chosenQuote.author;
+}
+
+function appendQuote() {
```
illicitonion: Append isn't the most clear name here - it suggests that each time you call it an _additional_ quote will be added to the page, rather than replacing the existing quote.

(Even though you implementation does use `append` internally, the _effect_ of the function isn't to append something to the page).

Can you think of a more clear name?

Sprint-3/quote-generator/quotes.js
```diff
@@ -491,3 +491,34 @@ const quotes = [
...
+  const quoteElem = document.getElementById("quote");
+  const authorElem = document.getElementById("author");
+  quoteElem.textContent = "";
+  authorElem.textContent = "";
+  quoteElem.append(state.quote);
```
illicitonion: This works, but you could also directly set `textContent` to this value:
```js
quoteElem.textContent = state.quote
```

which may be more clear as to your intent when reading, and may avoid a brief flash of an empty quote in the middle.

Sprint-3/quote-generator/quotes.js
```diff
@@ -491,3 +491,34 @@ const quotes = [
...
+  authorElem.append(state.author);
+}
+
+document.getElementById("new-quote").addEventListener("click", function () {
+  setup();
```
illicitonion: `setup` isn't a very clear name for this action. It makes sense on load that this is what you're doing.

Personally I'd probably just have one function in file: `showNewQuote`, and have that be called both `onload` and on button click.

illicitonion (commented):
This works, but there are a few naming things to think about - please take a look :)

Sprint-3/quote-generator/quotes.js
Alex-Jamshidi: renamed function to showNewQuote

Sprint-3/quote-generator/quotes.js
Alex-Jamshidi: renamed to replaceQuote

Sprint-3/quote-generator/quotes.js
Alex-Jamshidi: instead of resetting .textcontent to "" and appending new quote/author, replaced quote/author directly with .textcontent
and removed append lines

PR #1144, opened 2026-03-28
London | 26-ATP-Jan | Boualem Larbi Djebbour | sprint 3 | Alarm clock app

djebsoft:
## Learners, PR Template

Self checklist

- [x] I have titled my PR with Region | Cohort | FirstName LastName | Sprint | Assignment Title
- [x] My changes meet the requirements of the task
- [x] I have tested my changes
- [x] My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/reviewing/style-guide/)

## Changelist
buildin the Alarn clock app.

## Questions

PR #1230, opened 2026-05-14
 London | 26-ATP-May | Boualem Larbi Djebbour | sprint 3 | Alarm clock app

djebsoft:
## Learners, PR Template

Self checklist

## Changelist
This PR previously discussed in #1144
changes since last PR: no changes

### DIFF
--- a/Sprint-3/alarmclock/alarmclock.js
+++ b/Sprint-3/alarmclock/alarmclock.js
@@ -1,5 +1,52 @@
-function setAlarm() {}
+//global variables: 1- userInput, 2- timeRemaining, 3- stop, 4- seconds, 5- minutes
 
+// get access to the "input" element
+const userInput = document.querySelector("#alarmSet");
+// get access to the "timeRemaining" element
+const timeRemaining = document.querySelector("#timeRemaining");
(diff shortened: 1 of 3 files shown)


Sprint-3/alarmclock/alarmclock.js
```diff
@@ -1,5 +1,53 @@
-function setAlarm() {}
+//global variables: 1- userInput, 2- timeRemaining, 3- stop, 4- seconds, 5- minutes
```
illicitonion: Please take a look at all of your comments in this file and see if they add more detail than just the line of code itself. e.g. the comment `get access to the "timeRemaining" element` simply restates exactly what the line after does. Remove any that aren't useful.

Comments should ideally be replaced by clear naming with code, and where they can't, should typically explain _why_ we're doing something, rather than _what_ we're doing. The code itself says what we're doing.

illicitonion (commented):
This looks good, but has one issue - if I set the alarm for 1 second, the alarm only actually plays after 2 seconds. Can you work out why and fix it up?

illicitonion (commented):
Looks great, well done!

